PR #2, opened 2014-09-30
PEP8 80 char line length

luc-demeyer:


dreispt:
- The copyright year should be left unchanged.
- The `# vim:` lines at the end of files can be removed.

luc-demeyer:
Daniel,

Thanks for the feedback.

Some questions:
-   The copyright year should be left unchanged.

I suppose there is some legal reason behind this guideline.

Can you explain (or point an article with explanation) ?
-   The # vim: lines at the end of files can be removed.

Is this blocking for PR approval (hence do you expect a new PR) or a tip for future PR’s ? 

Are the above guidelines documented somewhere ?

Kind Regards,

Luc

 http://www.noviat.com/ www.noviat.com

Rusatiralaan 1, 1083 Brussel

+32 2 808 86 38

dreispt:
> - The copyright year should be left unchanged. 

I don't have a documentation to point you at, but AFAIK the copyright counts since the date the work was first published. At best, you could write 2013 - 2014. I believe OpenERP/Odoo uses this last format frequently.

> - The # vim: lines at the end of files can be removed.

Not a blocking issue, of course. Just the recent best practice.

pedrobaeza:
We should put first Travis stuff to get the automatic status on this PR.

dreispt:
That's desirable, but I don't see a problem merging this before someone does that.

gurneyalex:
:+1:

pedrobaeza:
I also see reasonably to put copyright as 2013-2014, to see that you are working in this module since then, but in the end, it's your copyright: you can do what you want with it :wink: 

:+1: 

Regards.

luc-demeyer:
I don’t worry too much about this.

If someone works on the 2013 version of this module they see copyright from 2013 onwards.

Those downloading the 2014 version see the copyright as from 2014.

I’ll just stop changing the year field whenever I make a new version in 2015 or beyond.

Luc

 http://www.noviat.com/ www.noviat.com

Rusatiralaan 1, 1083 Brussel

+32 2 808 86 38

yvaucher:
@luc-demeyer you cannot remove the start year of the copyright this determines from when the copyright start.

Please restore 2013

From http://www.gnu.org/licenses/gpl-howto.html

```
The **copyright notice should include the year in which you finished preparing the release** (so if you 
finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each 
release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and 
some were finished in 1999. If several people helped write the code, use all their names.

For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead 
of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a 
“copyrightable” year that would be listed individually; and you make an explicit statement in your 
documentation about this usage.
```

luc-demeyer:
I have restored 2013 in the copyright hence this PR should now be ready for merge.

Regards,

Luc

 http://www.noviat.com/ www.noviat.com

Rusatiralaan 1, 1083 Brussel

+32 2 808 86 38

yvaucher:
:+1:  @luc-demeyer thanks

PR #6, opened 2014-10-25
[FIX] Installation dependency

alejandroperezcosio:
Ubuntu 14.04 - Odoo 8
While trying to install this module, a "no module xlwt" appears even if the module is properly installed.
Removing this line the module could be installed without error messages.

### DIFF
--- a/report_xls/__openerp__.py
+++ b/report_xls/__openerp__.py
@@ -75,7 +75,6 @@ Excel reports in odoo.
 
     """,
     'depends': ['base'],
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Hi, Alejandro, the dependency is put there at purpose and detection machinery is working well in all installations I have tested, so you have to check your installation. Maybe an incorrect virtualenv or PYTHONPATH. What you cannot do is to remove dependency check, because this will lead to an error when you use the module (exporting to Excel, for example).

:-1:

PR #14, opened 2015-06-10
Add module report_xml.

yajo:
In order to fix https://github.com/OCA/l10n-spain/issues/174 I need to generate XML files, so here is the reporting engine for that.

It allows developers to generate XML reports using QWeb, and print them using the standard Odoo tools. See README for instructions.


report_xml/README.rst
```diff
@@ -0,0 +1,82 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Module name
```
pedrobaeza: Put module name here

pedrobaeza:
Can you provide a _report_xml_sample_ module as an example for this?

yajo:
Are you sure you want a sample module? Is it not better to add demo data to this one? After all, reports are data files...

pedrobaeza:
Well, it can be a solution, but I think people would prefer to test it and study independently from having the option of demo data checked.

mylbco:
@pedrobaeza yes, I can agree on that

yajo:
Creating a dummy module does not seem very pleasing to me. This module is going to be extensively used in the upcoming [`l10n_es_training`](https://github.com/grupoesoc/l10n-spain/tree/l10n_es_training/l10n_es_training) module, so I took that one as the sample module and referenced it in the README.

BTW I also added a couple of missing pieces to this module: a base template and a base XSD report checker.

pedrobaeza:
Then it's going to be very difficult to review this PR unless we have any sample

Yajo:
You can see a sample XML template [here](https://github.com/grupoesoc/l10n-spain/blob/l10n_es_training/l10n_es_training/views/event_event_templates.xml) and sample usage of the XSD validator [here](https://github.com/grupoesoc/l10n-spain/blob/l10n_es_training/l10n_es_training/models/report.py). It's better than samples, since it's a real module. What's the problem?

Yajo:
Done.

yajo:
Is the sample OK?

pedrobaeza:
Thanks for the sample! Maybe I miss an option to select the fields of the XML, but as a first version is very appreciated.

:+1:

yajo:
What do you mean?

pedrobaeza:
I don't see in the sample a way to define only some fields of the record to export in the XML, or may I miss something?

yajo:
It's a [QWeb report](https://www.odoo.com/documentation/8.0/reference/reports.html) in all means. [In the sample I only export id and name](https://github.com/OCA/reporting-engine/pull/14/files#diff-522bb8b9e1448081f11bcec6f4989322) because it is an example and it should be simple by definition, but you can export any field you want.

See a [real world sample here](https://github.com/grupoesoc/l10n-spain/blob/l10n_es_training/l10n_es_training/views/event_event_templates.xml) to see how versatile it is (it's a WIP).

pedrobaeza:
Ah, OK, I didn't get the real usage then (that's why I insist then on the sample and/or documentation), and you can see that still the sample maybe it's not enough, but I'm not going to insist more for now on this.

yajo:
The READMEs are short, but just follow _exactly_ what they say and you should have your XML report in no time, and the real docs are those of QWeb, that are linked in them. Currently QWeb reports are HTML, which is itself an XML, so the only thing this module does is to make it download it as XML instead of rendering it in the browser.

XSD are XML schemas that validate XML documents, very often used to make sure the XML you send is going to be accepted by the other party. There are other validation methods (DTD, RelaxNG...), but I did not include any of them since I don't need them right now and XSD seems to be the most used.

The sample is small on purpose and IMHO it's better that way. It includes also [an XSD](https://github.com/OCA/reporting-engine/pull/14/files#diff-bddc5e4e81ef3ecc7289fa8387d37b40) and [the method to check it](https://github.com/OCA/reporting-engine/pull/14/files#diff-7b4010b86fa141ffe1869c47aa913347). I encourage you to install it and see how it works and read carefuly the READMEs, you'll see it's very easy. :wink:

yajo:
Please don't merge this still, I've got another PR coming that I consider more important.

yajo:
Closing this until https://github.com/OCA/reporting-engine/pull/18 gets fixed.

PR #16, opened 2015-06-16
change old link to lp, add license and summary

yvaucher:


mdietrichc2c:
:+1:

base_report_assembler/__openerp__.py
```diff
@@ -34,11 +35,13 @@
...
 
-- Invoices: invoice_report_assemble (lp:account-invoice-report)
+- Assemble invoice report: `invoice_report_assemble`
+(git@github.com:OCA/account-invoice-report)
 
 """,
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'maintainer': 'Camptocamp',
+ 'license': 'AGPL-3',
  'website': 'http://www.camptocamp.com/',
```
lmignon, line 45: @yvaucher  The website name should it be 'https://github.com/OCA/reporting-engine/'?

yvaucher, line 45: @lmignon do we have a rule for that? I think most of modules have something else. But that's a good idea to point to where the code is if not to oca website.

Being said it seems we need a `__openerp__.py` template in here:
https://github.com/OCA/maintainer-tools/tree/master/template/module

sbidoul:
:+1:

While you are at it, can you add the missing ACL so runbot becomes green?
The model assembled.report has no access rules, consider adding one. E.g. access_assembled_report,access_assembled_report,model_assembled_report,,1,1,1,1

base_report_assembler/__openerp__.py
yvaucher, line 45: Seems website field is a mess

```
grep -rn website --include=__openerp__.py | grep -v OCB | cut -d ':' -f 4- | cut -d "'" -f 2 | cut -d '"' -f 2 | sort | uniq -c | sort -nr
159 http://www.camptocamp.com
114 http://www.savoirfairelinux.com
 54 http://www.agilebg.com
 49 
 38 http://www.akretion.com
 36 http://camptocamp.com
 35 http://www.akretion.com/
 34 http://miketelahun.wordpress.com
 27 http://www.julius.fr
 24 http://www.camptocamp.com/
 20 https://github.com/YannickB/community-management
 14 http://www.openerp.com
 12 http://www.openerp-italia.org
 12 http://www.acsone.eu
 12 http://launchpad.net/oemedical
 11 http://openerpbrasil.org
 10 https://github.com/OCA/bank-payment
  9 http://openerp.camptocamp.com
  9 http://acsone.eu
  ...
```

yvaucher:
@sbidoul done

sbidoul:
@yvaucher Thanks. :+1: 

No question about the website URL from my side. It may be the topic of an OCA policy in the future, but ATM there is none.

pedrobaeza:
:+1:

sbidoul:
@yvaucher runbot is still red. The .csv must be referenced from `__openerp__.py` I guess.

yvaucher:
right did that a bit fast

yvaucher:
amended

lmignon:
@yvaucher In my mind the website is the one where I can find the sources. Since I don't know if there is a specific rules for that:  :+1:

base_report_assembler/__openerp__.py
dreispt, line 45: @yvaucher The truth is that there is no directive to follow for the website key. IMO it should be the URL for the corresponding, GitHub repo, or for a documentation homepage in case it exists.

PR #18, opened 2015-07-03
Add unified COPYRIGHT & LICENSE.

yajo:
This repository lacks a LICENSE file, and that must be fixed.

By the way, [based in the logic of upstream Odoo, GPL headers are not required nowadays](https://github.com/odoo/odoo/commit/0bd454534818941bc3216025ebe3dcac3d78c19d), so less garbage in files is always nice.

I imagine this can cause some controversial, but if you want the garbage please at least add the LICENSE & COPYRIGHT or I will not submit code here.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
+  The licenses for most software and other practical works are designed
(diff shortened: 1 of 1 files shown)


yajo:
I decided to split in two commits. The first adds LICENSE, with IMHO is required. I think anyone understands that sending code to a repository without a proper LICENSE is somehow uncomfortable.

The second commit handles and explains the unification.

In case you don't want the second, at least pick up the first one please.

pedrobaeza:
You should remove COPYRIGHT file, as this is not really true. The copyright is of the corresponding people. OCA has through the CLA some rights only. Look at the template for a repo here: https://github.com/OCA/maintainer-quality-tools/tree/master/sample_files

pedrobaeza:
All rights reserved is an outdated expression

yajo:
Sorry, the COPYRIGHT stuff was something I thought it could be useful, and I based it in the upstream movements. IANAL, but surely Odoo have good ones.

Anyway, it's away now. At least LICENSE should be required.

yajo:
I understand changing copyright headers can be cumbersome, and it's not much needed after all. I removed that commit. At least merge the license please.

yvaucher:
:+1: I proceed with the merge as LICENSE file should be there in each OCA repo

PR #21, opened 2015-07-22
Report xml

yajo:
Resending https://github.com/OCA/reporting-engine/pull/14 after https://github.com/OCA/reporting-engine/pull/18 has been merged. Quoting from there:

> In order to fix OCA/l10n-spain#174 I need to generate XML files, so here is the reporting engine for that.
> 
> It allows developers to generate XML reports using QWeb, and print them using the standard Odoo tools. See README for instructions.


pedrobaeza:
I repeat my :+1:

rafaelbn:
:+1:

PR #25, opened 2015-10-07
[FIX] add arg to workbook creation

bealdav:
## Issue

When you use a lot of styles applied to a big number of cells, you fall into this error
http://stackoverflow.com/questions/17130516/xlwt-set-style-making-error-more-than-4094-xfs-styles
## Solution

this fix


pedrobaeza:
It looks very straight-forward, so :+1:

yvaucher:
:+1:

PR #26, opened 2015-10-08
7.0 Fix TravisCI tests

dreispt:


.travis.yml
```diff
@@ -0,0 +1,63 @@
...
+      - python-lxml  # because pip installation is slow
+      - python-simplejson
+      - python-serial
+      - python-yaml
+      - python-pypdf  # repo specific
+
+env:
+  global:
+  - VERSION="7.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  #- TRANSIFEX_USER='transbot@odoo-community.org'
```
pedrobaeza: You can remove all this part, but you need the secure line. Check here what is the corresponding one: https://github.com/OCA/maintainer-quality-tools/issues/194

oca_dependencies.txt
```diff
@@ -0,0 +1,17 @@
...
+# sale-workflow
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# sale-workflow https://github.com/OCA/sale-workflow
+#
+# To provide both the URL and a branch, use:
+# sale-workflow https://github.com/OCA/sale-workflow branchname
+
+account-invoice-reporting
```
pedrobaeza: Why is this needed?

dreispt: The dependency is announced here: https://github.com/OCA/reporting-engine/blob/7.0/base_report_assembler/__openerp__.py#L39

pedrobaeza: That's only a complementary module that depends on this one, not the reverse. You can check manifests.

yvaucher:
:+1:

pedrobaeza:
:+1:

PR #28, opened 2015-11-05
Add Module report_element_visibility

hurrinico:
Adds new Module Report Element Visibility
Use Case:
I have an invoice with a huge number of lines, if i print the report i'll have a printing of many pages. In this case i may need to have on print's header the invoice's number in all the pages but first (because in first page invoice's number is already in). In this module i've created two new classes "not-first-page" and "not-last-page" to solve this issue.


eLBati:
:+1: thanks

report_element_visibility/README.rst
```diff
@@ -0,0 +1,74 @@
...
+ - not-last-page: Shows element in every page but last
+
+Usage
+=====
+
+To use this module, you need to:
+
+Install the module, then in the QWEB ``ir.ui.views`` used by your
+report, you can add an element with css class with not-first-page or not-last-page.
+For exemple if i need to improve invoice report header with
```
pedrobaeza: s/exemple/example

pedrobaeza:
Module name should be _report_qweb_element_visibility_, because it only applies to QWeb.

pedrobaeza:
Can you add the reverse ones (first-page and last-page) to allow to have a report header and a report footer?

report_element_visibility/README.rst
```diff
@@ -0,0 +1,74 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Report Element Visibility
+=========================
+
+This module allows you to add 2 elements in QWEB reports:
+ - not-firt-page: Shows element in every page but first
```
pedrobaeza: firt/first

report_element_visibility/README.rst
```diff
@@ -0,0 +1,74 @@
...
+ - not-last-page: Shows element in every page but last
+
+Usage
+=====
+
+To use this module, you need to:
+
+Install the module, then in the QWEB ``ir.ui.views`` used by your
```
pedrobaeza: Don't put "Install the module", because that's not something about the usage.

pedrobaeza: I think this can be in Configuration section instead.

report_element_visibility/README.rst
pedrobaeza: s/i/you

report_element_visibility/README.rst
```diff
@@ -0,0 +1,74 @@
...
+report, you can add an element with css class with not-first-page or not-last-page.
+For exemple if i need to improve invoice report header with
+invoice's number in every page but first, and sale order report header 
+with order's name in every page but last, i'll add this code to external_layout_header::
```
pedrobaeza: s/i'll add/add

report_element_visibility/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+    <t t-if="o._table=='account_invoice'">
+        <div class="not-first-page">
```
pedrobaeza: Why don't implement this like a method that you can evaluate with t-if?

hurrinico: Because this is just an example of module's usage.
This module only adds classes "not-in-first-page" and "not-in-last-page".
But it's up to the user use them in header/footer or body's report.

report_element_visibility/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+# GNU Affero General Public Licensefor more details.                    #
+#                                                                       #
+# You should have received a copy of the                                #
+# GNU Affero General Public License                                     #
+# along with this program.  If not, see <http://www.gnu.org/licenses/>. #
+#                                                                       #
+#########################################################################
+
+{
+    'name': 'Report Extend Not in first Page',
```
tafaRU: Please change it accordingly its technical name.

report_element_visibility/README.rst
pedrobaeza: I'm talking of implementing classes as conditions, so you can make:

```
<t t-if="o._table=='account_invoice' and is_not_first_page()">
    <span t-esc="o.number"/>
</t>
```

hurrinico: @pedrobaeza is_not_first_page() is a python or javascript method?
We have use this solution because with python we don't know how to get current page number, and because we think that a css class is more user's friendly.

pedrobaeza: A python expression seems friendlier for me, but I have no idea if it can be achieved

pedrobaeza:
What about the module renaming and the other suggestion to add first-page/last-page? Thinking twice, even a better name is _report_qweb_element_page_visibility_

hurrinico:
@pedrobaeza  I've changed module's name and added first-page/last-page.
About "A python expression seems friendlier for me, but I have no idea if it can be achieved", this module is based on JavaScript because i need to know current page, first page and last page of a report, and i don't know how to achieve them by python

report_qweb_element_page_visibility/README.rst
```diff
@@ -0,0 +1,76 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Report Qweb ElementPage Visibility
+==================================
+
+This module allows you to add 4 elements in QWEB reports:
```
pedrobaeza: s/add 4 elements/use 4 classes

report_qweb_element_page_visibility/README.rst
```diff
@@ -0,0 +1,76 @@
...
+ - not-first-page: Shows element in every page but first
```
pedrobaeza: Put a line break above to allow the correct rendering and remove the initial space.

report_qweb_element_page_visibility/README.rst
```diff
@@ -0,0 +1,76 @@
...
+ - first-page: Shows element only on first page
+ - last-page: Shows element only on last page
+
+Usage
+=====
+
+To use this module, you need to:
+
+In the QWEB ``ir.ui.views`` used by your report,
+you can add an element with css class with not-first-page or not-last-page.
```
pedrobaeza: s/with.../with any of the classes described above

pedrobaeza:
Minors remarks, but :+1:

report_qweb_element_page_visibility/README.rst
```diff
@@ -0,0 +1,76 @@
...
+
+==================================
+Report Qweb ElementPage Visibility
+==================================
+
+This module allows you to add 4 elements in QWEB reports:
+ - not-first-page: Shows element in every page but first
+ - not-last-page: Shows element in every page but last
+ - first-page: Shows element only on first page
+ - last-page: Shows element only on last page
```
tafaRU: Please replace Show by show in all lines above.

hurrinico:
@pedrobaeza @tafaRU i've done the requested changes

pedrobaeza:
:+1:

tafaRU:
:+1:

gurneyalex:
nice addon! thanks :smile_cat:

leemannd:
Thank you

PR #31, opened 2015-11-22
[ADD] report_qweb_signer addon

antespi:
This module was written to extend the functionality of report addon to sign
PDFs using a PKCS#12 certificate.

To install this module, you need to install PortableSigner Java application:

```
apt-get install openjdk-7-jre-headless unzip
wget -O PortableSigner-Generic-2.0.38c0573.zip "http://downloads.sourceforge.net/project/portablesigner/portablesigner/2.0-Release/PortableSigner-Generic-2.0.38c0573.zip?r=http%3A%2F%2Fsourceforge.net%2Fprojects%2Fportablesigner%2Ffiles%2F%3Fsource%3Dtyp_redirect&ts=1448015010&use_mirror=vorboss"
unzip -d PortableSigner-2.0 PortableSigner-Generic-2.0.38c0573.zip
cd PortableSigner-2.0
chmod +x linux-install.sh
./linux-install.sh
chmod +x /usr/local/bin/PortableSigner
```


antespi:
Is it possible to install in runbot PortableSigner Java application as indicate in README? This is necessary to test this PR

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,123 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+==================
+PDF reports signer
+==================
+
+This module was written to extend the functionality of report addon to sign
```
elicoidal: Hi @antespi 
Thanks for your contribution

"This module extends"

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,123 @@
...
+* Save as attachment: ``(object.number or '').replace('/','_') + '.signed.pdf'``
+
+**Note**: Linux user that executes Odoo server process must have read access
+to certificate file and password file
+
+
+Usage
+=====
+
+User just print PDF documents (only Qweb PDF reports supported) as usually,
```
elicoidal: s/print/prints
s/usually/usual

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,123 @@
...
+but signed PDF is automatically downloaded is this document model is configured
```
elicoidal: s/is this document/if this document

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,181 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import os
```
elicoidal, line 7: Please sort import in alphabetical order

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,181 @@
...
+                "No PDF certificate found for report '%s'" %
+                report.report_name)
+            return False
+        for cert_id in certificate_ids:
+            cert = m_cert.browse(cr, uid, cert_id, context=context)
+            # Check allow only one document
+            if cert.allow_only_one and len(ids) > 1:
+                _logger.info(
+                    "Certificate '%s' allows only one document, "
+                    "but printing %d documents" %
```
elicoidal: I dont understand this sentence

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,181 @@
...
+                continue
+            # Check domain
+            if cert.domain:
+                m_model = self.pool[cert.model_id.model]
+                domain = [('id', 'in', tuple(ids))]
+                domain = domain + eval(cert.domain)
+                doc_ids = m_model.search(cr, uid, domain, context=context)
+                if not doc_ids:
+                    _logger.info(
+                        "Certificate '%s' domain not satisfied" % cert.name)
```
elicoidal: s/satisfied/provided

report_qweb_signer/models/report_certificate.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
```
elicoidal: sort order

report_qweb_signer/models/report_certificate.py
```diff
@@ -0,0 +1,42 @@
...
+    domain = fields.Char(
+        string="Domain",
+        help="Domain for filtering if sign or not the document")
+    allow_only_one = fields.Boolean(
+        string="Allow only one document", default=True,
+        help="If True, this certificate can not be useb to sign "
+             "a PDF from several documents.")
+    attachment = fields.Char(
+        string="Save as attachment",
+        help="Filename used to store signed result as attachment. "
```
elicoidal: s/result/output

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,123 @@
...
+This module was written to extend the functionality of report addon to sign
+PDFs using a PKCS#12 certificate.
+
+
+Installation
+============
+
+To install this module, you need to install PortableSigner Java application:
+
+::
```
yajo: You can put that at the end of previous sentence, it's prettier: _[blabla] Java application::_

report_qweb_signer/__openerp__.py
```diff
@@ -0,0 +1,28 @@
...
+# © 2015 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "PDF reports signer",
+    "summary": "Sign PDFs usign a PKCS#12 certificate",
+    "version": "8.0.1.0.0",
+    "category": "Reporting",
+    "website": "http://www.antiun.com, "
+               "https://odoo-community.org",
```
Yajo: Only 1 URL is supported AFAIK.

report_qweb_signer/models/report.py
antespi: This is logging that certificate is configured for signing PDF from one document, but users is printing several documents in one PDF.
My english is not my best skill, feel free to suggest me a more understandable sentence.

report_qweb_signer/models/report.py
antespi: Here logging message is because domain configured is not returning any result. This means that this PDF must not be signed.
Domain is provided but not satisfied its condition.

yajo: cert.name should be an argument:

```
"Certificate '%s' domain not satisfied", cert.name)
```

That delegates string interpolation to the inner function, and skips it if not needed.

antespi: Good point!
I've just change this in all `_logger.info` callings

antespi:
I've just changed utility to sign PDF, by jPdfSign.
It's simpler than PortableSigner and depends only on iText v1.4.8.
Now jar files are in addon `static/jar` folder and it is only neccesary to install OpenJDK v1.7

@gurneyalex has already installed JRE in runbot, so we can get runbot green

yajo:
Then please add JRE to external_dependencies.

antespi:
@Yajo dependency added.
Also source code added, because I had to modify it for not prompting certificate password, just read from file

antespi:
I've just added a test certificate in order to test it in runbot. I can't add demo data without adding more dependencies, because this addon only depends on `report`. 

When runbot starts I'll configure a test PDF certificate for signing customer invoices.

yajo:
> I can't add demo data without adding more dependencies

Why?

antespi:
@Yajo, because there is no Qweb-PDF reports in `base` and `report` addons, only RML ones. This addon only applies to Qweb-PDF report type.

yajo:
Ah I understand.

pedrobaeza:
Create a fake/demo one for for example users.

antespi:
How? What's fake/demo means?

pedrobaeza:
Declare a mini report with no content or minimal content for that object.

antespi:
OK guys @gurneyalex @pedrobaeza @Yajo @rafaelbn, now you can fully test this addon in runbot. 
Go to `Sales > Customers > Agrolait` and click on `Print > Test PDF certificate`
If partner is a customer, then a signed PDF is downloaded. If partner is not a customer a normal PDF is dowloaded.

Happy review :smile:

You can install `document` addon in order to see attachment files saved

antespi:
cc @elicoidal

report_qweb_signer/models/report.py
elicoidal: "Certificate '%s' is only valid for one document, "
"while you try printing several (%d)" %

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,197 @@
...
+                continue
+            # Check domain
+            if cert.domain:
+                m_model = self.pool[cert.model_id.model]
+                domain = [('id', 'in', tuple(ids))]
+                domain = domain + eval(cert.domain)
+                doc_ids = m_model.search(cr, uid, domain, context=context)
+                if not doc_ids:
+                    _logger.info(
+                        "Certificate '%s' domain not satisfied", cert.name)
```
elicoidal, line 65: Certificate '%s' domain not satisfied
=> Certificate '%s' domain is incorrect 
?

rafaelbn:
- Print a signed PDF
- Delete from "Attachment" the PDF created
- Refresh screen
- Print again a signed PDF
- you get:

```


Odoo Server Error

  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/addons/report/controllers/main.py", line 121, in report_download
    response = self.report_routes(reportname, docids=docids, converter='pdf')
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/http.py", line 403, in response_wrap
    response = f(*args, **kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/addons/report_custom_filename/controllers/report_controller.py", line 34, in report_routes
    reportname, docids=docids, converter=converter, **data)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/http.py", line 403, in response_wrap
    response = f(*args, **kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/addons/report/controllers/main.py", line 65, in report_routes
    pdf = report_obj.get_pdf(cr, uid, docids, reportname, data=options_data, context=context)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3135849-31-6e943c/openerp/addons/report_qweb_signer/models/report.py", line 185, in get_pdf
    with open(signed, 'rb') as pf:
IOError: [Errno 2] No such file or directory: '/tmp/report.tmp.Sy1vEl.pdf.signed.pdf'
```

rafaelbn:
@antespi  one of the benefits and objectives of this module is that you can send an invoice to a Partner and this partner get the invoice (1) signed, can trust is yours (2) the pdf is protected and nobody can change it if he/she doesn't have your certificate.

In this case funcionality (2) is not present as I can modify the PDF signed.

Please @sergio-incaser take care on this pull as I see your invoices are protected and nobody can change them. Thanks

antespi:
I'm trying to figure out this bug, please do not review until next commit, thanks.

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+from openerp.tools.safe_eval import safe_eval as eval
+
+import logging
+_logger = logging.getLogger(__name__)
+
+
+def _normalize_filepath(path):
+    path = path or ''
+    path = path.strip()
+    if not path.startswith('/'):
```
gurneyalex: use os.path.isabs() -> it will work on windows too :-)

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+import base64
+from contextlib import closing
+import os
+import subprocess
+import tempfile
+import time
+
+from openerp import SUPERUSER_ID, models, api, _
+from openerp.exceptions import Warning, AccessError
+from openerp.tools.safe_eval import safe_eval as eval
```
gurneyalex: please don't rename safe_eval: it makes looking for unsafe 'raw' eval calls harder than needed when reviewing the code.

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+                _logger.info(
+                    "Certificate '%s' allows only one document, "
+                    "but printing %d documents",
+                    cert.name, len(ids))
+                continue
+            # Check domain
+            if cert.domain:
+                m_model = self.pool[cert.model_id.model]
+                domain = [('id', 'in', tuple(ids))]
+                domain = domain + eval(cert.domain)
```
gurneyalex: use safe_eval (explicitly, please). 

and consider using `domain += safe_eval(...)' to avoid the list copy.

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+    return path
+
+
+class Report(models.Model):
+    _inherit = 'report'
+
+    def _user_company_id_get(self, cr, uid, context=None):
+        m_users = self.pool.get('res.users')
+        user = m_users.browse(cr, SUPERUSER_ID, uid, context=context)
+        return user.company_id.id
```
gurneyalex: why not use:

`return self.pool['res.users']._get_company(cr, uid)`

pedrobaeza: And why in old API?

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def _certificate_get(self, cr, uid, ids, report, context=None):
+        if report.report_type != 'qweb-pdf':
+            _logger.info(
+                "Can only sign qweb-pdf reports, this one is '%s' type",
+                report.report_type)
+            return False
+        m_cert = self.pool.get('report.certificate')
```
gurneyalex: self.pool['report.certificate']

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+        certificate_ids = m_cert.search(cr, uid, [
+            ('company_id', '=', company_id),
+            ('model_id', '=', report.model)], context=context)
+        if not certificate_ids:
+            _logger.info(
+                "No PDF certificate found for report '%s'",
+                report.report_name)
+            return False
+        for cert_id in certificate_ids:
+            cert = m_cert.browse(cr, uid, cert_id, context=context)
```
gurneyalex: move the browse outside the loop

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def _attach_signed_read(self, cr, uid, ids, certificate, context=None):
+        if len(ids) != 1:
+            return False
+        filename = self._attach_filename_get(
+            cr, uid, ids, certificate, context=context)
+        if not filename:
+            return False
+        signed = False
+        m_attachment = self.pool.get('ir.attachment')
```
gurneyalex: self.pool['ir.attachment']

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def _attach_signed_write(self, cr, uid, ids, certificate, signed,
+                             context=None):
+        if len(ids) != 1:
+            return False
+        filename = self._attach_filename_get(
+            cr, uid, ids, certificate, context=context)
+        if not filename:
+            return False
+        m_attachment = self.pool.get('ir.attachment')
```
gurneyalex: self.pool['ir.attachment']

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,204 @@
...
+            attach_id = m_attachment.create(cr, uid, {
+                'name': filename,
+                'datas': base64.encodestring(signed),
+                'datas_fname': filename,
+                'res_model': certificate.model_id.model,
+                'res_id': ids[0],
+            })
+        except AccessError:
+            _logger.warning(
+                "Cannot save signed PDF '%s' as attachment", filename)
```
gurneyalex: the user will never see this.

report_qweb_signer/models/report.py
antespi: Right, I did because Odoo devs do, but I agree with you.

report_qweb_signer/models/report.py
antespi: I write in old API because I have to inherit from `get_pdf`, called from a controller and in v7 form.
I got issues when start in new API and then I go with old one. But maybe we can migrate before merge.
How can inherit from `get_pdf` method at `report` core addon, using new API?

antespi:
@gurneyalex all suggestions done, thanks for your review!.
@pedrobaeza still in old API, but running well.

rebased and squashed

report_qweb_signer/models/report.py
pedrobaeza: Use @api.v7 and @api.v8 for adding a compatibility layer

gurneyalex:
license of jpdffile is not clear to me. Not sure if you are allowed to modify it and redistribute it under AGPLv3.

gurneyalex:
should be ok (uses itext which is AGPL according to http://sourceforge.net/projects/itext/ so AGPL too). But needs to be written down with proper copyright attribution.

gurneyalex:
:+1:

rafaelbn:
:+1: tested in runbot and customer devel enviroment

rafaelbn:
This PR can be merge, 2 review and 8 days, thanks

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,111 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+==================
+PDF reports signer
```
pedrobaeza: This should be "QWeb PDF..."

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,111 @@
...
+==================
+
+This module extends the functionality of report addon to sign
```
pedrobaeza: s/report addon/_report_ module

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,111 @@
...
+  apt-get install openjdk-7-jre-headless
+
+
+Configuration
+=============
+
+In order to start signing PDF documents you need to configure certificate(s)
+to use in your company.
+
+* Go to ``Settings > Companies > Companies > Your company``
```
pedrobaeza, line 27: Why don't make this configuration at ir.actions.xml_report level? It makes more sense

report_qweb_signer/README.rst
```diff
@@ -0,0 +1,111 @@
...
+
+
+Usage
+=====
+
+User just prints PDF documents (only Qweb PDF reports supported) as usual,
+but signed PDF is automatically downloaded if this document model is configured
+as indicated above.
+
+If 'Save as attachment' is configured, signed PDF is saved as attachment and
```
pedrobaeza, line 52: If the configuration is on the report itself, there's no need of this extra field.

report_qweb_signer/models/report.py
```diff
@@ -0,0 +1,191 @@
...
+            signer, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+        out, err = process.communicate()
+        if process.returncode:
+            raise Warning(
+                _('Signing report (PDF)'),
+                _('jPdfSign failed (error code: %s). '
+                  'Message: %s') % (str(process.returncode), err))
+        return pdfsigned
+
+    @api.v7
```
pedrobaeza, line 151: This decorator is useless in your situation. For using it, you have to declare another method with the same name that is the one to be called when using new API, and this one is for calls from old API. If there's only one, you don't need both.

pedrobaeza:
I think your module deserves an specific icon :wink:

report_qweb_signer/models/report.py
antespi, line 151: But `report` model declares both:
- API v7: https://github.com/odoo/odoo/blob/9.0/addons/report/models/report.py#L137
- API v8: https://github.com/odoo/odoo/blob/9.0/addons/report/models/report.py#L246

That's why I write in old API, I don't know how to do it in API v8 and allow inheriting in controller: https://github.com/odoo/odoo/blob/9.0/addons/report/controllers/main.py#L45

report_qweb_signer/README.rst
antespi, line 52: Odoo Admin can select this also regardless report to support this use case:
`Save not signed PDFs, but sign every time this PDF is downloaded`

report_qweb_signer/models/report.py
pedrobaeza, line 151: OK, I see, thanks for the clarification. It's OK this way then.

report_qweb_signer/README.rst
pedrobaeza, line 52: But I don't think this would happen any time. If you want to store the documents, you want them signed, not unsigned.

report_qweb_signer/README.rst
antespi, line 27: Because there are several `ir.actions.xml_report` for the same model. And this object is not company dependent.

Here you select model (for every report who print it) an apply a signature.

I think this way we cover more flexible use cases.

pedrobaeza, line 27: I don't think so. You can add a field company_dependent on the ir.actions.xml_report, and it has more sense. For example, you can want to sign an invoice, but not an invoice list you print for reporting. Both goes again same model, but different reports.

antespi, line 27: There is a field to configure this `allow_only_one` in order to configure if want to sign or not for multiple docs in one PDF.

pedrobaeza, line 27: They can be complementary, but again I think this is better to have granularity at report level, not at model level.

report_qweb_signer/README.rst
antespi, line 52: You always get both, and this is a forced feature in order to respect inherit chain. That's why gives to Odoo Admin to configure this, not only if signed is saved, even the name of attachment.

pedrobaeza, line 52: OK, let's move with this and respect your current implementation.

antespi:
@pedrobaeza comments fixed or answered, thanks for your review. Happy weekend :smile: 

BTW, I don't know why here there's no travis or coveralls

pedrobaeza:
Trying to restart Travis this way.

pedrobaeza:
Please make a `git push -f` to see if we activate Travis.

antespi:
Done, let's see...

pedrobaeza:
Still no luck...

sergio-teruel:
Tested local instance and :+1: . Great job @antespi

antespi:
Hi @OCA/tools-maintainers, I think this addon is ready to be merged. Thanks for merging

rafaelbn:
:+1:

pedrobaeza:
:+1: after an internal talk and decided to leave this for not redoing all.

PR #41, opened 2016-01-27
[IMP] Define jPdfSign max heap size

antespi:
This PR maybe fix memory problems in runbot (and other servers too)

### DIFF
--- a/report_qweb_signer/models/report.py
+++ b/report_qweb_signer/models/report.py
@@ -116,8 +116,8 @@ class Report(models.Model):
             })
         except AccessError:
             raise Warning(
-                _('Saving signed report (PDF):') + '\n',
-                _('You do not have enought access rights to save attachments'))
+                _('Saving signed report (PDF): '
+                  'You do not have enought access rights to save attachments'))
         else:
             _logger.info(
                 "The signed PDF document '%s' is now saved in the database",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

antespi:
jejeje, it runs well !! I got it!

antespi:
@gurneyalex this PR fix memory problem at runbot in this addon, please review it, thanks

carlosdauden:
:+1:

sergio-teruel:
:+1:

report_qweb_signer/models/report.py
```diff
@@ -145,7 +147,8 @@ def pdf_sign(self, pdf, certificate):
             raise Warning(
                 _('Signing report (PDF)'),
                 _('jPdfSign failed (error code: %s). '
-                  'Message: %s') % (str(process.returncode), err))
+                  'Message: %s. Output: %s') %
+                (str(process.returncode), err, out))
```
yajo: [`str()` is implicit with `%s`](https://docs.python.org/2/library/stdtypes.html#string-formatting-operations) AFAIK.

pedrobaeza: Yeah, that is

rafaelbn:
:+1:

report_qweb_signer/models/report.py
```diff
@@ -145,7 +147,8 @@ def pdf_sign(self, pdf, certificate):
             raise Warning(
                 _('Signing report (PDF)'),
```
pedrobaeza: Warning only uses one argument, not 2 as the old `except_orm`

antespi: But It's not what i see here: https://github.com/odoo/odoo/blob/8.0/openerp/exceptions.py#L33
isn't it?

report_qweb_signer/models/report.py
pedrobaeza: Warning inherits from Exception, not from except_orm

report_qweb_signer/models/report.py
pedrobaeza: Warning inherits from Exception, not from except_orm

antespi: After a depth looking... you're right :smile:

antespi:
@Yajo, @pedrobaeza fixes suggested done, thanks.

yajo:
:+1:

PR #42, opened 2016-02-11
[FIX] [report_xls] Override the field to add the value instead of breaking the validation system

Garamotte:
I think this is better to override the field in the new API way, to really add a possible value in the selection, instead of "breaking" the fields's value validation system.

This also allows Odoo to display the "XLS" value in the reports list, instead of "unknown".

### DIFF
--- a/report_xls/ir_report.py
+++ b/report_xls/ir_report.py
@@ -20,15 +20,10 @@
 #
 ##############################################################################
 
-from openerp import models
+from openerp import models, fields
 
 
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
:+1:, thanks!

lasley:
Thanks @sylvain-garancher - Can you please rebase & fix the conflicts?

Garamotte:
Rebase done.

lasley (approved):
Thanks @sylvain-garancher

PR #44, opened 2016-02-15
[SEC] report_xls: fix unsafe eval

gurneyalex:


lmignon:
:+1:

pedrobaeza:
The same concern about 9.0 patches in non migrated modules, but :+1:

antespi:
:+1:

PR #45, opened 2016-02-15
[SEC] report_xls: fix unsafe eval

gurneyalex:


lmignon:
:+1:

pedrobaeza:
:+1: 

No Travis?

gurneyalex:
@pedrobaeza feel free to propose another PR for this :smile_cat:

pedrobaeza:
Travis is already defined, so that's why I wonder what's the problem...

antespi:
:+1:

gurneyalex:
@pedrobaeza then it's probably just a temporary overload, I'm being throttled...

sebastiken:
Hi, @gurneyalex. After this merge, one of my reports using _report_xls_ is failing with this error:

```
  File "/opt/buildout/reporting_engine/report_xls/report_xls.py", line 166, in render
    row[i] = safe_eval(row[i], render_space)
  File "/opt/buildout/parts/odoo/openerp/tools/safe_eval.py", line 249, in safe_eval
    raise TypeError("safe_eval does not allow direct evaluation of code objects.")
TypeError: safe_eval does not allow direct evaluation of code objects.
```

Should I modifiy something in the report?
Thanks

gurneyalex:
we will need to revert this, I'm afraid, as the patch obviously breaks things.

pedrobaeza:
But I think that's one of the goals of safe_eval: don't allow risky operations, like direct evaluation of objects.

gurneyalex:
@pedrobaeza the issue is that what is being passed to `eval` is not a string, but compiled bytecode (see https://docs.python.org/2/library/functions.html?highlight=compile#compile) which `safe_eval` does not handle correctly. 

I'm currently nose deep in a customer project, and I won't be able to fix this in the coming days. 

A possible solution could be to change the _render function in `utils.py` to this:

```
def _render(code):
    compile(code, '<string>', 'eval')
    return code
```

-> we attempt compiling the code to get an error if there is a syntax issue, and then return the string to be processed by `safe_eval`. However I think this will fail because the evaluation context will not be passed. 

@luc-demeyer do you have an opinion on this?

pedrobaeza:
OK, I see, I haven't used report_xls directly, so I can't say without deeper inspect.

gurneyalex:
I just reverted the change in https://github.com/OCA/reporting-engine/commit/551c2b17753b8be9cab87ffd470992f2875cab0b

cc @sebastiken

gurneyalex:
example use of _render with report_xls : https://github.com/OCA/account-financial-reporting/blob/9.0/account_journal_report_xls/report/nov_account_journal_xls.py#L84

I believe the code in there is not vulnerable because we are not evaluating user supplied strings.

luc-demeyer:
This patch should indeed be reverted.

sebastiken:
Thanks @gurneyalex !

PR #46, opened 2016-02-19
[NEW][8.0] report_header_full_width: More header width for company address.

sergio-teruel:
cc. @rafaelbn


rafaelbn:
:+1:  thanks.

pedrobaeza:
Why don't you propose a patch to Odoo? This is clearly a non-desired feature (if you don't want to talk about a bug).

sergio-teruel:
Ok.. I think the same as you... so i close this PR and propose to odoo directly.

sergio-teruel:
PR proposed to odoo @rafaelbn @pedrobaeza 
https://github.com/odoo/odoo/pull/11031

PR #52, opened 2016-05-10
9.0 Migrate report_custom_filename

sla-subteno-it:
Hello,

I migrated report_custom_filename in v9 but I used the same format of attachement field instead of mako template.


report_custom_filename/controllers/report_controller.py
```diff
@@ -38,18 +38,11 @@ def report_routes(self, reportname, docids=None, converter=None, **data):
...
                 .browse(docids or [])
-            generated_filename = email_template.mako_template_env\
-                .from_string(report.download_filename)\
-                .render({
-                    'objects': objects,
-                    'o': objects[:1],
-                    'object': objects[:1],
-                    'ext': report.report_type.replace('qweb-', ''),
-                })
+            generated_filename = eval(report.download_filename, {'object': objects, 'time': time})
```
Garamotte: You should use `from openerp.tools.safe_eval import safe_eval` instead of `eval`.

pedrobaeza:
Your migration is ripping out the basic things of the module, so the fail in Travis is expected. Are you sure this is working?

andreparames:
:+1:  I've tested this PR and seems to work fine, at least for the regular use case (document → print). Not sure if the removed controller is needed for some other use case, but Travis seems to be fine now.

report_custom_filename/controllers/report_controller.py
```diff
@@ -38,18 +39,12 @@ def report_routes(self, reportname, docids=None, converter=None, **data):
...
                 .browse(docids or [])
-            generated_filename = email_template.mako_template_env\
-                .from_string(report.download_filename)\
-                .render({
-                    'objects': objects,
-                    'o': objects[:1],
-                    'object': objects[:1],
-                    'ext': report.report_type.replace('qweb-', ''),
-                })
+            generated_filename = safe_eval(report.download_filename,
```
pedrobaeza, line 46: safe_eval uses Python syntax, not Jinja2 syntax, so it changes from the way Odoo parses this in the attachments.

ovnicraft:
any updates ?
just for the record if wkhtmltopdf is not installed, this module return a html file with pdf extension, but i think need a bug report.

pedrobaeza (changes requested):
There's a pending comment from my part

pedrobaeza:
Closing as there's no answer.

PR #53, opened 2016-05-12
fix(reporting-engine/report_custom_filename): Agregated try except be…

AlexandroTapia:
…cause show a attributeError

Aggregated try except because show the error: 'ir.actions.report.xml' object has no attribute
'generate_filename'

### DIFF
--- a/report_custom_filename/controllers/reports.py
+++ b/report_custom_filename/controllers/reports.py
@@ -32,9 +32,13 @@ class Reports(main.Reports):
         context = dict(req.context)
         context.update(action["context"])
         report_xml = req.session.model('ir.actions.report.xml')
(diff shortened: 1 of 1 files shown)


yvaucher:
It doesn't seems to be the right fix. Either the method has been renamed or removed. Thus it would be better to fix it by using the right method or to remove this part of code if it is deprecated.

Plus in a try catch it is better to use `logger` than print to add it in logs if it must be treated as an exception that we can't display to the user.

:-1:

StefanRijnhart:
Can you add a test demonstrating the problem? i.e. a test that fails without this change? (asking because I too fail to see the use of this change at first instance)

Issue #56, opened 2016-06-01
8.0 Crash in report_custom_filename


alexis-via:
Here is the scenario to reproduce the bug on an up-to-date Odoo v8 with up-to-date OCA/reporting-engine/8.0 branch:
1) Install the module report_custom_filename
2) Replace the native sale order report by an aeroo report (I guess the problem is the same with reporting engines other than Qweb)

To make it easier for you to reproduce the bug, I have prepared a small module "test_bug_reportfilename" that depend on report_aeroo and report_custom_filename and that replaces the native quotation report by a small aeroo report. Get it here:
http://people.via.ecp.fr/~alexis/test_bug_reportfilename.zip

3) In Settings > Technical > Actions > Report
Edit the report "Quotation / Order" and set the field "Download filename" to
Quote-currency-${o.currency_id.name}.pdf
This filename may be stupid, but it's easier to reproduce/understand the bug that way, with a field o.currency_id with is present on sale.order but not on res.partner.
4) Go to the Form view of the customer Agrolait, click on the "Sales" button, select the first quotation in the tree view and click on "Print": you get a crash:

```
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo8/addons/web/controllers/main.py", line 70, in wrap
    return f(*args, **kwargs)
  File "/home/alexis/new_boite/dev/reporting-engines8/report_custom_filename/controllers/reports.py", line 49, in index
    'object': objects[0],
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 1, in top-level template code
  File "/usr/lib/python2.7/dist-packages/jinja2/sandbox.py", line 329, in getattr
    value = getattr(obj, attribute)
UndefinedError: 'openerp.api.res.partner object' has no attribute 'currency_id'
```

**Code analysis:**

Qweb reports use report_custom_filename/controllers/report_controller.py
Aeroo reports (and probably other report engines) use report_custom_filename/controllers/reports.py

[...]

PR #57, opened 2016-06-01
8.0 FIX bug #56 crash in report_custom_filename

alexis-via:
Please read the bug report #56 that has a complete scenario to reproduce the bug and the analysis of the code.

### DIFF
--- a/report_custom_filename/controllers/reports.py
+++ b/report_custom_filename/controllers/reports.py
@@ -39,7 +39,14 @@ class Reports(main.Reports):
         for report in report_xml.browse(report_ids):
             if not report.download_filename:
                 continue
-            objects = http.request.session.model(context['active_model'])\
+            # When you are on the form view of a customer and you
+            # click on the "Sales" button, context['active_model']
+            # is still res.partner (probably a bug...) but
+            # context['active_ids'] has the ID of the sale order
+            # cf https://github.com/OCA/reporting-engine/issues/56
+            # That's why we use report.model instead of context['active_model']
+            # together with context['active_ids']
+            objects = http.request.session.model(report.model)\
                 .browse(context['active_ids'])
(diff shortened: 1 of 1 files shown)


report_custom_filename/controllers/reports.py
```diff
@@ -39,7 +39,14 @@ def index(self, action, token):
...
-            objects = http.request.session.model(context['active_model'])\
+            # When you are on the form view of a customer and you
+            # click on the "Sales" button, context['active_model']
+            # is still res.partner (probably a bug...) but
+            # context['active_ids'] has the ID of the sale order
+            # cf https://github.com/OCA/reporting-engine/issues/56
+            # That's why we use report.model instead of context['active_model']
+            # together with context['active_ids']
+            objects = http.request.session.model(report.model)\
                 .browse(context['active_ids'])
```
hbrunn: thanks for the ping, I think what you do is the right thing, and it's coincidence that the active_ids thing works very often. Can you change this line too to get the report's object ids? Cf https://github.com/OCA/OCB/blob/8.0/addons/web/controllers/main.py#L1574

alexis-via: @hbrunn I can't, because I don't have any 'datas' key, not 'ids' key inside the action dict, nor any other information where I could get the ID of the object for the report except in context['active_ids']

hbrunn: ah, what they do is using active_ids if those keys don't exist. Can you maybe say
`action.get('datas', {}).get('ids', context['active_ids'])`? Then this should behave the exact same way as `super`

alexis-via:
I made the change

pedrobaeza:
:+1:

hbrunn:
:+1:

yvaucher:
:+1:

PR #59, opened 2016-07-08
[9.0][MIG][bi_view_editor] Ported BI View Editor module

tarteo:


bi_view_editor/README.rst
```diff
@@ -0,0 +1,100 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/143/9.0
+
+Known issues / Roadmap
+======================
+
+* Non-stored fields are not supported yet
+* Provide graph view for table relations (would be a "nice to have")
+* Porting to Odoo 9.0
```
elicoidal: I think you can safely remove this sentence

bi_view_editor/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/reporting-engine/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
```
elicoidal: Remove following latest README template

bi_view_editor/__openerp__.py
```diff
@@ -0,0 +1,34 @@
...
+        'templates/assets_template.xml',
+        'views/bve_view.xml',
+    ],
+    'qweb': [
+        'templates/qweb_template.xml',
+    ],
+    'js': [
+        'static/src/js/bve.js'
+    ],
+    'demo': [],
```
elicoidal: remove empty keys

bi_view_editor/models/ir_model.py
```diff
@@ -0,0 +1,223 @@
...
+        res = super(IrModel, self).create(cr, user, vals, context)
+        if vals.get('state', 'base') == 'bve':
+            vals['state'] = 'manual'
+
+            # add model in registry
+            self.instanciate(cr, user, vals['model'], context)
+            self.pool.setup_models(cr, partial=(not self.pool.ready))
+
+            # update database schema
+            # model = self.pool[vals['model']]
```
elicoidal: Remove dead code

bi_view_editor/models/ir_model.py
```diff
@@ -0,0 +1,223 @@
...
+            # ctx = dict(
+            #     context,
+            #     field_name=vals['name'],
+            #     field_state='manual',
+            #     select=vals.get('select_level', '0'),
+            #     update_custom_fields=True)
+            RegistryManager.signal_registry_change(cr.dbname)
+
+#        self.write(cr, user, [res], {'state': 'manual'})
```
elicoidal: idem

bi_view_editor/models/ir_model.py
```diff
@@ -0,0 +1,223 @@
...
+        q = ("""UPDATE ir_model SET state = 'manual'
```
elicoidal: why use SQL statement?
Shouldnot you use ORM functions?

bi_view_editor/templates/qweb_template.xml
```diff
@@ -0,0 +1,78 @@
...
+            <p>Please choose the join node</p>
+            <div id="join-nodes">
+            </div>
+        </div>
+    </t>
+
+
+    <t t-name="BVEEditor">
+        <div class="oe_form_field_bi_editor">
+          <!--
```
elicoidal: remove dead code

bi_view_editor/templates/qweb_template.xml
```diff
@@ -0,0 +1,78 @@
...
+                    </table>
+                </div>
+                <div class="clear"></div>
+            </div>
+            <div class="footer">
+                <div class="left"></div>
+                <div class="right"><button id="clear"><span class="fa fa-eraser"></span> Clear</button></div>
+            </div>
+            <ul class="context-menu">
+                <!--
```
elicoidal: idem

elicoidal:
@tarteo Thanks for the contribution!
Can you try to increase coverage with Unit test?

bi_view_editor/models/ir_model.py
astirpe: The commented line (write method) is not working anymore, as in Odoo V9 a new orm constraint is restricting the modification of the state while updating ir.model.
That's the reason of using the SQL statement.

elicoidal: Thanks for the reply: my technical level is not good enough here to challenge!
Is there any workaround: @pedrobaeza @dreispt ?

JordiBForgeFlow:
@lreficent can you review and add test cases?

LoisRForgeFlow:
I've been trying to do some test but it seems to be a little complex, I am not being able to add proper data to test the action_create() method. Can you help us @tarteo?

astirpe:
@lreficent we experienced such issues too. It seems that creating a model while running tests raises some errors. Is that the case? Are you experiencing the same?

JordiBForgeFlow:
@astirpe yes. I think that it is not a good practice to introduce cr.commit here: https://github.com/OCA/reporting-engine/pull/59/files#diff-824d3c4f48b7f296e915d38877993ceaR274

Perhaps there are other ways to accomplish the same thing. Perhaps @hbrunn knows.

astirpe:
@jbeficent the issue is not the use of cr.commit, as you can notice we removed it while porting to V10 [here](https://github.com/onesteinbv/addons-onestein/tree/10.0/bi_view_editor). The problem is when creating a manual model during the tests.

Probably it is useful if you take a look at the travis failures while we were working on it [https://github.com/onesteinbv/addons-onestein/commits/10.0/bi_view_editor](https://github.com/onesteinbv/addons-onestein/commits/10.0/bi_view_editor). You can see a long list of attempts, about one month ago.

With version 9 we had the same issues as with V10.

hbrunn:
@astirpe is this similar to the model purging tests in database_cleanup? After a lot of pain and misery, we decided to run these tests in its own run, because we couldn't figure out how to mess with the registry in tests and afterwards clean up in a way that doesn't affect the other modules: https://github.com/OCA/server-tools/blob/9.0/.travis.yml#L24-L27

astirpe:
@hbrunn I will check the database_cleanup module tomorrow. But I can already say that I'm quite sure that running the tests on its own env will solve this kind of issues. This is the case we've encountered: after running the bi_view_editor tests (successfully), the subsequent module being tested encounters gets troubles.

hbrunn:
this was exactly the issue over there. And despite being theoretically and technically interesting to find out how to do this, I think after dedicating a couple of tens of hours to that without finding a good solution, I think it's okay to admit defeat for now

JordiBForgeFlow:
@hbrunn @astirpe OK, so can we merge this one now then?

astirpe:
@jbeficent I would like to update the code with a refactoring we already made. I will do it today, as soon as possible. After that I would be very happy if this PR is merged. Thanks to you all!

astirpe:
I've made a commit and of course I was expecting a failure of the coverage, since no tests are present. But here codecov says "No report found to compare against". Also runbot seems not responding, I just clicked on "Force Rebuild" button...

In the meanwhile does someone want to take a look at the latest changes in the code?

astirpe:
After understanding the job made for the tests in database_cleanup, finally I got travis successfully running with green unit tests! Special thanks to @hbrunn to mention that example!

Still I see codecov status "No report found to compare against". Still runbot not responding.
Does someone know the reason?

LoisRForgeFlow (approved):
LGTM now, the test were ran correctly in v10 btw. Thank you all!

JordiBForgeFlow:
@astirpe in this PR I introduced the tree view as a possible visualization option. At this moment the columns that appear in the list view are the ones of the BI model that you defined with the option "Row".

But in reality that is not 100% correct. The options "Row", "Column", "Measure" are intended to be used to define the output of the Pivot view. 

Perhaps it would be better to have a flag 'List', that you can check on each field that you want to appear in the tree view. Perhaps even this flag should be set by default.

astirpe:
@jbeficent you're right! I could work on that next week or, if you want, you can open a PR that I will merge quickly. Thank you!

JordiBForgeFlow:
@astirpe I tried to do but did not succeed myself.

astirpe:
Rebased, in order to restore the green status for runbot and codecov..

JordiBForgeFlow:
@astirpe Now green!

astirpe:
@jbeficent @lreficent With the latest commit we have the selection of the fields to be displayed in the tree view. Could you maybe take a look at it?

astirpe:
@lasley may I ask you to update the label of this PR, from "needs fixing" to "needs review"?

elicoidal (changes requested):
I have tested on runbot with 2 feedback
* usability of cross model fields is misleading
* a little glitch with the following example.

Test:
* Added Currency + currency rate
* At that moment a popup windows appears asking for options. 

*Issue 1*: The usability is not obvious because user dont know whether to click on the link or on button OK (but we have no radio buttons in Odoo ...).
In my opinion the button should not be call OK but Cancel
Another option would be to use checkbox to emulate the radio buttons (along with a third option "None of the above"). 
Still those are simple suggestion about improving the usability and none blocking.

*Issue 2*: The glitch actually is that if you click on the OK button, the selected field is not added to the view (which is OK) but also disappears from the list on the left.
You have to save the view and re-enter the form to have the field back.

astirpe:
@elicoidal Thank you!

I've just committed the fix for _Issue1_, improving the user experience and making strings translatable. There is also a bit of js refactoring, based on the job already made while porting to V10. Would you like to see the new look&feel?

About _Issue2_, I tried to reproduce the glitch you described but it doesn't seem to occur to me. If I select a field and then I close the popup, that field remains at its place. Or maybe I misunderstood the issue?

elicoidal:
@astirpe 
Thanks for fixing issue1! Much clearer now :)
I have recorded the following to show the issue2:
bi_editor_disappearing_filed

astirpe:
@elicoidal thanks a lot!
Now I can understand the problem: you selected fields by drag&drop!
I was selecting fields by just clicking on the fields, that's why I was not able to reproduce the issue.

Working on it...

elicoidal:
> Now I can understand the problem: you selected fields by drag&drop!

My bad :p! Users are so unpredictable... ;)

astirpe:
@elicoidal I think this simple fix is enough, may I ask you check again?

elicoidal (approved):

elicoidal:
:+1:

bi_view_editor/README.rst
```diff
@@ -0,0 +1,93 @@
...
+  including tree, graph, pivot views.
+
+
+Usage
+=====
+
+
+To graphically design your analysis data-set:
+
+- From the Reporting menu, select "Custom BI Views"
```
JordiBForgeFlow: From the Dashboards menu

bi_view_editor/README.rst
```diff
@@ -0,0 +1,93 @@
...
+- Browse trough the business objects in the Query tab
+- Pick the interesting fields (Drag & Drop)
+- For each selected field, right-click on the Options column and select whether it's a row, column or measure
```
JordiBForgeFlow: If you want to remove the field from the list view, unflag the checkbox ´List´ in the Options column.

bi_view_editor/models/bve_view.py
```diff
@@ -0,0 +1,389 @@
...
+
+    state = fields.Selection(
+        [('draft', 'Draft'),
+         ('created', 'Created')],
+        default='draft',
+        copy=False)
+    data = fields.Text(
+        help="Use the special query builder to define the query "
+             "to generate your report dataset. "
+             "NOTE: Te be edited, the query should be in 'Draft' status.")
```
JordiBForgeFlow: typo: 'NOTE: To be edited'

bi_view_editor/README.rst
```diff
@@ -0,0 +1,93 @@
...
+- Save and click "Generate BI View"
+- Click "Open BI View" to view the result
+- If module Dashboard (board) is installed, the standard "Add to My Dashboard" functionality would be available
+- Click "Create a menu" to create a new menu item directly linked to your new BI view (this feature is available in developer mode)
```
JordiBForgeFlow: When the BI view is reset back to draft this menu will be removed, and you will need to re-create the menu entry.

JordiBForgeFlow (approved):
:+1: Functional review

astirpe:
Yeeeeeee !!! Thank you all for your reviews, suggestions and interest !!!

JordiBForgeFlow:
Thank you for this excellent contribution!!!

PR #60, opened 2016-07-19
[ADD] added add-on pdf_compression / pdfa

Benniphx:
This addon provides the ability to convert outgoing and imported PDF attachments. You can compress the PDFs to save space in you database or file system. Also the possibility exists to convert PDFs in PDF/A

moved pull request from https://github.com/OCA/server-tools/pull/474

### DIFF
--- a/pdf_compression/README.rst
+++ b/pdf_compression/README.rst
@@ -1,0 +1,115 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=======================
+PDF Compression & PDF/A
+=======================
+
+This addon provides the ability to convert outgoing and imported PDF attachments.
+You can compress the PDFs to save space in you database or file system.
+Also the possibility exists to convert PDFs in PDF/A. Currently implemented:
+
+* PDF/A-1b
+* PDF/A-2b (recommend to use this setting because it is faster and result in smaller files)
+
+
+Installation
+============
+
+#. You need at least Gostscript 9.18 (I recommend 9.19)
+   `http://ghostscript.com <http://ghostscript.com/download/gsdnld.html>`_
+
+
+Configuration
+=============
+
+If the odoo module is installed:
+
+#. You can define ``Compession level`` and ``PDF/A TYPE`` via
+   `Settings -> Configuration -> Knowledge`.
+#. Also you can override PDF/A option for individual Reports apart from Global Settings via
+   `Settings -> Technical -> Reports -> Reports -> Select a Report -> PDF/A archiving`.
+
+
+
+Usage
+=====
+
+
+#. Compression rate outcome:
+
+   I uploaded a 1MB PDF file with a mix of pictures tables and text.
+
+   Settings:
+
+   * screen: selects low-resolution output similar to the Acrobat Distiller "Screen Optimized" setting.
+   * ebook: selects medium-resolution output similar to the Acrobat Distiller "eBook" setting.
+   * printer: selects output similar to the Acrobat Distiller "Print Optimized" setting.
+   * prepress: selects output similar to Acrobat Distiller "Prepress Optimized" setting.
+   * default: selects output intended to be useful across a wide variety of uses, possibly at the expense of a larger output file.
+
+   Examples:
+
+   * screen: 1MB -> 200KB some quality loss notice (jpg artefacts in pictures)
+   * ebook: 1MB -> 540KB little quality loss notice (some jpg artefacts in pictures but way better then screen)
+   * printer: 1MB -> 740KB no quality loss notice
+   * prepress: 1MB -> 820KB  no quality loss notice
+   * default: 1MB -> 680KB  little quality loss notice (some jpg artefacts in pictures but way better then screen)
+
+#. PDF/A conversion:
+
+   `Info wiki page <https://en.wikipedia.org/wiki/PDF/A>`_
+
+   PDFs will be converted to:
+
+   * PDF/A-1b
+   * PDF/A-2b
+
+
+
+Known issues / Roadmap
+======================
+
+* Had to override the _run_wkhtmltopdf Function to pass context through it
+* Implementing other PDF/A options
+* Future Goal to integrate ZUGFeRD
+* only PDFs which are saved as attachments are converted
+* (if you use the print button to generate pdf the "first" time the compession/pdfa is not used but the saved attachment is converted)
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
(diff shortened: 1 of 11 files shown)


pdf_compression/README.rst
```diff
@@ -0,0 +1,115 @@
...
+Also the possibility exists to convert PDFs in PDF/A. Currently implemented:
+
+* PDF/A-1b
+* PDF/A-2b (recommend to use this setting because it is faster and result in smaller files)
+
+
+Installation
+============
+
+#. You need at least Gostscript 9.18 (I recommend 9.19)
```
pedrobaeza, line 20: s/Gostscript/Ghostscript

pdf_compression/README.rst
```diff
@@ -0,0 +1,115 @@
...
+   `http://ghostscript.com <http://ghostscript.com/download/gsdnld.html>`_
+
+
+Configuration
+=============
+
+If the odoo module is installed:
+
+#. You can define ``Compession level`` and ``PDF/A TYPE`` via
```
pedrobaeza, line 29: s/Compession/Compression

pdf_compression/README.rst
```diff
@@ -0,0 +1,115 @@
...
+   `Settings -> Configuration -> Knowledge`.
```
pedrobaeza, line 30: Why not in something related to reporting?

pdf_compression/README.rst
```diff
@@ -0,0 +1,115 @@
...
+
+   * PDF/A-1b
+   * PDF/A-2b
+
+
+
+Known issues / Roadmap
+======================
+
+* Had to override the _run_wkhtmltopdf Function to pass context through it
```
pedrobaeza, line 74: Isn't a cleaner way to do it?

pdf_compression/README.rst
```diff
@@ -0,0 +1,115 @@
...
+#. Also you can override PDF/A option for individual Reports apart from Global Settings via
+   `Settings -> Technical -> Reports -> Reports -> Select a Report -> PDF/A archiving`.
+
+
+
+Usage
+=====
+
+
+#. Compression rate outcome:
```
pedrobaeza, line 40: Explain here that you have to print a report that is saved as attachment, and the attachment will be compressed. I understand that the save prompt is for the non compressed PDF. Isn't it?

pdf_compression/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': "PDF Compression",
+    'summary': """compresses the pdf files for ir.attachement""",
+    'author':
+        "Benjamin Bachmann ,"
```
pedrobaeza, line 7: Individual names are not put here, but on contributors section of README.

pdf_compression/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+        "bloopark systems GmbH & Co. KG ,"
+        "Odoo Community Association (OCA)",
+    'website': "http://www.bloopark.de",
+    'category': 'attachement',
```
pedrobaeza, line 11: Better "Reporting"

pdf_compression/i18n/de.po
```diff
@@ -0,0 +1,136 @@
...
+
+#. module: pdf_compression
+#: view:knowledge.config.settings:pdf_compression.view_knowledge_configuration_pdfa
+msgid "Configure Attachment"
+msgstr "Konfiguriere Dateianhänge"
+
+#. module: pdf_compression
+#: field:knowledge.config.settings,create_uid:0
+msgid "Created by"
+msgstr "Created by"
```
pedrobaeza, line 53: If not translated, leave it empty instead of the original English string

pdf_compression/README.rst
pedrobaeza, line 20: You can put the ghostscript executable as external dependency in the manifest file.

pdf_compression/models/ir_attachment.py
```diff
@@ -0,0 +1,171 @@
...
+            with NamedTemporaryFile(delete=False, suffix='.pdf') as f1:
+                f1.write(base64.decodestring(pdf))
+
+            f2 = NamedTemporaryFile(delete=False, suffix='.pdf')
+            f2.close()
+
+            input_file = f1.name
+            output_file = f2.name
+
+            if pdfa_option != 'no' and (report_override == 'default' or
```
lasley, line 99: I don't see `pdfa_option` or `report_override` changing between this `if` and the previous, so you could probably remove the first two conditionals because we already have confirmed above.

pdf_compression/models/ir_attachment.py
```diff
@@ -0,0 +1,171 @@
...
+                        '-dProcessColorModel=/DeviceRGB',
+                        '-dColorConversionStrategy=/RGB',
+                        '-dEmbedAllFonts=true',
+                        '-dMaxSubsetPct=100',
+                        '-dSubsetFonts=true'])
+
+            if compression == 'no' and (pdfa_option == 'pdfa1b' or
+                                        report_override == 'pdfa1b'):
+                local_command_args.extend(['-dPDFSETTINGS=/default'])
+            elif compression == 'prepress':
```
lasley, line 136: These elif blocks could be simplified by using string interpolation for the settings:

```
if compression in ['prepress', ...]: 
    local_command_args.extend(['-dPDFSETTINGS=/%s' % compression])
```

pdf_compression/models/ir_attachment.py
```diff
@@ -0,0 +1,171 @@
...
+                if err:
+                    _logger.info('ghostscript failed . Message: %s' % err)
+
+                with open(output_file, 'rb') as f3:
+                    pdf = base64.encodestring(f3.read())
+
+                os.unlink(input_file)
+                os.unlink(output_file)
+
+            except (OSError, IOError, ValueError):
```
lasley, line 168: IMO the try/except:raise is unnecessary unless we're doing something before the raise.

pdf_compression/models/report.py
```diff
@@ -0,0 +1,203 @@
...
+    def get_pdf(self, cr, uid, ids, report_name, html=None, data=None,
+                context=None):
+        """override to add pdf_option to the context"""
+        ctx = dict(context)
+        report = self._get_report_from_name(cr, uid, report_name)
+        ctx.update({'pdf_option': report.pdf_option})
+        return super(Report, self).get_pdf(
+            cr, uid, ids, report_name, html=html, data=data, context=context)
+
+    def _run_wkhtmltopdf(self, cr, uid, headers, footers, bodies, landscape,
```
lasley, line 44: This looks to be identical to the core method - https://github.com/odoo/odoo/blob/8.0/addons/report/models/report.py#L370

lasley:
Thanks for the awesome new module @Benniphx ! I have added some comments inline.

Also, we're adding a decent amount of core logic here without test coverage. Would you mind adding some more tests too please?

pdf_compression/README.rst
lasley, line 74: Oh I see, I made a note about this in the code. @Benniphx can you explain a little bit more about what the issue was here?

PR #62, opened 2016-07-25
[8.0][FIX][report_xls] Protect xlwt import.

yajo:
Given `xlwt` could happen to not be installed because it is an optional dependency.

See https://github.com/odoo/odoo/issues/12897#issuecomment-234928120.

@Tecnativa

### DIFF
--- a/report_xls/__init__.py
+++ b/report_xls/__init__.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 #    OpenERP, Open Source Management Solution
@@ -22,4 +22,3 @@
 
 from . import ir_report
 from . import report_xls
-# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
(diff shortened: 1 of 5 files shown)


hbrunn:
:+1:

pedrobaeza:
:+1:

carlosdauden:
:+1:

pedrobaeza:
Please correct Travis now that the module has changed. In concrete, the line https://travis-ci.org/OCA/reporting-engine/jobs/147163836#L459. You should bypass the error with `# pylint: disable=W0123` in the line, because `eval` is needed by some of the modules as I remember.

sergiocorato:
Hi,
after this merge this import 
from openerp.addons.report_xls.report_xls import report_xls
doesn't fail?

pedrobaeza:
What are you talking about?

sergiocorato:
https://github.com/OCA/account-financial-reporting/blob/8.0/mis_builder/report/mis_builder_xls.py
row 7

sergiocorato:
*row

pedrobaeza:
That's something totally different (in another repo). That should be protected too.

sergiocorato:
I refer to report_xls class name, renamed to ReportXls, which is not more recognized

pedrobaeza:
OK, please fill a PR with that change.

sergiocorato:
#63

StefanRijnhart:
More breakage in account-financial-tools. https://github.com/OCA/account-financial-tools/pull/392

PR #63, opened 2016-07-26
[FIX][8.0] revert class renamed report_xls

sergiocorato:
Reverts class renamed from report_xls to ReportXls as it will break all modules which import this class

### DIFF
--- a/report_xls/report_xls.py
+++ b/report_xls/report_xls.py
@@ -44,7 +44,7 @@ class AttrDict(dict):
         self.__dict__ = self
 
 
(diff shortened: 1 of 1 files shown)


yajo:
Although it does not follow the guidelines, it's true that it could break things, so :+1:, but I'd prefer the CamelCase name for future versions.

StefanRijnhart:
:+1: This breaks all of @luc-demeyer's modules ;-) Let's leave this refactoring until a major upgrade.

pedrobaeza:
Fast-tracking

PR #64, opened 2016-08-03
[FIX][report_xls] Protect import.

yajo:
Even after merging 2bf93a1d49f7c79062ab7c237ceb794cdbdec983, import still breaks when trying to use module's stuff at class definition time when module is not imported.

I move xsl types to a failure-safe scope.

I checked that no other modules use these variables.

@Tecnativa

### DIFF
--- a/report_xls/report_xls.py
+++ b/report_xls/report_xls.py
@@ -29,13 +29,29 @@ from types import CodeType
 from openerp.report.report_sxw import report_sxw
 from openerp import pooler
 import logging
+
 _logger = logging.getLogger(__name__)
+
+xls_types_default = {
+    'bool': False,
+    'date': None,
+    'text': '',
+    'number': 0,
+}
 
 try:
     import xlwt
     from xlwt.Style import default_style
+
+    xls_types = {
+        'bool': xlwt.Row.set_cell_boolean,
+        'date': xlwt.Row.set_cell_date,
+        'text': xlwt.Row.set_cell_text,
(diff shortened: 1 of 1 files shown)


report_xls/report_xls.py
```diff
@@ -29,13 +29,29 @@
 from openerp.report.report_sxw import report_sxw
 from openerp import pooler
 import logging
+
 _logger = logging.getLogger(__name__)
 
+xls_types_default = {
```
pedrobaeza, line 35: This is not needed to be moved

pedrobaeza:
I prefer that you mock the types this way instead: https://github.com/OCA/account-financial-tools/blob/8.0/account_move_batch_validate/account.py#L38

report_xls/report_xls.py
yajo, line 35: Its purpose is to have a default in xls_types. You need to declare it before for that.

yajo:
Types are a dict, while in that module `job` is a decorator. These are different cases I think.

pedrobaeza:
OK, as this is only for avoiding problems if the module is not used, then :+1:

lasley:
👍

lasley:
Wow this merge button is new and intriguing! Ok so per the rules it's a 5 day wait if two thumbs, but is that 5 days from the initial thumb or 5 days from the second?

lasley:
~~Also, is there any insight into what's up with Runbot in this repo? Seems all the PR builds are broken~~ 😦  #66

PR #66, opened 2016-08-12
[9.0][FIX] report_xls: Move description to first line

lasley:
I'm not sure why the [RunBot builds are failing](https://runbot.odoo-community.org/runbot/build/3152292) on a non-installable module, but they are and it's killing my OCD. 

This is minimalistic change that ~~hopefully~~ fixes the builds.

### DIFF
--- a/report_xls/README.rst
+++ b/report_xls/README.rst
@@ -1,0 +1,78 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================
+Excel report engine
+===================
+
+This module adds Excel export capabilities to the standard odoo reporting
+engine.
+
+Usage
+=====
+
+In order to create an Excel report you can:
+
+- define a report of type 'xls'
+- pass ``{'xls_export': 1}`` via the context to the report create method
+
+The ``report_xls`` class contains a number of attributes and methods to
+facilitate the creation XLS reports in OpenERP.
+
+* cell types
+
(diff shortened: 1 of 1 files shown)


report_xls/__openerp__.py
```diff
@@ -26,8 +26,7 @@
     'author': "Noviat,Odoo Community Association (OCA)",
     'website': 'http://www.noviat.com',
     'category': 'Reporting',
-    'description': """
-Excel report engine
+    'description': """Excel report engine
```
elicoidal: Simply remove the description key. This is not mandatory anymore with the README file.
(even more this becomes redundant and difficult to keep in sync)

lasley: In this instance there isn't a ReadMe, and I was trying to keep the change-set minimal as this isn't a full conversion & I am just trying to fix the build (all 9.0 PRs are failing right now). If this is a block however, I will move this over to a ReadMe using OCA template.

pedrobaeza: You can cherry-commit https://github.com/OCA/reporting-engine/commit/fddca415b8d255b9d877ea92217ea086205ce077 from 8.0

lasley: Excellent call, thanks @pedrobaeza

lasley:
Oops forgot to drop a note saying comments handled after handling comments. Comments handled! 😄

report_xls/__openerp__.py
atchuthan: you could remove the description field as README.rst was added

pedrobaeza (changes requested):
Although the module is not yet merged, but we can fast-track it anyway.

pedrobaeza (approved):

PR #68, opened 2016-08-17
[FIX] report_xlsx: Protect import + CamelCase class

pedrobaeza:
This includes improvements in the README to avoid to generate modules depending on this one that generates problems:
- Protect the import to avoid the breakage when you don't have _report_xlsx_ module accessible.
- Put the name of the class in CamelCase as PEP8 says.

### DIFF
--- a/report_xlsx/README.rst
+++ b/report_xlsx/README.rst
@@ -22,9 +22,14 @@ An example of XLSX report for partners:
 
 A python class ::
 
-    from openerp.addons.report_xlsx.report.report_xlsx import ReportXlsx
+    try:
+        from openerp.addons.report_xlsx.report.report_xlsx import ReportXlsx
+    except ImportError:
+        class ReportXlsx(object):
+            def __init__(self, *args, **kwargs):
+                pass
 
-    class partner_xlsx(ReportXlsx):
+    class PartnerXlsx(ReportXlsx):
 
         def generate_xlsx_report(self, workbook, data, partners):
             for obj in partners:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @Tecnativa

pedrobaeza:
@antespi, please review

sbidoul:
:+1: for the camelcasing of course

Adding the import error in the readme, I'm -0. At the end of the day, all this hackery is needed because we have suboptimal dependency management. When we have https://github.com/OCA/maintainer-quality-tools/pull/343, it should become largely unnecessary  (we probably need the same in rundbot)

lasley:
👍

yajo:
:+1: :+1: :+1:

pedrobaeza:
@sbidoul, not all the people use or will use that deployment strategy, so I prefer to explicitly put this. I merge then.

yajo:
@sbidoul I think the real solution would be that Odoo did not load uninstalled modules, and thus you could not install them without before parsing `__openerp__.py` and finding out `external_dependencies`, but that's not the way it works right now, so having an `ImportError` just for adding a repo to your addons path is annoying (under some deployment strategies, of course).

sbidoul:
@Yajo That would help yes, but IMO it's not the root cause. The real issue is that people just _copy_ code arround (or git clone it or whatever), where they should use a proper mechanism to _install_ it on the environments including dependencies. That's why I'm working so much on setuptools-odoo and pip... with very good results.

Yajo:
As I told you before, the problem is IMHO that Odoo executes code of uninstalled addons. It should only execute it when it gets installed, after checking for its dependencies.

While your point would indeed fix many of these problems, actually what I'm doing here is fix a bug in correctly installed addons, so this patch is needed no matter your deployment method.

PR #72, opened 2016-09-19
[ADD] report_barcode_elaphe: Module to generate barcode with other lib

jefmoura:
The current barcode generating method in Odoo uses the Reportlab and if I want to generate e.g a Pharmacode I cannot; because, the lib doesn't provide this type nor configurations for it. To provide more options and settings for barcodes, we create this module that uses the elaphe library and overrides the controller that makes barcode. It uses other library to generate barcodes but it will keep the current behavior and add options/settings at the same time.


jefmoura:
I have already signed and sent back the Contributor License Agreement.
As you can see the new module needs a external dependency.
I'm waiting for the next steps.

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+        'pharmacode', 'plessey', 'postnet', 'qrcode', 'royalmail', 'rss14',
+        'symbol', 'upc'
+        :param barmargin: Accepted positive integer values
+        :param width: Accepted positive integer values
+        :param height: Accepted positive integer values
+        :param scale: Accepted positive float values
+        :param extraopts: Accepted different params per type of barcode. Eg: extraopts='{eclevel:M,version:8}'
+        """
+        try:
+            width, height, scale, margin = int(width), int(height), float(
```
lasley: These should all be declared on independent lines.

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+        'i2of5', 'japanpost', 'kix', 'maxicode', 'msi', 'onecode', 'pdf417',
+        'pharmacode', 'plessey', 'postnet', 'qrcode', 'royalmail', 'rss14',
+        'symbol', 'upc'
+        :param barmargin: Accepted positive integer values
+        :param width: Accepted positive integer values
+        :param height: Accepted positive integer values
+        :param scale: Accepted positive float values
+        :param extraopts: Accepted different params per type of barcode. Eg: extraopts='{eclevel:M,version:8}'
+        """
+        try:
```
lasley: This guard is too long & not very Pythonic IMO. Please try to isolate what you are guarding for, and leave everything else outside of it. Suggestions are to move the variable declarations and nested try/except out of this try/except

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+            width, height, scale, margin = int(width), int(height), float(
+                kw.get('scale', 2.0)), int(kw.get('barmargin', 0))
+            opts = {}
+
+            try:
+                extra_opts = {}
+                if kw.get('extraopts', False):
+                    for opt in kw['extraopts'].split(','):
+                        key = opt.split(':')[0]
```
lasley: `opt.split` is being called a lot of times on the same object. I recommend you assign it to a variable and use that instead

report_barcode_elaphe/README.md
```diff
@@ -0,0 +1,61 @@
+![Licence](https://img.shields.io/badge/licence-AGPL--3-blue.svg)
```
lasley: Please remove this file, only the RST is necessary

report_barcode_elaphe/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
lasley: Please use the shorter header & apply to all - https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py#L3

report_barcode_elaphe/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+
+    'depends': [
+        'report',
+        'web'
+    ],
+    "external_dependencies": {
+        'python': ['elaphe']
+    },
+
+}
```
lasley: Please add a newline to end of this file - https://robots.thoughtbot.com/no-newline-at-end-of-file

report_barcode_elaphe/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+##############################################################################
+{
+    'name': "Report Barcode Elaphe",
+    'summary': """Barcode generation controller using elaphe library.""",
+    'description': """Overwrite the report barcode controller to use a more powerful python library of barcode generation.""",
+    'author': "bloopark systems GmbH & Co. KG",
+    'website': "http://www.bloopark.de",
+    'license': 'AGPL-3',
+    'category': 'Base',
+    'version': '1.0',
```
lasley: 8.0.1.0.0

report_barcode_elaphe/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+##############################################################################
+{
+    'name': "Report Barcode Elaphe",
+    'summary': """Barcode generation controller using elaphe library.""",
+    'description': """Overwrite the report barcode controller to use a more powerful python library of barcode generation.""",
+    'author': "bloopark systems GmbH & Co. KG",
+    'website': "http://www.bloopark.de",
+    'license': 'AGPL-3',
+    'category': 'Base',
```
lasley, line 11: I recommend a better category. Likely `Warehouse Management` but I'm not good with the naming/grouping unfortunately. Maybe someone else will have a better recommendation.

report_barcode_elaphe/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "Report Barcode Elaphe",
+    'summary': """Barcode generation controller using elaphe library.""",
+    'description': """Overwrite the report barcode controller to use a more powerful python library of barcode generation.""",
```
lasley: Description can be removed, this is auto-generated from Readme

report_barcode_elaphe/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "Report Barcode Elaphe",
+    'summary': """Barcode generation controller using elaphe library.""",
```
lasley: No need for `"""` unless multiline

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+from openerp.addons.web.http import route, request
+
+import cStringIO
+import re
+
+from werkzeug import exceptions
+
+try:
+    from elaphe import barcode
+except Exception:
```
lasley: Always try to be as specific in your guards as possible. `except ImportError:`

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+            opts = {}
+
+            try:
+                extra_opts = {}
+                if kw.get('extraopts', False):
+                    for opt in kw['extraopts'].split(','):
+                        key = opt.split(':')[0]
+                        if opt.split(':')[1] == 'True':
+                            values = True
+                        elif opt.split(':')[1] == 'False':
```
lasley: IMO you should use [`openerp.tools.safe_eval`](https://github.com/odoo/odoo/blob/8.0/openerp/tools/safe_eval.py#L231) instead of this set of conditionals. Will be more flexible & more efficient

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+    pass
+
+
+class ReportBarcodeController(ReportController):
+
+    """The class of Report controller"""
+
+    @route()
+    def report_barcode(self, type, value, width=0, height=0,
+                       humanreadable=0, **kw):
```
lasley, line 16: Completely re-implementing a method is not desirable in a lot of instances, particularly when unknown modules could depend on, or add to via inheritance, implementation on the parent. You can't guarantee module load order, so inconsistent results are common if not calling the superclass.

I would prefer if there were an argument to explicitly use this implementation & send to super if not defined. This will guard against unpredictable behavior while still allowing for the fanciness that this module provides 😄

report_barcode_elaphe/controllers/main.py
```diff
@@ -0,0 +1,96 @@
...
+            if width and height:
+                barcode_img = barcode_img.resize((width, height))
+            barcode_img.save(barcode_out, "png", resolution=100.0)
+
+        except (ValueError, AttributeError):
+            raise exceptions.HTTPException(
+                description='Cannot convert into barcode.')
+
+        return request.make_response(barcode_out.getvalue(),
+                                     headers=[('Content-Type', 'image/png')])
```
lasley: Please add a newline to end of file - https://robots.thoughtbot.com/no-newline-at-end-of-file

report_barcode_elaphe/readme.rst
```diff
@@ -0,0 +1,65 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+=====================
+Report Barcode Elaphe
+=====================
```
lasley, line 3: Please use  OCA ReadMe template (Usage section, newer bug tracker note, etc) - https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

lasley (changes requested):
Hi @jefmoura - thanks for the submission! I put some comments inline & will create an issue to resolve the Runbot dependency

lasley:
@jefmoura - Travis dependency you can resolve. Just add a `pip install elaphe` in the Travis file install section, [after the `travis_install_nightly`](https://github.com/OCA/reporting-engine/blob/10.0/.travis.yml#L32)

We will also need some tests in order to make sure everything works. Please let me know if you have questions on how to test. What to test should become apparent once Travis passes and Codecov drops a comment in this PR.

Also note I did the review here, but I think this module may be better in the `stock-logistics-barcode` repo. @pedrobaeza , @OCA/logistics-maintainers - thoughts?

pedrobaeza:
About Travis, better to add a regular requirements.txt file like Python packages.

About the repo, this is not something for dealing with barcodes in the backend, but for "printing" them, so I think this is the right place.

lasley:
Thanks for the confirmation @pedrobaeza 

@jefmoura - Let's perform the QA process here, then we can migrate to the correct repo for merge. Thanks again for the contribution! Drop a message here when you're ready for another review or need help.

pedrobaeza:
@lasley, I think you haven't read correctly my comment: this is the correct repository.

lasley:
Oh whoops - I should stop reading my Github messages until after the morning coffee 😆

report_barcode_elaphe/__openerp__.py
jefmoura: Maybe @pedrobaeza can help us here. I don't know which category is better. Can you help us, please?

pedrobaeza: Use the same as the `report` module in Odoo has.

jefmoura: Ok. Thanks, @pedrobaeza

jefmoura:
@lasley I took long to reply but I'm almost done with the changes. I have one question: How to test it? I have created unit test as you can see here https://github.com/OCA/reporting-engine/pull/72/commits/07218a5b29c8b09595522d6c188a62a6aeac971c
Is it enough? Thanks!

report_barcode_elaphe/models/report.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 bloopark systems (<http://bloopark.de>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models
+from openerp.tools.safe_eval import safe_eval as eval
```
lasley, line 6: I would prefer this left as the name `safe_eval`. Overriding namespaces you don't own isn't good IMO & when I saw it later in the code I had to jump back up to the imports to see if it was being done right. Best to leave it explicit

lasley (commented):
Thanks for the update @jefmoura 

Looking at the code, there's still a few missing spots in coverage. Biggest one is in the controller. I typically just mock out the parent of the controller I'm testing then go from there. You can then just call the methods you need, providing the proper returns from parent when you need it.

There's also some failing lints - https://travis-ci.org/OCA/reporting-engine/jobs/183906326#L461

Realistically I'm also good with pushing the controller tests back to the next migration in order to speed up merge. You're already raising the coverage of the repo quite a bit here. Lints are a blocker though definitely.

PR #76, opened 2016-10-11
[MIG] Migration of report_xlsx to 10.0

adrienpeiffer:


lmignon:
@moylop260 Travis fails with a false negative... The check for the use of new api must not be applied to python classes that don't inherit from base.model.....  https://github.com/odoo/odoo/blob/10.0/odoo/report/report_sxw.py#L329

lmignon (changes requested):
We need to find a way to ignore the travis error...Can you declare PYLINT_EXPECTED_ERRORS=1 in the env matrix in .travis.yml

pedrobaeza:
That's a warning, not an error. The error is:

```
report_xlsx/__manifest__.py:4: [W0109(duplicate-key), ] Duplicate key 'installable' in dictionary
```

pedrobaeza (approved):
Squashing the commits on merge

lmignon (approved):
@pedrobaeza you're right...

alexis-via (approved):

sbidoul:
@adrienpeiffer please update the runbot link branch in the readme.

adrienpeiffer:
Thanks @sbidoul !

PR #78, opened 2016-10-14
[10.0] Port report_py3o

alexis-via:


alexis-via:
It seems we have an issue with the filename of the files generated by Py3o: there are 2 dots "toto..pdf"

lmignon:
@alexis-via @faide Before the merge I would propose a refactoring of the code to ease the extension...
For one of my projects I need to extend/change the way and the place where the report are stored. With the current implementation we have no way to easily inherit from Py3OParser and add our specific methods... Moreover the creation of the ir.attachment in the report_sxw is not overridable. My idea is the following:
Replace Py3oParser with a new model Py3oReport that inherit from Report (so we can reuse part of the logic implemented by odoo to check if our reports must be saved etc...)
In this new model we just need to implement the method `create` expected by the method render_report of  ir_actions_report_xml.
https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_actions.py#L197
@alexis-via @faide What do you thing?

faide:
@lmignon I like this proposal ^^

alexis-via:
@lmignon I don't know the code well enough to have an opinion... but if you modify the code here, I'll port it to my v10 PR to keep the 2 versions in sync.

lmignon:
@alexis-via @faide here it's https://github.com/faide/reporting-engine/pull/14

alexis-via:
The problem with double dot in filename is now fixed

JonathanNEMRY:
Last refactoring of @lmignon is done here:
https://github.com/akretion/reporting-engine/pull/2

For this current PR take also a special attention to https://github.com/akretion/reporting-engine/pull/1

-jne

faide:
@lmignon is it ok for you to merge this PR as is or do you need more time? I have people who want to base work on yours to port back to older odoo versions...

lmignon:
@faide  The proposed refactoring  akretion#2 works fine in production with a lot of improvements. I test finalize my tests in the next hours of https://github.com/acsone/reporting-engine/commit/4048c0b8c134f6ac55a8779803b4fc21fe4f5339 and keep you informed once everything is ok

lmignon:
@faide @alexis-via akretion#2 is ready to be merged. I've developed a new module that extends  Py3o Report by letting you specify a rule to choose the appropriate template to use according to an expected language....https://github.com/acsone/reporting-engine/commit/ac196ccee1add42f803ed20e7f4716bc76d7c224 This new module use the changes in the last commit of akretion#2 as extension point.

lmignon (approved):
It's time to move forward :smile:

rousseldenis (approved):
:+1: Code review + tested

JonathanNEMRY (approved):
Ready for merge...please don't squash commits...

pedrobaeza:
Seriously 133 commits without any squash? You have a total crusade against squashing, but I think this is not sustainable as the repo will grow too much in size for no real benefit with several diffs that nullify each other.

yajo:
I can see the commit history is huge and mostly irrelevant (many things seem to come from a migration from a mercurial repo, with tagging, updating .hgignore and so on).

Indeed rebasing it would be a nightmare, but also we get lots of unneeded history. There's the fine balance between authorship and usefulness of commits, but in general I'm more against useless commits, so IMHO it would be good if you squash it (at least, per author and following history, we'd still have about 30 commits, but that feels better in the balance).

alexis-via:
I think that 133 commits for a 6-year-long history of work is pretty small. And there are many different authors that worked together on it, so it's not possible to merge all the commits together.

So we should merge this PR WITHOUT SQUASHING anything. I'm not the only one that thinks that this pro-squashing obsession is starting to be unbearable. And there is a significant risk of killing the motivation of contributors like me and others. This thread will be observed by sraps and other Aeroo fans and we are giving them additional arguments NOT to join the OCA... because working in the OCA is starting to be a nightmare and new rules and regulations are invented every month or so.

pedrobaeza:
OK, do what you think is correct.

yajo:
Ok, merged, not worth it the fight. Thanks for your contribution pals 😉

adrienpeiffer:
:clap:

PR #82, opened 2016-11-08
9.0-mig-report_custom_filename

susport:
Change to version 9.0

PR #84, opened 2016-11-29
[9.0] MIG Custom Report Filenames

nicomacr:


pedrobaeza:
There is already a migration in https://github.com/OCA/reporting-engine/pull/52. Is there any change in yours from this one?

pedrobaeza:
Please review #82 instead of #52, as there's no answer from the original author.

PR #87, opened 2016-12-13
[FIX] use JSON instead of simplejson

hbrunn:
in 9, we should use the `JSON` library instead of simplejson because this is Odoo's new standard dependency for json issues


pedrobaeza (approved):

pedrobaeza:
I think you need to change requirements.txt file

hbrunn:
thanks for the pointer, seems I had to wake up first

lmignon (approved):
Thank you @hbrunn

yajo (approved):

PR #90, opened 2016-12-31
[10.0] Add module report_qweb_txt

alexis-via:
This modules adds a text reporting engine based on Qweb, as discussed in the contributors mailing-list in this thread: https://odoo-community.org/groups/contributors-15/contributors-48500?mode=thread&date_begin=&date_end=

See README.rst for more details.


pedrobaeza:
Should we reflect the underlaying technology calling this report_qweb_txt?

report_txt/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Usage
+=====
+
+To add a text report, you should develop an Odoo module that contains:
+
+* a Qweb template,
+* a Qweb report definition with *report_type = qweb-txt*.
+
+The module *purchase_dilicom_csv* available on the `dilicom Github repository <https://github.com/akretion/dilicom>`_ (branch *10.0*) is a good example: it adds a report *Dilicom CSV Order* on purchase orders. This report is a CSV file with one line per order line and 2 columns: EAN13 and order quantity.
```
pedrobaeza: Instead of this, you should generate a demo report with base objects. For example, a res.users report.

report_txt/models/ir_actions_report_xml.py
```diff
@@ -0,0 +1,19 @@
...
+class IrActionsReportXml(models.Model):
+    _inherit = 'ir.actions.report.xml'
+
+    report_type = fields.Selection(selection_add=[('qweb-txt', 'Text')])
+
+    @api.model
+    def render_report(self, res_ids, name, data):
+        if data.get('report_type') == 'qweb-txt':
+            return self.env['report'].get_html(res_ids, name, data=data), 'txt'
+        else:
```
pedrobaeza: No need for this else

report_txt/models/report.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+
+class Report(models.Model):
+    _inherit = "report"
+
+    @api.model
+    def _get_report_from_name(self, report_name):
+        reports = self.env['ir.actions.report.xml'].search([
```
pedrobaeza: Why this overwrite?

report_txt/README.rst
alexis-via: Done in my last commit.

report_txt/models/report.py
alexis-via: We need this overwrite, cf the original code: https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L493

And I don't like how it's done in report_xml: https://github.com/OCA/reporting-engine/blob/8.0/report_xml/models.py#L52
I prefer not to call super() and have a code that will be good for everyone, independantly of the reporting engines installed.

report_txt/models/ir_actions_report_xml.py
alexis-via: Yes, but I like it :)

alan196 (approved):

report_qweb_txt/demo/res_users_report.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="report_res_users_csv">name;login;last_login_date;lang
+<t t-foreach="docs" t-as="o"><t t-esc="o.name"/>;<t t-esc="o.login"/>;<t t-esc="o.login_date"/>;<t t-esc="o.lang"/>
```
gdgellatly: Just a minor point, if a demo report is called csv, should it not use comma separators?

report_qweb_txt/demo/report.xml
```diff
@@ -0,0 +1,13 @@
...
+
+<odoo>
+
+<report id="res_users_csv"
+    model="res.users"
+    name="report_qweb_txt.report_res_users_csv"
+    string="CSV List"
+    report_type="qweb-txt"
+    file="report_qweb_txt.report_res_users_csv"
+    />
```
gdgellatly: Where is the file extension coming from to save as .csv?

report_qweb_txt/models/ir_actions_report_xml.py
```diff
@@ -0,0 +1,19 @@
...
+
+class IrActionsReportXml(models.Model):
+    _inherit = 'ir.actions.report.xml'
+
+    report_type = fields.Selection(selection_add=[('qweb-txt', 'Text')])
+
+    @api.model
+    def render_report(self, res_ids, name, data):
+        if data.get('report_type') == 'qweb-txt':
+            return self.env['report'].get_html(res_ids, name, data=data), 'txt'
```
gdgellatly: Suggested TODO: Add a filename extension field, visible when report_type is qweb-txt or some similar mechanism to support different file extensions. Alternatively I suppose something like

(qweb-txt, 'Text') , (qweb-txt-csv, 'CSV') then

```python
def render_report(self, res_ids, name, data):
        if data.get('report_type').startswith('qweb-txt'):
            return self.env['report'].get_html(res_ids, name, data=data), data['report_type'].rsplit('-')[0]
```

Would mean adding for example ZPL extension is as simple as

```python
report_type = fields.Selection(selection_add=[('qweb-txt-zpl', 'ZPL')])
```

Couldn't find any way on runbot to change extension.  Seemed to ignore even the report file name field

report_qweb_txt/models/report.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+
+class Report(models.Model):
+    _inherit = "report"
+
+    @api.model
+    def _get_report_from_name(self, report_name):
```
hbrunn, line 12: call `super` here and do your search only if it's result is falsy. And given `super` doesn't raise if nothing is found, this function shouldn't either

hbrunn (changes requested):

report_qweb_txt/models/ir_actions_report_xml.py
hbrunn: I don't think any of this is necessary given we have the custom file name addon

gdgellatly:
I looked at that addon, but it looks like it is for user interface only?
IMO the majority of use cases for text files is not users downloading text
files, but automated communication between systems. Seems crazy to have to
fake an https connection just to call a csv file csv.

report_qweb_txt/demo/res_users_report.xml
gurneyalex: @gdgellatly 'csv' in France defaults to using semi colons (because comma is used as decimal separator). But I agree. 
@alexis-via I suggest using config.csv_internal_sep for this (this is a parameter which can be set in the Odoo configuration file).

alexis-via: @gurneyalex using config.csv_internal_sep would be nice, but I don't know how to access such parameters from a qweb template. I tried a few things without success...

report_qweb_txt/models/report.py
alexis-via: I don't see the advantage of calling super... it seems like a waste a resources for me.
I decided to raise an error if nothing is found in order to have a clean error msg instead of having a crash later on in the code (because I guess it's what happens if nothing is found).
I think the only clean solution is to propose a PR to fix the original code...

alexis-via:
@gdgellatly Yes, most of the time it is used for automated communication between systems, but often these systems don't have webservices or API, and you have to manually get the text file and drop it somewhere... I regularly face this requirement, that's why I decided to develop this small module.

report_qweb_txt/models/report.py
hbrunn: yes, that's the good one. and people/code have adapted to it. we don't know what brilliant overrides for this exist, which is why I think every module should stick to the rules of modularity

hbrunn:
...and a web service usually doesn't care about the mime type or file name, so I don't see where the crazyness comes in here. @gdgellatly please elaborate

gdgellatly:
@alexis-via yes, exactly the same as me, not consumed by web service, we often need to drop a file somewhere or email it with a specific custom format or name.

@hbrunn idk I guess my industry is different, we get lots of specific file format requests with custom extensions for communication that must be followed developed by industry bodies.  e.g. you must ftp file to this server with this format and extension. Obviously its a big deal for you guys to support this so I'll just fork.  Was supported out of box by aeroo which we were replacing.

alexis-via:
One important thing that we need to add and which was supported by aeroo: specify the output encoding. The qweb template should always be in UTF8, but we may want to export the TXT/CSV file in latin1 or any other encoding.

alexis-via:
I'm reviving this PR because I plan to have an OCA module based on report_qweb_txt, so this PR would need to be merged beforehand.
In fact, I already use this module quite often in non-OCA modules, and it works great.

report_qweb_txt/models/report.py
alexis-via: done.

report_qweb_txt/models/ir_actions_report_xml.py
alexis-via: On v10, there is the field print_report_name on ir.actions.report.xml (which has "replaced" the OCA module "report_custom_filename"), but it seems to only work on PDF files, cf this code https://github.com/odoo/odoo/blob/10.0/addons/report/controllers/main.py#L107
So the solution proposed by @gdgellatly seems the best to me.

gdgellatly: @alexis-via actually in v11 I ended up solving my problems differently by integrating directly into the controller.  It solves the encoding and filename issues but has not really been tested in production yet.

https://github.com/odoonz/odoonz-addons/tree/11.0/qweb_txt_report

Take a look and take what you want.

hbrunn (approved):

alexis-via:
I have now implemented all the suggestions from the reviewers @hbrunn and @gdgellatly

gdgellatly:
> I'm reviving this PR because I plan to have an OCA module based on report_qweb_txt, so this PR would need to be merged beforehand.
> In fact, I already use this module quite often in non-OCA modules, and it works great.

@alexis-via I had a working version of this for v11.  A v12 surprise is qweb-text in core. Not sure if it works or not.

patel1008b:
> @alexis-via I had a working version of this for v11

I need it for v11. Can you please share this module if you can.

PR #92, opened 2017-01-05
[9.0][FIX][report_custom_filename] API migration

pedrocs-exo:
Incorrect API change migration on class Reports

### DIFF
--- a/report_custom_filename/controllers/reports.py
+++ b/report_custom_filename/controllers/reports.py
@@ -17,10 +17,10 @@ class Reports(main.Reports):
         context = dict(http.request.context)
         context.update(action["context"])
         report_xml = http.request.session.model('ir.actions.report.xml')
-        reports = report_xml.search(
+        report_ids = report_xml.search(
             [('report_name', '=', action['report_name'])],
             0, False, False, context)
-        for report in reports:
+        for report in report_xml.browse(report_ids):
             if not report.download_filename:
                 continue
             objects = http.request.session.model(context['active_model'])\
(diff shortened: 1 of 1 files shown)


report_custom_filename/controllers/reports.py
```diff
@@ -17,10 +17,10 @@ def index(self, action, token):
         context = dict(http.request.context)
         context.update(action["context"])
         report_xml = http.request.session.model('ir.actions.report.xml')
```
pedrobaeza: Better to use `htt.request.env['ir.actions.report.xml']` for using new API and don't change the rest.

pedrocs-exo: Thank you Pedro! Didn't know that. Changed it and with another minor change and it works fine. Another PR on the way...

lasley: @pcs-sossia No need for another PR, you can just make the change and push to this branch in order to update this PR. This is definitely within scope of "API migration" IMO

report_custom_filename/controllers/reports.py
```diff
@@ -16,10 +16,9 @@ def index(self, action, token):
         action = json.loads(action)
         context = dict(http.request.context)
         context.update(action["context"])
-        report_xml = http.request.session.model('ir.actions.report.xml')
+        report_xml = http.request.env['ir.actions.report.xml']
         reports = report_xml.search(
-            [('report_name', '=', action['report_name'])],
-            0, False, False, context)
+            [('report_name', '=', action['report_name'])])
```
pedrobaeza: As you're touching this, you can include here the condition `('download_filename', '!=', False)` to avoid the later `if`

pedrobaeza (approved):

jjscarafia (approved):

jjscarafia:
:+1: Tested locally and can confirm that it fix the issue

PR #95, opened 2017-01-20
[MIG] Migrate report_xls to v9.0

MiquelRForgeFlow:
Excel report engine
===================

This module adds Excel export capabilities to the standard odoo reporting
engine.

### DIFF
--- a/report_xls/README.rst
+++ b/report_xls/README.rst
@@ -57,10 +57,16 @@ help us smashing it by providing a detailed and welcomed feedback.
 Credits
 =======
 
+Images
+------
+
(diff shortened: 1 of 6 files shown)


JordiBForgeFlow:
Why migrate? https://github.com/OCA/reporting-engine/tree/9.0/report_xlsx already exists

MiquelRForgeFlow:
But this module you mention is another different module. XLS != XLSX

JordiBForgeFlow:
In v9 we should only use XLSX. 

XLS is very old

StefanRijnhart:
I see the rationale for this (the mrp bom structure module port) was also migrated to xlsx, so closing this.

PR #101, opened 2017-02-27
[9.0] [FIX] report_custom_filename: Don't change the initial HTTP response …

lmignon:
…if not OK.

If an error occurs into the base class, the response received from the call to super is already an HTTP error response.
By modifying this response we break the UI since this reponse is no more the one expeceted in case or error.
This change into the initial HTTP reŝponse leeds to different kind of errors into the UI depending of
the browser used. (Invalid CSRF error, UI locked in waiting response or 'Blocked a frame with origin http://localhost:8069 from accessing a cross-origin frame' or ...)

### DIFF
--- a/report_custom_filename/controllers/report_controller.py
+++ b/report_custom_filename/controllers/report_controller.py
@@ -41,6 +41,9 @@ class ReportController(ReportController):
     @http.route(['/report/download'])
     def report_download(self, data, token):
         response = super(ReportController, self).report_download(data, token)
+        if response.status_code != 200:
+            # In case of error don't change the response.
+            return response
         # if we got another content disposition before, ditch the one added
         # by super()
         last_index = None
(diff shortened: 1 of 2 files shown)


lmignon:
@hbrunn Can you review this one? It's really an annoying bug...

hbrunn (commented):
it might be a nice way to future proof this by checking if 200 <= status_code < 300 (cf https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), but I don't see a practical issue, so this is fine I think.
What this fixes is mostly to deliver a proper error message when report generation crashes, right?

hbrunn (approved):
forgot to check 'approve'

lmignon:
@hbrunn 
> What this fixes is mostly to deliver a proper error message when report generation crashes, right?

Yes it's but not only. With some browsers, since the response is not coherent it's blocked by the browser and your app remains blocked with the grayed screen and the waiting message.

StefanRijnhart (approved):

PR #102, opened 2017-02-28
[8.0] [ADD] bi_sql_editor (materialized view and reporting)

legalsylvain:
This module add the possibility to create graphical reporting based on sql requests. (MATERIALIZED VIEW can be used to improve performance)

A wizard allow user to tip an sql request

01_sql_request

Once some settings done, a new report is available
05_reporting_pivot

Note : This PR add a dependency to server-tools, because the module is based on the new module 'sql_request_abstract'.


bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -0,0 +1,194 @@
...
+        'double precision': 'float',
+        'numeric': 'float',
+        'text': 'char',
+        'character varying': 'char',
+        'date': 'datetime',
+        'timestamp without time zone': 'datetime',
+    }
+
+    # Mapping to try to guess most common many2one model, based on field name
+    _MODEL_MAPPING = {
```
mourad-ehm: Hi @legalsylvain , Thanks for this module. 
 "_MODEL_MAPPING" can be replaced by a method that return a dict. 
like this: 
 relation_fields =  self.env['ir.model.fields'].search([('relation', '!=', False)])
dic = {}
  for f in  relation_fields:
      dict.update({f.name: f.relation'})
return dict

pedrobaeza:
Isn't this similar as bi_view_editor?

legalsylvain:
Hi  @pedrobaeza ! 

Yes that's why the name is similar. The concept is similar, but this is another technical approach.

The main differences with bi_view_editor are :
* `bi_sql_editor` provides reporting, based on sql requests,  `bi_view_editor` provides an UI to select fields. A lot of operation are so not possible in `bi_view_editor` (WHERE clauses, UNION, EXTRACT, CAST, etc...)
* `bi_sql_editor`  provides possibility to use materialized views with refreshment managed by cron

I checked code of both modules, and there is very few code to share( ~ some lines about menu and action creation). Another more similar module was [sql_export](https://github.com/OCA/server-tools/tree/8.0/sql_export) (server-tools). SQL -> CSV.
I so refactored this module to share all of the common features in the module [sql_request_abstract](https://github.com/OCA/server-tools/tree/8.0/sql_request_abstract) now merged. (sql checks, avoid sql injonction, groups for security, ...)

I explain more the module in the [readme file](https://github.com/grap/reporting-engine/tree/8.0_ADD_bi_sql_editor/bi_sql_editor) 

kind regards.

legalsylvain:
@mourad-ehm. Thanks for your review ! 
Good Idea indeed ! 
I refactored code to detect relation of many2one field.

Thanks !

mourad-ehm:
It sounds good.
:+1:

bi_sql_editor/README.rst
```diff
@@ -0,0 +1,184 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===========================================================
+BI Views builder, based on Materialized or Normal SQL Views
```
bealdav, line 6: Not sure of my position but I think it's better to have almost the same name between folder name and this title name (just \_ replaced by space and camel case).
Isn't it ?
May be you could add a summary with this title.

bi_sql_editor/README.rst
```diff
@@ -0,0 +1,184 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/143/8.0
+
+Known issues / Roadmap
+======================
+
+* Add 'interval', after type (row/col/measure) field for date(time) fields.
+
+* Dinamically change displayed action name to mention the last refresh of the
```
bealdav: s/Dinamically/Dynamically

bealdav:
just tested on runbot, except 2 minor things :+1: really thanks

legalsylvain:
Hi @bealdav , thanks for your review. I fixed the typo. About the name, the problem is that if I write only, BI SQL Editor, it will be confusing with the very near other module, bi_view editor. So, I tried to resume in on sentence, the objective of this module, to be very clear.

bealdav:
Ok

florian-dacosta:
It seems pretty goot to me
I have tested it and I did not see any problems, it worked like a charm!

:+1:

legalsylvain:
Thanks for your reviews.

PR #103, opened 2017-03-03
set installable flag on report_xlsx

mohamedhagag:
set installable flag on report_xlsx

### DIFF
--- a/report_xlsx/__openerp__.py
+++ b/report_xlsx/__openerp__.py
@@ -16,4 +16,5 @@
     'depends': [
         'base',
     ],
+    'installable': True,
 }


StefanRijnhart:
Why do you need this? It is the default value.

mohamedhagag:
to differentiate between ported and unported modules
In production environments which depend on git repos we have to ensure that production ready modules are intentionally marked as installable and have the the same version of odoo instance running to enable it or add it to the addons path.
Example rule is if module version is 10.0.xxx and installable == true then add it to the addons directory of odoo 10 instance.

lasley:
@mohamedhagag - IMO this logic will cause more trouble than it is worth. I recommend that you set your scripts to handle `True` as the default to the installable flag. It's as simple as `manifest_vals.get('installable', True)`

We had the exact same issue with the OCA script that automatically updates the installable flag to False upon making the new branches. There ended up being a lot of effort being spent on updating the manifest key of hundreds of modules, both in terms of engineering and reviewing power, that the script was eventually changed to handle this condition.

mohamedhagag:
Ok, I will follow this advice, you can close this PR now.

PR #106, opened 2017-03-06
[10.0][MIG] bi_view_editor: Port BI View Editor to V10

astirpe:
Based on the porting made for V9 #59


pankk (approved):
Tested on runbot by creating a BI view. Ran into a javascript error (string of undefined), but clicked it away too fast and wasn't able to reproduce this error. For reference: I got this error when going to the pivot view from the pie view.

pedrobaeza:
Can you maybe squash a bit the commits?

pedrobaeza:
In a quick test, I see that tables are always joined with `INNER JOIN`. Can this be configurable?

astirpe:
@pedrobaeza yes, is possible. The initial idea behind this module is to keep the usability very simple, so adding a checkbox or a select to choose the type of join shouldn't be a problem. The potentials of this  module are a lot, so we can add many other features, but then the actual user interface starts to show its limits. It would be nice to implement a more advanced UI, unfortunately I never had enough time for it.
Anyway I would prefer to use this PR only for the pure porting and, after the merge, develop the improvements in a separate PR.

pedrobaeza:
Please add it to the roadmap then, but I would use `LEFT JOIN` as default instead of `INNER JOIN`, as it's the usual way of thinking when building a view: see all starting table records, and let empty those that doesn't have data for the linked tables.

astirpe:
Rebased and squashed a bit

LoisRForgeFlow (approved):
Functional review LGTM :+1:

LoisRForgeFlow:
@astirpe Maybe I'm lucky... :smile: have you ever migrated using OpenUpgrade a database with this module? I'm facing some issues at the very beginning loading registry cause of this module.

astirpe:
@lreficent are you talking about V8? Anyway I never migrated databases with this module. Uninstalling and installing again doesn't help?

JordiBForgeFlow:
my advice is to uninstall in the original version and re-install in the new version.

LoisRForgeFlow:
@astirpe @jbeficent yes, maybe that's the easiest. Thanks!

yung-wang (approved):
Tested on runbot. I created a simple view consisting of the partner and country tabel. Please merge,

PR #114, opened 2017-03-15
[9.0] [ADD] [report_base_filename] Prints dynamic report name

serpentcs-dev1:
@elicoidal 

This module adds the feature to have dynamic report name as per the attachment name of the report.

### DIFF
--- a/report_base_filename/__init__.py
+++ b/report_base_filename/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Elico Corp (www.elico-corp.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import controllers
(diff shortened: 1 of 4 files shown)


elicoidal (changes requested):
@darshan-serpent Thanks!
Please add README files following OCA template and fix travis+runbot

pedrobaeza:
This module collides with https://github.com/OCA/reporting-engine/tree/9.0/report_custom_filename, and that one is already migrated.

elicoidal:
Actually after checking the  https://github.com/OCA/reporting-engine/tree/9.0/report_custom_filename, I think this module actually makes sense to consider in the repo.
* report_custom_filename proposes a highly flexible way to customize the filenames and it is welcome. Nevertheless it requires customization and technical skills.
* report_base_filename proposes a simple and systematic way to name the filenames for all files

I think both have their space in this repo
@pedrobaeza @dreispt

PR #116, opened 2017-03-23
10.0 del report custom filename

mourad-ehm:
This feature is in 10. 
 Field: Printed Report Name 
ex for sale order: ( object.state in ('draft','sent') and 'Quotation' or 'Sale') + object.name +'.pdf'


dreispt:
I wish the removal commit message were more explicit: in the feature people that want to know why it was removed will try to refer to it.

mourad-ehm:
@dreispt , what do you think about this commit message:  [DEL] module report_custom_filename: feature is already in odoo 10.0

dreispt:
Sounds good.

StefanRijnhart (approved):
Thanks! Is it necessary to take the updates to this module from the previous branch before deleting the module?

pedrobaeza (approved):
Squashing on merge solves the problem!

PR #117, opened 2017-03-27
[IMP] Modification of Report-xls module 

Zhouhao12345:
@seb-elico
@xuan139
This modification is going to update the report-xls module based on the repositories from author 'git@github.com:ju-san/report_xls.git', please have a check and review, thx


elicoidal:
@Zhouhao12345 I fixed the CLA for you (This is Enterprise CLA for Elico).
Nevertheless, before we merge this, we need to confirm with the original author whether they agree to have the code in OCA and they need to provide a Individual or Enterprise CLA.
Do you have a way to contact them?

PR #118, opened 2017-03-27
[MIG] Migration of 'report_xls' Module

Zhouhao12345:
This modification was going to update the report_xls module, and port it to version.9 of odoo, which was based on the responsity 'git@github.com:ju-san/report_xls.git' @elicoidal @xuan139

### DIFF
--- a/report_xls/__openerp__.py
+++ b/report_xls/__openerp__.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 #    OpenERP, Open Source Management Solution
--- a/report_xls/__openerp__.py
+++ b/report_xls/__openerp__.py
@@ -76,7 +76,6 @@ Excel reports in odoo.
     """,
     'depends': ['base'],
     'external_dependencies': {'python': ['xlwt']},
-    'active': False,
     'installable': True,
 }
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
--- a/report_xls/__openerp__.py
+++ b/report_xls/__openerp__.py
@@ -1,24 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#
-#    Copyright (c) 2014 Noviat nv/sa (www.noviat.com). All rights reserved.
-#
-#    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


report_xls/__openerp__.py
```diff
@@ -21,7 +21,7 @@
 ##############################################################################
 {
     'name': 'Excel report engine',
-    'version': '8.0.0.6.1',
+    'version': '9.0.0.0.6',
```
elicoidal: 9.0.1.0.0

elicoidal (changes requested):
Small details. Nevertheless you need to change the README file as well.

report_xls/__openerp__.py
```diff
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
```
elicoidal: @Zhouhao12345   Thanks for your contribution to the OCA. 
Let me introduce to you some basic information about contribution in OCA and I will review other points in more comments inside the code..
In general, please follow the principles: 
* Contribution guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
* Please follow the OCA README template:
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
* We recommend to include a screenshot as well to help new comers (nice to have)
* Please use simple headers in all py files following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

elicoidal (changes requested):
some details

report_xls/__openerp__.py
Zhouhao12345: Thx

LoisRForgeFlow:
This was discussed in a PR that ended closed, maybe is related, please have a look: https://github.com/OCA/reporting-engine/pull/95

elicoidal:
@lreficent I see that. 
Still does not hurt to have this PR accepted as it is done. Would it be possible?
One point is that if the module is not gonna be ported, it should be definitively removed from the v9 and v10 repo.

elicoidal:
We will use the report_xlsx in our side.

LoisRForgeFlow:
@elicoidal Yes, maybe we can use this PR to remove it definitely.

elicoidal:
@dreispt @pedrobaeza would you agree with the idea? 
(completely removing from the repo the uninstallable 9.0/report_xls as we can use from now report_xlsx)

PR #120, opened 2017-04-03
Fix py3o freeze when using server fusion

guewen:
The template file has been partially read for the parsing of the
expressions, so use the original template data instead.  When the fusion
server receives an incomplete template, we don't receive any answer
back.

### DIFF
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -247,7 +247,7 @@ class Py3oReport(models.TransientModel):
             if report_xml.py3o_is_local_fusion:
                 template.render(localcontext)
                 out_stream.seek(0)
-                in_stream = out_stream.read()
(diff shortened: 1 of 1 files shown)


simahawk (approved):

report_py3o/models/py3o_report.py
```diff
@@ -256,6 +256,8 @@ def _create_single_report(self, model_instance, data, save_in_attachment):
                 convertor = Py3oConvertor()
                 data_struct = convertor(py_expression)
                 datadict = data_struct.render(localcontext)
+                in_stream.seek(0)
+                in_stream = in_stream.read()
```
gurneyalex: would this not be the same as `in_stream = tmpl_data` ?

gurneyalex (approved):

PR #125, opened 2017-04-12
[10.0] report_py3o: add option py3o_multi_in_one

alexis-via:
I added this option to be able to have severals records in the same report. For example: you need to print several invoices at once. If you select N invoices and ask for the report, Odoo will generate a zip file with N ODT files inside. With this option, Odoo will generate 1 ODT file with the N invoices inside.

This code has been written rapidly for "make it work" for one of my users and I don't have much experience about the internals of report_py3o...


lmignon (approved):
@alexis-via Thank you for this improvement! 
:+1: Code review + functionnal test on runbot

PR #126, opened 2017-04-22
10.0 report qweb pdf watermark

StefanRijnhart:
Actual migration in this commit: https://github.com/StefanRijnhart/reporting-engine/commit/58e717a

Only functional change is that I removed the technical settings group tag (i.e. debug mode) on the expression field, because it is not possible to edit reports anyway without entering debug mode in the first place.

cc @hbrunn


StefanRijnhart:
@hbrunn is it a known issue that the watermark is not visible under the text in the report as in this example?

invoices

hbrunn:
@StefanRijnhart no, in 9, this works. Does something set a background here? Or is https://github.com/OCA/OCB/blob/10.0/addons/report/static/src/css/reset.min.css not used for some reason?

StefanRijnhart:
@hbrunn thanks for the pointer! This is actually reproducible on the 9.0 runbot. I'm guessing it's the background: transparent property from that css file that does the trick?

StefanRijnhart:
@hbrunn What a minefield. Probably related to https://github.com/wkhtmltopdf/wkhtmltopdf/pull/2084. Works with wkhtmltopdf 0.12.4 (instead of 0.12.2.4) and loading order of report.assets_pdf.0.css and report.assets_common.0.css reversed (the pdf one last, as it includes the reset CSS).

hbrunn:
nice catch!

StefanRijnhart:
nice module!

hbrunn (approved):

pedrobaeza (commented):
Please squash together all "OCA Transbot" commits (https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests)

StefanRijnhart:
@pedrobaeza you got it!

pedrobaeza (approved):
Thanks! Let's wait for the Odoo's answer before merging this with the hack

StefanRijnhart:
I replaced the workaround for https://github.com/odoo/odoo/issues/16610 with the fix that I also proposed to Odoo as per discussion on the upstream bug.

lmignon:
@pedrobaeza @StefanRijnhart Can we merge this one?

StefanRijnhart:
@lmignon Why, yes!

Issue #119, opened 2017-03-30
[10.0] report_py3o crash when using the button "Print Invoice" when report as attachement


alexis-via:
Here is the scenario to reproduce the bug on Odoo 10 with an up-to-date reporting-engine branch:
1) install report_py3o and set-it-up to have a py3o invoice with PDF output and keep the default value "(object.state in ('open','paid')) and ('INV'+(object.number or '').replace('/','')+'.pdf')" for the parameter "Save as attachment prefix"
2) Create a new customer invoice. Validate it.
3) On this invoice in "open" state, click on the button "Print Invoice" -> you get the PDF and this PDF is saved as attachement
4) Click on the button "Print Invoice" a second time: you get this crash:
```
2017-03-30 08:44:19,397 1012 INFO test8 odoo.addons.report.models.report: The PDF document INVFA00000399.pdf was loaded from the database
2017-03-30 08:44:19,397 1012 ERROR test8 odoo.service.report: Exception: coercing to Unicode: need string or buffer, tuple found
Traceback (most recent call last):
  File "/home/odoo/erp/odoo/odoo/service/report.py", line 94, in go
    result, format = odoo.report.render_report(cr, uid, ids, object, datas, context)
  File "/home/odoo/erp/odoo/odoo/report/__init__.py", line 19, in render_report
    return env['ir.actions.report.xml'].render_report(ids, name, data)
  File "/home/odoo/erp/reporting-engine/report_py3o/models/ir_actions_report_xml.py", line 92, in render_report
    }).create_report(res_ids, data)
  File "/home/odoo/erp/reporting-engine/report_py3o/models/py3o_report.py", line 360, in create_report
    with open(result_path, 'r+b') as fd:
TypeError: coercing to Unicode: need string or buffer, tuple found
2017-03-30 08:44:19,612 1012 ERROR test8 odoo.addons.web.controllers.main: An exception occured during an http request
Traceback (most recent call last):
  File "/home/odoo/erp/odoo/addons/web/controllers/main.py", line 68, in wrap
    return f(*args, **kwargs)
  File "/home/odoo/erp/odoo/addons/web/controllers/main.py", line 1468, in index
    request.session.db, request.session.uid, request.session.password, report_id])
  File "/home/odoo/erp/odoo/odoo/http.py", line 118, in dispatch_rpc
    result = dispatch(method, params)
  File "/home/odoo/erp/odoo/odoo/service/report.py", line 35, in dispatch
[...]

PR #139, opened 2017-05-30
[10.0]  [FIX] report_py3o:  must return the path to the report

lmignon:
refs #119


alexis-via (approved):
Tested on a real environnement, works fine.

Issue #133, opened 2017-05-11
[10.0] report_py3o: use print_report_name to generate output filename


alexis-via:
On ir.actions.report.xml, there is a field print_report_name on which you can define a filename with a dynamic expression. It works well with Qweb, but it is not used by report_py3o.

A feature to add during the next code sprint in Barcelona ! :)

PR #140, opened 2017-05-30
[IMP] report_py3o: Take into account print_report_name

lmignon:
If a Printed Report Name is set on the action report, use it as
downloaded filename
refs #133


alexis-via (approved):
Tested in a real environnement. Works very well.

PR #144, opened 2017-06-20
[FIX] Fix Watermark from PNG

etobella:
When we are trying to create a Watermark from a PNG, we need to import PdfImagePlugin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,17 @@ cache: pip
 
 addons:
   apt:
+    sources:
+      - pov-wkhtmltopdf
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
+      - wkhtmltopdf  # only add if needed and check the before_install section below
+
+# set up an X server to run wkhtmltopdf.
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
 
 language: python
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,9 @@ cache: pip
 
 addons:
   apt:
-    sources:
-      - pov-wkhtmltopdf
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
-      - wkhtmltopdf  # only add if needed and check the before_install section below
-
-# set up an X server to run wkhtmltopdf.
-before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
 
 language: python
 
@@ -37,6 +29,7 @@ virtualenv:
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - export WKHTMLTOPDF_VERSION=0.12.4
   - travis_install_nightly
 
 script:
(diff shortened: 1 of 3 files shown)


report_qweb_pdf_watermark/__manifest__.py
```diff
@@ -3,8 +3,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Pdf watermark",
-    "version": "10.0.1.0.0",
-    "author": "Therp BV,Odoo Community Association (OCA)",
+    "version": "10.0.2.0.0",
```
pedrobaeza: This is a fix, not a new refactoring, so please change only last number. Check https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

report_qweb_pdf_watermark/models/report.py
```diff
@@ -40,6 +40,7 @@ def get_pdf(self, docids, report_name, html=None, data=None):
         except PdfReadError:
             # let's see if we can convert this with pillow
             try:
+                from PIL import PdfImagePlugin
```
pedrobaeza: Where is this import used?

pedrobaeza (changes requested):

etobella:
If PdfImagePlugin is not initialized, we cannot transform a PNG image to PDF. The problem is related to Pillow creation: https://github.com/python-pillow/Pillow/blob/master/PIL/PdfImagePlugin.py

pedrobaeza:
Please add this explanation as comment before the import line so that others that see the code don't wonder the same.

JordiBForgeFlow:
@alexis-via We need your help to test the generation of reports in PDF. Currently we are getting stuck when the wkhtmltopdf process is run, and a subprocess is open:
https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L445

However it seems that you managed to get past this issue here!:
https://github.com/akretion/edi/blob/10-migrate/account_invoice_factur-x/tests/test_facturx_invoice.py#L16

.travis.yml
```diff
@@ -29,6 +29,7 @@ virtualenv:
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - export WKHTMLTOPDF_VERSION=0.12.4
```
etobella, line 32: It is currently working on travis. The problem was related to wkhtmltopdf version. It must be 0.12.4, so I download it using travis_install_nightly.
Default Wkhtmltopdf version is not working because it does not add headers on the exported reports.

report_qweb_pdf_watermark/models/report.py
```diff
@@ -8,6 +8,12 @@
 from PIL import Image
 from StringIO import StringIO
 from odoo import api, models, tools
+from PIL import PdfImagePlugin # flake8: noqa
+
```
etobella, line 12: Comment added on the code.
Codecov will fail because previous commits were failing.

JordiBForgeFlow (approved):
Tested on runbot. Allows to print from a watermark using a  PNG image

pedrobaeza (approved):
This will mean a problem with my other PR (https://github.com/OCA/reporting-engine/pull/145/files#diff-354f30a63fb0907d4ad57269548329e3R9), but when the first will be merged, then we solve the conflicts

pedrobaeza:
Fast-tracking this for leading with the other PR

PR #159, opened 2017-08-27
[ADD] report_qweb_chrome_renderer

hbrunn:
Chrome reports
==============

This module was written to make it possible to use [Chrome](https://www.google.com/chrome/browser/desktop/index.html) or [Chromium](https://www.chromium.org/getting-involved/download-chromium) in [headless mode](https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md) to render PDF reports. This allows you to use way more modern CSS for rendering the PDF, in the future hopefully enabling some things [CSS3 adds](https://drafts.csswg.org/css-page-3/).

Installation
============

To install this module, you need to:

- install the `websocket` module: ``pip install websocket-client``
- install the version of Chrome you want to use

Usage
=====

To use this module, you need to:

- create a report as usual, but set the flag `is_chrome_pdf` on the ir.actions.report.xml record

Known issues / Roadmap
======================

* it should be possible to emulate wkhtmltopdf's behavior and use Chrome exclusively
* provide a layout that includes some optional divs for headers, footers, page counter etc
* provide some nice internal and external layouts
* support all report features
* support paper formats
* use --remote-debugging-fd instead of port
* see if it's feasible to keep the browser running


hbrunn:
@pedrobaeza I'm of a mind to rename this to `report_qweb_render_chrome` or `report_qweb_chrome`. Any opinion about that? (this is going to let us create awesome reports btw...)

pedrobaeza:
Yeah, I have been watching it and looks promising. What about `report_qweb_chrome_renderer` for being very verbose about the purpose?

hbrunn:
sounds good, will do, thanks

pedrobaeza:
I have tried on runbot and it's not working: after a lot of time, it says `Unexpected end of data`. Something related to runbot itself?

hbrunn:
runbot doesn't have any chrome installed, so that's an expected result for now. But there should be a proper error message. I'll patch that into the code that looks for wkhtmltopdf probably

pedrobaeza:
I'ts a pity that we can't try on runbot... Maybe when we switch to a full dockerized system.

pedrobaeza:
With last commit, can't you use less/sass with inheritance for avoiding such a big CSS?

hbrunn:
no

pedrobaeza:
OK, if no other solution...

hbrunn:
@gurneyalex, what do you think about installing https://hub.docker.com/r/justinribeiro/chrome-headless/~/dockerfile/ on runbot? Then I could add some code to make it possible to connect a remote chrome, and with the seccomp file mentioned there the setuid root call shouldn't be a problem, right?

lasley:
@hbrunn - will it suffice to run a headless chrome instance alongside the test instance itself? We could rig up something like [this](http://www.amihaiemil.com/2017/07/14/selenium-headless-chrome-travis.html)

hbrunn:
@lasley that will work as long as it's localhost, which is why this won't really work. But I was planning to fix the issue by allowing to configure a hostname too, then we can do exactly that

lasley:
Ah yeah this would be more useful with a hostname anyways IMO - would allow for proper resource isolation.

I'll play and get us a headless chrome in a Travis file then. I think I'll just add a helper into MQT like with the current `WKHTMLTOPDF_VERSION`. This will be interesting, because we could actually play with the idea of fully replacing the PoS that is Phantom...

PR #164, opened 2017-09-14
[10.0][report_qweb_pdf_watermark] Fixes on tests and use pyPDF2

JordiBForgeFlow:
Forward port of past commits made by @hbrunn on the 9.0 branch:
https://github.com/OCA/reporting-engine/commit/c0252f7ce6bec407fcb5eadc814469562f866b7f
https://github.com/OCA/reporting-engine/commit/fcd1117331f8bbd9560d24798ef1aa05fa6acc77

### DIFF
--- a/report_qweb_pdf_watermark/README.rst
+++ b/report_qweb_pdf_watermark/README.rst
@@ -8,6 +8,11 @@ Pdf watermark
 
 This module was written to add watermarks (backgrounds) to PDF reports. Because of the way wkhtmltopdf handles headers and footers in the current versions, it is quite impossible to have a background for the complete page using HTML and CSS. That is why this module inserts the image at the PDF level.
 
+Installation
+============
+
+This module works out of the box, but is faster if you install the python library PyPDF2.
+
 Usage
 =====
 
--- a/report_qweb_pdf_watermark/__manifest__.py
+++ b/report_qweb_pdf_watermark/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Pdf watermark",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "author": "Therp BV, "
               "Odoo Community Association (OCA)",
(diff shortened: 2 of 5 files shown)


hbrunn:
any reason not to cherry pick those?

You'll also want to add https://github.com/OCA/reporting-engine/pull/161, because this speeds up the merging by factor 100-1000

JordiBForgeFlow:
OK, will do that then.

JordiBForgeFlow:
@hbrunn I propose to add #161 on a separate PR. This PR is intended to greenify Travis, which now fails in 10.0 due to `report_qweb_pdf_watermark` and `report_qweb_signer`.

@pedrobaeza Travis fails in lint checks on `report_qweb_signer` https://travis-ci.org/OCA/reporting-engine/jobs/275429857#L705, but I think it's an error of lint. The readme.rst repeats some key words.

pedrobaeza:
No, the real error is
```
report_qweb_signer/tests/test_report_qweb_signer.py:9:1: W293 blank line contains whitespace
```
but it's very curious that this didn't raise before... I will check.

pedrobaeza:
You are modifying that file, so it's your error...

JordiBForgeFlow:
Yes, I just fixed the error. The problem is that Travis was stalling due to this module. The root cause is that seems that one can only test PDF reports using HttpCase. Otherwise it stalls. The same happened with `report_qweb_pdf_watermark`.

But not sure why HttpCase works, and others don't. Perhaps @hbrunn knows better.

pedrobaeza:
I think it was @sylvain-garancher who found that problem in OCA/reporting-engine

JordiBForgeFlow:
In any case this PR greenifies Travis for the 10.0 branch. For 9.0 this had already been fixed.

JordiBForgeFlow:
I propose to merge this PR now, as it will greenify the current 10.0 branch, which is failing now and causes all open PR's to fail also.

pedrobaeza (approved):
OK, fast-tracking

hbrunn (approved):
agreed, next approval should fast track imho

PR #172, opened 2017-10-31
10.0 py3o improvements

omar7r:
Hi,

Adds two new functions to report's context, one allows to interprete \n and \t chars in Odoo's strings and the other one cleans texts from html tags. Without this py3o doesn't respect carriage return in text field and print the html tags in the html fields of Odoo.

Regards

### DIFF
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -9,7 +9,6 @@ import json
 import logging
 import os
 from contextlib import closing
-from genshi.core import Markup
 
 import pkg_resources
 import requests
@@ -28,6 +27,7 @@ try:
     from py3o.template.helpers import Py3oConvertor
     from py3o.template import Template
     from py3o import formats
+    from genshi.core import Markup
 except ImportError:
     logger.debug('Cannot import py3o.template')
 try:
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -9,6 +9,7 @@ import json
 import logging
 import os
 from contextlib import closing
+from genshi.core import Markup
 
 import pkg_resources
 import requests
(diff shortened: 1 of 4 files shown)


houzefa-abba:
Hey,

LGTM though quite unfortunate it is done here on Odoo's side rather than in `py3o.template`, which is the lib that knows about ODF internals (using `<text:line-break/>` is very much ODT specific).

But again, LGTM as in I wouldn't mind having this added into `report_py3o` for now, especially as it is completely optional.

In the long run, we have [an issue open about HTML data handling](https://bitbucket.org/faide/py3o.template/issues/15/question-how-to-insert-html-snippet) into `py3o.template` itself that would be pretty awesome to have! =)

report_py3o/models/py3o_report.py
```diff
@@ -64,12 +65,22 @@ def fct1(fct):
     return fct1
 
 
+def format_multiline_value(value):
+    if value:
+        return Markup(unicode(value).replace('<', '&lt;').replace('>', '&gt;').
```
lmignon: @omar7r Why do you need to call 'unicode'??? What if the value is not unicode compliant?

lmignon (commented):
@omar7r Thank you for this contrib. As @houzefa-abba says the best place for this change should be in `py3o.template`.  Nevertheless, I have tried many times to bring this functionality at the level of the genshi parser, but unfortunately without success given the' event-driven' nature of the parser. The proposed solution deserves to be functional.

report_py3o/models/py3o_report.py
omar7r: It's not needed. I just deleted it.

lmignon:
@omar7r Can you fix the 2 lint issues? (Not related to you changes but making travis unhappy). Once done, this PR is ok for me. Thank you for this enhancement!

omar7r:
Done @lmignon Thank you for your review

lmignon (approved):
Thank you @omar7r 

@alexis-via Can you take a look at this one?

alexis-via (approved):
Strange to do this in the odoo module rather than py3o.template, but if laurent says it's too difficult to do in py3o.template... let's go. But it means we may have to re-write py3o.template one day...

houzefa-abba:
Hey,

See <https://bitbucket.org/faide/py3o.template/pull-requests/43/add-multiline-format-function/diff> for a related PR in the `py3o.template` lib.

PR #174, opened 2017-11-17
[FIX] report_qweb_signer: Remove hardcoded memory limits

yajo:
These limits were being hit when printing PDF reports with just 80 pages.

Now it printed 500 pages with no hassle, just with default OpenJDK settings.

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
@santiarg I remember you had this problem.

sergio-teruel (approved):
Thanks @Yajo

carlosdauden (approved):

Issue #179, opened 2017-12-17
[RFC] report_py3o using libreoffice --headless --convert-to


hbrunn:
@alexis-via I think I'm not the only one who is a bit scared by the dependencies of report_py3o when we want document conversions. From studying the code, I don't see a reason why we can't call the libreoffice binary in headless mode to do the conversion after py3o has done its job. Did you try this and it didn't work or does the fusion server have any advantages we want to have that I'm ignorant of? Anyways, I'm of a mind to PR a patch to make this (local libreoffice conversion) an option.

Yakulu:
I guess direct call to LO binary is what @elicoidal has done into [his workaround](https://github.com/Elico-Corp/odoo-addons/blob/10.0/report_py3o_convert_xlsx/controllers/main.py#L49) to get XLSX conversion working with py3o

rvalyi:
@hbrunn I may be wrong but I guess the fusion server would have a speed advantage as Libreoffice is mostly Java based and hence will have a slow startup time (time to allocate lots of RAM and let the JVM optimizations kick in) so server mode would perform much better than a command line call. That being said a degraded command line mode might be welcome.

hbrunn:
thanks for the input everyone, I'll propose a PR then to move the fusion server stuff into its own module, to have minimal dependencies for `report_py3o` in order to spread its usage. Whoever needs the fusion server can install that, others won't be affected

hbrunn:
For the record: The claim about Libreoffice being java is [not entirely true](https://wiki.documentfoundation.org/Development/Java)

hbrunn:
done in #180

Yakulu:
Thanks @hbrunn , great work ! About performance, something like [unoconv](https://github.com/dagwieers/unoconv) might be explored.

hbrunn:
no, that would be reinventing the wheel. If libreoffice startup times are an issue, that's what the fusion server solves for you as @rvalyi pointed out

PR #180, opened 2017-12-17
[RFR] split off the fusion server to its own module; use libreoffice for conversions in the base version

hbrunn:
This is the result of #179, I propose to split `report_py3o` into a base module that does conversions using libreoffice, and `report_py3o_fusion_server` for those who want to use it. Now it's a lot simpler to get started with this, I promise myself more users from this.


elicoidal (approved):
Great job!

report_py3o/models/py3o_report.py
```diff
@@ -244,60 +241,63 @@ def _create_single_report(self, model_instance, data, save_in_attachment):
         """ This function to generate our py3o report
         """
         self.ensure_one()
-        report_xml = self.ir_actions_report_xml_id
-        filetype = report_xml.py3o_filetype
         result_fd, result_path = tempfile.mkstemp(
-            suffix='.' + filetype, prefix='p3o.report.tmp.')
+            suffix='.ods', prefix='p3o.report.tmp.')
```
Yakulu, line 245: Not sure I got it, why is .ods suffix fixed ? Single report is not also about odt, odp... ?

hbrunn: that's an openofficeism: soffice when called with convert-to will want to create a file with the extension of the wanted format, and determines this by finding the first dot from end. so you need to provide something (different from the input file), and an extension known to be native seems safe as dummy value

lmignon (approved):
@hbrunn Thank you for this useful refactoring. It greatly improves modularity! (Code review only)

Before merging we must bump the major version number.

hbrunn:
you can test this in runbot however by selecting some user and printing the demo report.

The addon behaves weirdly with multiple records (emits a zip file with the original extension, this confuses my browser), but I'll rather have this merged first and then follow up with improvements

lmignon:
@hbrunn Indeed the name of the result file in case of multiple records is not right. On the other side, the merge of pdf works well. (Tested on runbot).
@alexis-via Can you look at this one so we can merge it.

sbidoul:
Very nice!

report_py3o_fusion_server/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'name': 'Py3o Report Engine - Fusion server support',
+    'summary': 'Let the fusion server handle format conversion.',
+    'version': '10.0.1.0.0',
+    'category': 'Reporting',
+    'license': 'AGPL-3',
+    'author': 'XCG Consulting,'
+              'ACSONE SA/NV,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/reporting-engine',
+    'depends': ['report_py3o'],
```
lmignon, line 14: @hbrunn The external dependencies are missing
```python
     'external_dependencies': {
         'python': ['py3o.template',
                    'py3o.formats']
},
```

lmignon (changes requested):

report_py3o_fusion_server/__manifest__.py
hbrunn: that's already taken care of by the base module this one depends on: https://github.com/OCA/reporting-engine/pull/180/files#diff-e165a51d925b6f5fdbab80a9482fd3fdR17

lmignon: @hbrunn Since you explicitly import these modules in `report_py3o_fusion_server`you must depend on these. It's never safe to rely on transitively acquired python dependencies.

lmignon (changes requested):

report_py3o_fusion_server/__manifest__.py
hbrunn: @lmignon I disagree, protecting the imports is enough imho, but I won't press the point. Now with external dependencies.

lmignon (approved):

lmignon:
@hbrunn  Before I merge can you increment the version of report_py3o into the manifest PLZ
https://github.com/OCA/reporting-engine/pull/180/files#diff-e165a51d925b6f5fdbab80a9482fd3fdR8 (IMO 10.0.2.0.0 since it's a major change)

hbrunn:
thanks, bumped version and squashed commits

lmignon:
@hbrunn Thank you for the last changes.  I'm a little bit afraid to merge this PR. Indeed, it will break existing installations if the user don't pay attention to the changes introduced by this PR. In my case I can deal with it and your changes greatly improve the complexity of this addon by simplifying its installation. 
@alexis-via I know you use this addon in production. Is-it ok for you? cc @elicoidal @sbidoul
@hbrunn Would you agree to wait a few more days for others to give their opinion on this last point? As I said, I am in favour of merging.

hbrunn:
sure, we're not in a rush. It's indeed true that this will break when a user making use of the fusion functionality updates without thinking, thus won't know she has to install an extra module.

While I think we shouldn't enable not thinking out of principle, we might be able to take the sting a bit by providing a migration script that checks if there's any server configured, if so, mark the fusion server module as to install?

lmignon:
> While I think we shouldn't enable not thinking out of principle, we might be able to take the sting a bit by providing a migration script that checks if there's any server configured, if so, mark the fusion server module as to install?

If it's possible we have the solution. This migration script should fail if the fusion server module is not available.

report_py3o/README.rst
```diff
@@ -12,99 +12,30 @@ The py3o reporting engine is a reporting engine for Odoo based on `Libreoffice <
...
+* For a PDF report in A4/Letter format, it's easier to develop it with a tool such as Libreoffice that is designed to create A4/Letter documents than to develop it in HTML/CSS, also some print peculiarities (backgrounds, margin boxes) are not very well supported by the HTML/CSS based solutions.
 * If you want your users to be able to modify the document after its generation by Odoo, just configure the document with ODT output (or DOC or DOCX) and the user will be able to modify the document with Libreoffice (or Word) after its generation by Odoo.
-
 * Easy development of spreadsheet reports in ODS format (XLS output possible).
 
-This reporting engine is an alternative to `Aeroo <https://github.com/aeroo/aeroo_reports>`_: these 2 reporting engines have similar features but their codes are completely different.
+This reporting engine is an alternative to `Aeroo <https://github.com/aeroo/aeroo_reports>`_: these two reporting engines have similar features but their implementation is entirely different. You cannot use aeroo templates as drop in replacement though, you'll have to change a few details.
```
alexis-via: We could advertise the URL https://github.com/aeroo-community/aeroo_reports rather than https://github.com/aeroo/aeroo_reports

alexis-via: I updated this along with other things of the README in https://github.com/hbrunn/reporting-engine/pull/1/files

alexis-via:
Fine for me. A migration script would be the icing of the cake, but it's OK for me to do it manually (even if I have about 10 installation with py3o fusion in production :)

elicoidal:
> Indeed, it will break existing installations if the user don't pay attention to the changes introduced by this PR

Could we add a warning in the Roadmap/Issue section that the bump to version 2 implies a manual migration of the server?

Other than that, still think this is necessary and that the users currently using this feature will be able to face the migration process (so migration script better but nice to have for me).

hbrunn:
thanks @alexis-via, just merged your improvements. And added https://github.com/OCA/reporting-engine/pull/180/commits/da4ff7bd74fbce58344ab7fe97f7b4a628170e5c, with this, users who have the module already installed get report_py3o_fusion_server installed automatically on update, while new installs are not affected.

hbrunn:
can we merge?

PR #181, opened 2017-12-19
[10.0][FIX]report_py3o: Escape correctly html characters.

jesusVMayor:
The previous behavior fails when the string contained characters such as "&". And it chains many replace.

### DIFF
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -7,6 +7,7 @@ from base64 import b64decode
 from cStringIO import StringIO
 import logging
 import os
+import cgi
 from contextlib import closing
 import subprocess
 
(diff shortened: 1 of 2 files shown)


houzefa-abba:
LGTM

lmignon (changes requested):
@jesusVMayor Thank you for the fix. Can you add a unitest to avoid future regression PLZ?

jesusVMayor:
Test added. 
I don't know how to fix the travis error in the report_qweb_signer module

jesusVMayor:
Finally all green

lmignon (approved):

alexis-via (approved):
I tested this patch and I confirm that it fixes bug #220

PR #188, opened 2018-01-17
[FIX] unlink rule before model

legalsylvain:
If model is created, unlinking rule is not possible because the rule has a required field to the model.

This PR change the order of the deletion

++ calling super, avoiding infinite loop.

CC : @jbeficent, @aheficent, @nicomacr

regards.

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -176,6 +176,8 @@ class BiSQLView(models.Model):
     @api.multi
     def button_set_draft(self):
         for sql_view in self:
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
:+1:

AaronHForgeFlow (approved):
:+1:

legalsylvain:
@aheficent. thanks for updating your review.
Do you have another colleague who are using this module, for this quick review ? 
regards.

AaronHForgeFlow:
@jbeficent

hparfr (approved):

PR #193, opened 2018-02-05
WIP [11.0][NEW] report_pdf_combination

yelizariev:
report that combines multiple pdf into one

PR #194, opened 2018-02-05
WIP [10.0][NEW] report_pdf_combination

yelizariev:
report that combines multiple pdf into one.

See https://github.com/OCA/reporting-engine/pull/193 for more information


dreispt:
Just noticed that this is a similar feature to #334.

yelizariev:
Seems that in #334 multiple Qweb are merged into one, while this module merges PDFs.
But, I'd close this one anyway, cause I don't plan to finish it (seems that only tests are unfinished)

PR #197, opened 2018-02-28
[FIX] Don't break printing some upstream reports with invalid models

StefanRijnhart:
When no background configuration is present on the report definition,
don't try to browse the report model especially because this breaks
some upstream reports (see https://github.com/odoo/odoo/pull/23389)


StefanRijnhart:
Upstream fix was merged, so feel free to close this even if I personally feel that a day not preventing an unnecessary `eval` call is a day not lived.

pedrobaeza (approved):
I see this also OK, but please change module version number

StefanRijnhart:
@pedrobaeza done

feketemihai (approved):

PR #199, opened 2018-03-11
[IMP] report_fillpdf, report_xlsx, report_xml: Use content_disposition helper

naglis:
Reasoning:
- safer ([uses percent encoding](https://github.com/odoo/odoo/blob/350973ff36619bec129a3dace312ec0ae561c55e/odoo/http.py#L1632));
- handles [some browser quirks](https://github.com/odoo/odoo/blob/350973ff36619bec129a3dace312ec0ae561c55e/odoo/http.py#L1635-L1640);
- if some improvements are made to the `content_disposition` function in the future, one would get the benefits for free.

### DIFF
--- a/report_fillpdf/__manifest__.py
+++ b/report_fillpdf/__manifest__.py
@@ -9,7 +9,7 @@
               'Odoo Community Association (OCA)',
     'website': "http://github.com/oca/reporting-engine",
     'category': 'Reporting',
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):
TIL about `content_disposition`

etobella (approved):
LGTM :+1:

PR #202, opened 2018-03-22
[11.0][MIG] bi_view_editor

tarteo:
**Depends on**:
- [x] https://github.com/OCA/reporting-engine/pull/229 to be green

**Added / changed**:
- Visible join node in the widget
- Refactored js and qweb
- Cleaner / more user friendly join dialog
- Context menu is less clumsy
- Javascript implementation for left joins
- Translations (_not properly working_)


angelmoya:
Functional review, I'm not able to create new view:

```
Error:
Uncaught Error: QWeb2: Template 'bi_view_editor.Frame' not found

https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3689
Rastreo de error:
Error: QWeb2: Template 'bi_view_editor.Frame' not found
    at Object.exception (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3689:7)
    at Engine._render (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3734:103)
    at Engine.render (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3730:151)
    at Class.renderElement (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3943:1078)
    at Array.<anonymous> (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:3943:917)
    at Array.<anonymous> (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:802:681)
    at fire (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:796:299)
    at Object.add [as done] (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:797:467)
    at Array.<anonymous> (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:802:649)
    at Function.each (https://xxx.pesol.es/web/content/182628-61266ad/web.assets_common.js:625:767)
```

tarteo:
Hey @angelmoya I rebuilded runbot. But I don't get the error can you try again, please?

astirpe:
@tarteo could you cherry-pick the fix made in https://github.com/OCA/reporting-engine/pull/232 ?

tarteo:
@astirpe This field does not exists in 11.0.

astirpe:
It exists, it's just moved to a separate module `base_sparse_field`, but still in the core.

astirpe:
@tarteo also the migration script made in #232 should be included.
I'm testing this module but it seems that, after the changes with new serialized field, the widget is not working fine, could you check?

MiquelRForgeFlow:
Shouldn't be cherry-picked https://github.com/OCA/reporting-engine/pull/224?

bi_view_editor/models/models.py
```diff
@@ -0,0 +1,171 @@
...
+from odoo.exceptions import UserError
+
+from odoo.tools import (existing_tables, topological_sort)
+
+_logger = logging.getLogger(__name__)
+
+
+@api.model
+def _bi_view(_name):
+    return _name[0:6] == 'x_bve.'
```
MiquelRForgeFlow, line 14: if `_name` has less than 6 characters, it will broke here I think.

tarteo: It will not break. If there are less than 6 characters it will take the full string.

MiquelRForgeFlow: ok

bi_view_editor/models/ir_model.py
```diff
@@ -0,0 +1,291 @@
...
+    @api.model
+    def create(self, vals):
+        if self._context and self._context.get('bve'):
+            vals['state'] = 'base'
+        res = super(IrModel, self).create(vals)
+
+        # this sql update is necessary since a write method here would
+        # be not working (an orm constraint is restricting the modification
+        # of the state field while updating ir.model)
+        q = "UPDATE ir_model SET state = 'manual' WHERE id = %s"
```
MiquelRForgeFlow, line 280: I don't understand. Why putting `vals['state'] = 'base' `in https://github.com/OCA/reporting-engine/pull/202/files#diff-5d42ae09ec3aa29081fdaefd4d438201R274 and then put this query? Will all modules have `state = 'manual'?` Then why in https://github.com/OCA/reporting-engine/pull/202/files#diff-824d3c4f48b7f296e915d38877993ceaR390 there is the `'state' = manual`?

bi_view_editor/models/models.py
```diff
@@ -0,0 +1,171 @@
...
+
+models.BaseModel._auto_init = _auto_init
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    @api.model
+    def _auto_init(self):
+        if not _bi_view(self._name):
```
MiquelRForgeFlow: Why putting this here if already done the same in the monkey patch https://github.com/OCA/reporting-engine/pull/202/files#diff-c154b9950056aa3ae207dc2242d334c9R80?

bi_view_editor/models/ir_model.py
astirpe: @mreficent all such strange pieces of code you are finding out are used as workarounds, in order to bypass the standard Odoo constraints ...

bi_view_editor/models/models.py
astirpe: You're probably right, this part could be not necessary anymore. We should try what happens by removing it ...

bi_view_editor/models/ir_model.py
MiquelRForgeFlow: But I want to full understand them. Because maybe we can think out other workarounds less intrusive.

astirpe: If we keep state =  'manual' here, than a constraint of standard Odoo is triggered while invoking the create. In order to make the create() working fine, the workaround consists in:

- set state =  '< something >' (something different than 'manual' or 'auto')
- invoke the create()
- set state back to 'manual', using SQL to bypass anther ORM contraint

I think the best way to understand this is to play with it, commenting out those parts of code and debugging a bit. I would recommend you to do so, hopefully you will find a better way to do bypass the constraints. Unfortunately I spent so much time on that part of code that I give up to continue to search a better solution.

MiquelRForgeFlow: Ok. I will play with it :)

MiquelRForgeFlow:
I don't understand this

selection_015

tarteo:
@mreficent This is probably because you didn't select a field to be present in the list view. To enable a field to be in the list view you can click the checkbox with the list icon https://fontawesome.com/icons/list?style=solid

MiquelRForgeFlow:
@tarteo I don't know. Am I doing wrong something wrong?

First I create a simple view:
selection_017

And then, when I click on `Open BI View` I get:
selection_018

tarteo:
@mreficent I made fields by default appear in the tree view. Now you should see the fields by default in the tree view, can you test again?

MiquelRForgeFlow:
Yeah, now it works in tree view.

Two more questions:

- I try to click on the options icon, but it seems unclickable. What that icon means?
selection_019

- When appears this wizard, obviously I can click on this four options, but are the 1st and the 3rd just the same? and the 2nd and the 4th? If so, why showing repeated options?
selection_020

tarteo:
@mreficent to change the options you have to right click to open the context menu.
It's probably because you have 2 fields of the Contact model but connected with a join. So e.g. you can choose to show the name of the Contact or Parent Contact.
Can you send me the configuration you was using to see if this is the case?

MiquelRForgeFlow:
@tarteo I understood the problem with the icon. This icon can be opened with right click when the view is in draft state and your are in edit mode. Thus, I propose you to not show the icon if the view is not in draft state or put the icon in different color to differentiate when it's clickable from when it's not clickable.

For the 'join' wizard question, you can check it if for example you create a view and select first the 'name' for Contact (res.partner) and then you select 'company_id' for Contact and the wizard will appear. Thus, my question is what any of this 4 options mean? And if some of them are the same, then I think the duplicated ones should be deleted.

astirpe (approved):

ajaychauhan727:
Thanks, @tarteo your PR helped me.

pedrobaeza (approved):
I have squashed commit history, and merging when CIs finish.

PR #204, opened 2018-03-28
[10.0] add report_xlsx_helper

luc-demeyer:
This 'report_xlsx' add-on module adds the same functional level as the V8.0 report_xls module for creating xlsx reports (the functional coverage is actually a little bit higher than report_xls and the usability is highly improved).
I therefor recommend that we do not merge this functionality as a separate module but that the functionality in this module get merged into the report_xlsx module so that report_xlsx becomes a valid successor of the old report_xls module.

I'll rework this PR if the 'reporting-engine' PSC agrees to move towards a single xlsx  reporting engine module.


pankk:
This sounds like a good idea. If I understand correctly doing this will not break anything, as the changes you propose to add to report_xlsx will be strictly additive. The features in report_xlsx before the intended merge will remain the same. I'm in favor of your idea. I'd love to hear the opinion of others whether or not modularity is preferred in a case such as this though.

sbidoul:
I personally prefer to keep a separate `report_xlsx_helpers`, in order to keep the core `report_xslx`module small and easier to maintain for those who do not need the helpers.

pedrobaeza:
OK, but please remove the plurals in module name.

pankk (approved):
Fixed the boolean is integer pitfall.
int, long, float; yes, I don't think anyone will want to export complex numbers from Odoo any time soon. All good.

pedrobaeza:
But Travis is still red

luc-demeyer:
Travis is red because of bi_view_editor/models/bve_view.py:208: [E8103(sql-injection), 
No relation with this PR hence should not be blocking to merge this one.

@astirpe : any idea on the best way to greenify the bi_view_editor ?

astirpe:
@luc-demeyer you can add:
```
# pylint: disable=sql-injection
```
like we did in https://github.com/OCA/reporting-engine/pull/202

astirpe:
Anyway the warning on the bi_view_editor was not the cause of the travis failure.
It's this one:
```
report_xlsx_helper/report/abstract_report_xlsx.py:17: [R8110(old-api7-method-defined), AbstractReportXlsx.create] Method defined with old api version 7
```

luc-demeyer:
travis logs gives an overview of checks that are executed but do not affect the build status, the 'old-api7-method-defined' is in the list of non-blocking checks, cf. section:
"Next checks are still in beta they won't affect your build status for now:"

The first sql-injection error is gone now.
I added the pylint disable to clear the remaining two sql-injection pylint errors.

astirpe:
So it's the other one :smile:

luc-demeyer:
@pedrobaeza 
Pedro, it looks like I need your help on this one.
The "# pylint: disable=old-api7-method-defined" doesn't seem to work since pylint continues to report the error. I am sure it will take you less than a minute to tell me what is the right syntax to greenify this PR.

pedrobaeza:
I'm afraid you are not interpreting correctly the Travis log. It's true that warnings are not good differentiated from real errors, although the ideal is to not have any of both, but the real error is:
```
report_xlsx_helper/report/abstract_report_xlsx.py:558: [W8112(eval-referenced), AbstractReportXlsx._eval] "eval" referenced detected.
```
A use of an unsecured eval which is a security risk, and thus an error. Replace it with a `safe_eval` call (which is safer, although not infallible).

luc-demeyer:
@pedrobaeza 
Thanks for the help.
I had already a pylint disable on the eval (since this is code coming from the 8.0 report_xls module, but travis now seems to give another warning number. The code already includes documentation why eval is safe in this particular case.

A question on travis log: 
would it be possible to use a consistent color on blocking warnings (e.g. always red) so that non-travis skilled contributors like me can rapidly identify what needs to be corrected ?

tarteo (changes requested):
Can you increase coverage please.

luc-demeyer:
@pedrobaeza 
I have no idea what's wrong this time. It looks like travis job 844.2 simply stopped running.
Can you help me out (once again) ?

pedrobaeza:
Sometimes Travis fails by unknown circumstances. Restarting the build should make the trick. I have already restarted it. Let's see.

luc-demeyer:
Finally all green. 
I think this PR is ready for merge.

tarteo (approved):

tarteo:
@luc-demeyer Thanks! :blush:

report_xlsx_helper/report/abstract_report_xlsx.py
```diff
@@ -0,0 +1,561 @@
...
+
+from odoo import api, fields, _
+from odoo.addons.report_xlsx.report.report_xlsx import ReportXlsx
+from odoo.exceptions import UserError
+
+
+class AbstractReportXlsx(ReportXlsx):
+
+    # pylint: disable=old-api7-method-defined
+    def create(self, cr, uid, ids, data, context=None):
```
pedrobaeza, line 18: I don't get why do you need to put this in "old API"

luc-demeyer:
@pedrobaeza 
the underlying code for this method in Odoo standard addons as well as in the OCA report_xlsx module is still in old api. So I simply didn't touch it. It don't think this should be blocking for merging this PR. 

I am prepared to make a subsequent PR (hence after merging this one) that changes the create method in both report_xlsx and report_xlsx_helper to the new api.

PR #208, opened 2018-04-16
[IMP] check for template data, not for the id

hbrunn:
we should check if a template actually has data, given the binary field is not required: https://github.com/hbrunn/reporting-engine/blob/12be6cbca8895514e0cc63ad012876f9e0ec67e3/report_py3o/models/py3o_template.py#L11


lmignon (approved):
@hbrunn Thank you for the fix!

lmignon:
@alexis-via Can you check this one so we can merge-it? Thanks in advance.

alexis-via (approved):

PR #213, opened 2018-04-27
[FIX] Javascript definition name was duplicated

etobella:
Javascript function name was duplicated and was giving problem when `report_xlsx` and `report_xml` where both installed.


pedrobaeza (approved):
Fast-tracking as critical

PR #216, opened 2018-05-15
[11.0][MIG] report_py3o (Work in progress)

BT-nfrei:
Hello everyone,
This PR is just a starting point for the migration of this module. At brain-tec, we like this module and we will contribute to its migration to 11.0.
After looking at the technical implementation in 10.0, we still have some doubts about the parser part (report_sxw/rml_parse, contexts). So it would be nice to get some inputs from the initial creators like @lmignon, @alexis-via or @guewen.
In 11.0, it seems quite clear that we have to move away from the SXW/RML technology. So I'm not sure if we should try to keep as much as possible from the current implementation or maybe get some inspiration from report_xlsx which is already migrated to 11.0.
So if you have inputs or advice, please let us know.


lmignon:
@BT-nfrei You're welcome to migrate this module.  I have not yet had the opportunity to look in detail at the implementation of V11 reports and V10 changes but it seems to me that the way to go would be as follows...

To declare our py3o report, we just need to inherit from `ir.actions.report`, add `py3o`into the `report_type` selection, implement the method `render_py3o` and override `_get_report_from_name`

```python
from odoo import api, fields, models, _
from odoo.exceptions import UserError


class ReportAction(models.Model):
    _inherit = 'ir.actions.report'

    report_type = fields.Selection(selection_add=[("py3o", "py3o")])

    @api.model
    def render_py3o(self, docids, data):
        ....

    @api.model
    def _get_report_from_name(self, report_name):
        res = super(ReportAction, self)._get_report_from_name(report_name)
        if res:
            return res
        report_obj = self.env['ir.actions.report']
        qwebtypes = ['py3o']
        conditions = [('report_type', 'in', qwebtypes),
                      ('report_name', '=', report_name)]
        context = self.env['res.users'].context_get()
        return report_obj.with_context(context).search(conditions, limit=1)
```
You should also move all the fields from https://github.com/OCA/reporting-engine/blob/10.0/report_py3o/models/ir_actions_report_xml.py into `ReportAction`

We can keep the implementation of the logic to render the report into the wizard as it's done in 10.0 https://github.com/OCA/reporting-engine/blob/10.0/report_py3o/models/py3o_report.py#L83
All you need to do is to replace the field `ir_actions_report_xml_id` by `ir_actions_report_id`

By keeping the wizard we ease the maintenance of this module across versions and the implementation of `render_py3o` could look like

```python
    @api.model
    def render_py3o(self, docids, data):
        report = self.get_report_from_name(self.report_name)
        return self.env['py3o.report'].create({
                'ir_actions_report_id': report.id
         }).create_report(res_ids, data)
```

PR #228, opened 2018-07-16
[ADD] New module 'mass_reporting'

sebalix:
MASS REPORTING
==============

This module allows you to generate a large volume of reports (like several thousands), optionally concatenate/zip them, and alerts users when the process is done.
It is based on the Job queue provided by the ``connector`` framework to process the reports, and on the standard messaging system of Odoo to send alerts.

Configuration
=============

Go to the *Reporting / Mass reporting / Mass reports* menu to create a new
mass report. Here you have to set the following options:

  - set the data model from which reports should be printed (e.g. 'Invoice'),
  - select the user account used to print the reports,
  - select the reports to print in the right order (e.g. a PDF letter + the PDF invoice),
  - configure the filters to use (e.g. to get only validated invoices)
  - choose the output format:
      * do nothing (simple case, useful if you just want to generate your invoices as attachments)
      * concatenate all files into one,
      * put individual reports into a single ZIP archive
      * or implement your own output format (like sending the produced data by email or SFTP to another server).

Once configured, just push the "Run" button and wait. The configured user will
be notified via the Odoo messaging when the mass report is processed.
The resulting file (if any) will be available in the `File` field.


sebalix:
Isn't maintained anymore, closing.

Issue #231, opened 2018-07-27
[10.0][bi_view_editor] If model name has apostrophe can't generate view


SimoRubi:
Steps:
1. Go to model's tree in Technical > database structure > model;
2. Change a model's name to introduce an apostrophe (e.g. Account -> Account's model)
3. In Dashboard > Custom BI View, create a new view and select a field from the model just edited and save
4. Generate view and get  
<details><summary>Error</summary>
<p>

`Odoo Server Error`

```
Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 642, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 684, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 334, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 327, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 942, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 507, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 896, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 884, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/build/OCA/reporting-engine/bi_view_editor/models/bve_view.py", line 352, in action_create
    self._create_sql_view()
  File "/home/odoo/build/OCA/reporting-engine/bi_view_editor/models/bve_view.py", line 288, in _create_sql_view
    formatted_data = json.loads(self._get_format_data(self.data))
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
[...]

PR #232, opened 2018-07-31
[10.0] bi_view_editor ValueError if apostrophe

SimoRubi:
Solving https://github.com/OCA/reporting-engine/issues/231

### DIFF
--- a/bi_view_editor/models/bve_view.py
+++ b/bi_view_editor/models/bve_view.py
@@ -39,7 +39,7 @@ class BveView(models.Model):
          ('created', 'Created')],
         default='draft',
         copy=False)
-    data = fields.Text(
+    data = fields.Serialized(
         help="Use the special query builder to define the query "
              "to generate your report dataset. "
              "NOTE: To be edited, the query should be in 'Draft' status.")
@@ -62,12 +62,6 @@ class BveView(models.Model):
          'unique(name)',
          _('Custom BI View names must be unique!')),
     ]
-
-    @classmethod
-    def _get_format_data(cls, data):
-        data = data.replace('\'', '"')
-        data = data.replace(': u"', ':"')
-        return data
 
     @api.multi
     def _create_view_arch(self):
@@ -96,7 +90,7 @@ class BveView(models.Model):
                     view_fields.append(field_def)
             return view_fields
 
-        fields_info = json.loads(self._get_format_data(self.data))
+        fields_info = json.loads(self.data)
         view_fields = _get_field_list(fields_info)
         return view_fields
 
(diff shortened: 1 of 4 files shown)


astirpe (approved):
Thanks!

tarteo (approved):
Never knew 10.0 had such field, but it looks like all the tests are passing. Thanks! https://github.com/odoo/odoo/blob/11.0/odoo/fields.py

pedrobaeza (commented):
Isn't needed a migration script for moving/adapting previous contents to new field type?

SimoRubi:
@pedrobaeza migration script created, I also proposed the very same method to be added in openupgradelib with https://github.com/OCA/openupgradelib/pull/115

bi_view_editor/models/bve_view.py
```diff
@@ -96,7 +90,7 @@ def _get_field_list(fields_info):
                     view_fields.append(field_def)
             return view_fields
 
-        fields_info = json.loads(self._get_format_data(self.data))
+        fields_info = json.loads(self.data)
```
StefanRijnhart, line 93: Isn't the idea of using a fields.Serialized that it loads the json for you?

>>> cat._fields['testser'].type
'serialized'
>>> cat.write({'testser': {'a': 1}})
True
>>> type(cat.testser)
<type 'dict'>
>>> cat.testser
{u'a': 1}
```

SimoRubi: Thanks for your review @StefanRijnhart, it took me a while to understand that myself, but actually I'm only using the field for its escaping power, here is what happens:
1. In the client, the query is composed and `JSON.stringify`ed in order to send it to the server (I think the only way to send field values to the server is using strings);
2. In the server, the string is saved into the serialized field, it applies `json.dumps` on the string (then it is double-escaped);
3. When I access `self.data`, the serialized field itself `json.loads` it, but it is still a string, then I have to `json.loads` that *again* too

I recognize that it would be cleaner to escape the string only once, but I should override the `create` and intervene between steps 1 and 2 in order to load the field correctly (`json.load`ing that already).

astirpe:
Tested the migration script and it worked fine :+1:

pedrobaeza:
@StefanRijnhart it's OK for you too?

SimoRubi:
@StefanRijnhart is this ok for you?

StefanRijnhart (approved):
I'm struggling to accept the double serializion but there are worst things in the world and I just don't have the time ;-)

SimoRubi:
Thanks @StefanRijnhart for the review, I know, that part is no good code..
The other way around (intervening in the `create` method, I can't think to any other solution) is cleaner from the logic point of view but, in my opinion, much worse from the code point of view.

Anyway, I have added tests to make up for it :)

StefanRijnhart:
Sure! And you fixed the bug so that's good.

astirpe:
Can we merge this one?

PR #233, opened 2018-08-14
[10.0] report_xlsx new api

luc-demeyer:
as promised to @pedrobaeza  in https://github.com/OCA/reporting-engine/pull/204:
PR with support for new api.

### DIFF
--- a/report_xlsx/report/report_xlsx.py
+++ b/report_xlsx/report/report_xlsx.py
@@ -5,7 +5,7 @@
 from cStringIO import StringIO
 
 from odoo.report.report_sxw import report_sxw
-from odoo.api import Environment
+from odoo import api
 
 import logging
 _logger = logging.getLogger(__name__)
@@ -18,8 +18,9 @@ except ImportError:
 
 class ReportXlsx(report_sxw):
 
+    @api.cr_uid_ids_context
     def create(self, cr, uid, ids, data, context=None):
-        self.env = Environment(cr, uid, context)
+        self.env = api.Environment(cr, uid, context)
         report_obj = self.env['ir.actions.report.xml']
         report = report_obj.search([('report_name', '=', self.name[7:])])
         if report.ids:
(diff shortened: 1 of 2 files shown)


report_xlsx/report/report_xlsx.py
```diff
@@ -18,8 +18,9 @@
 
 class ReportXlsx(report_sxw):
 
+    @api.cr_uid_ids_context
     def create(self, cr, uid, ids, data, context=None):
```
pedrobaeza, line 22: Why don't we directly put this `create` in new API format?

luc-demeyer: in that case it fails because of the following call in standard addons:
> /opt/odoo/odoo/oc/odoo/addons/base/ir/ir_actions.py(234)render_report()
-> return report.create(self._cr, self._uid, res_ids, data, context=self._context)

pedrobaeza: But then the problem is on that line, not here. It's an incorrect removal of old API. A PR in Odoo for fixing it for sure will be accepted.

sbidoul: Just a question: report_xslx is working fine in v10 as it is now, why do we need to change anything?

pedrobaeza: Old API support was removed in v10, so it's logical to switch this properly for avoiding side effects.

luc-demeyer: I am not sure if Odoo will accept such a PR since technically everything works fine and the focus from Odoo is already on V12.
With this PR this create call can be called from old and new API hence we bypass the issue without touching the code from Odoo. I have done a couple of tests on a number of reports and didn't find any issue when adding the  '@api.cr_uid_ids_context' decorator.
But as stephan says: we can also decide not to change anything and hence close this PR without merge since report_xlsx and report_xlsx_helper work fine since they both use the old API.

pedrobaeza: OK, I let @sbidoul to decide.

sbidoul: Since the create method is declared without decorator in the base class, I think it's safer to keep it as is. So I suggest to close this PR.

pedrobaeza: OK, let's close it.

PR #234, opened 2018-08-15
[8.0][ADD] report_xlsx_helper

luc-demeyer:
backport of 10.0 report_xlsx_helper module


luc-demeyer:
@astirpe : can you review (I fixed a couple of flake8/pylint issues in your modules to greenify this PR).

report_xlsx_helper/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===========================
+Excel report engine helpers
```
pankk, line 6: Rename helpers to helper

report_xlsx_helper/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009-2018 Noviat.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Report xlsx helpers',
```
pankk, line 6: Rename helpers to helper

pankk (changes requested):

pankk (approved):
Successful functional test with six custom xlsx reports using this helper.

report_xlsx_helper/__openerp__.py
luc-demeyer: This module add a set tools to facilitate the creation of XLSX files, hence I think 'helpers' in the README is more appropriate.

pankk: Ok, I didn't think of it that way. I began to think of it as a "helper module". I think usually such a module would be named in plural form anyway, but the singular convention is a good measure for headache avoidance. But I now agree that the description should use the plural form to be accurate.

PR #235, opened 2018-08-16
[11.0][MIG] report_qweb_pdf_watermark

schout-it:
Migration of watermark module

pypdf does not work in python3 so hard requirement for pypdf2

Maybe someone can take a look at doing this with html/css (not my strong suite),
would probably be faster then this implementation.


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

schout-it:
second pull request send which preserves the commit history,
would have become a mess if I'd used this one

PR #238, opened 2018-08-25
Fix report_xlsx for printing from wizards

### LINKED ISSUE #237: 11.0 Bug affecting report_xml and report_xlsx unable to use with wizard
Ordinarily wizards pass their options in data, however the way qwebactionmanager.js is structured `data` is never passed to `report_routes` due to a switch on context.active_ids.

I am no js expert, however to resolve in report xlsx I changed qwebactionmanager.js to:
```javascript
ActionManager.include({
    ir_actions_report: function (action, options){
        var self = this;
        var cloned_action = _.clone(action);
        if (cloned_action.report_type === 'xlsx') {
            framework.blockUI();
            var report_xlsx_url = 'report/xlsx/' + cloned_action.report_name;
            if (_.isUndefined(cloned_action.data) || _.isNull(cloned_action.data) || (_.isObject(cloned_action.data) && _.isEmpty(cloned_action.data))) {
                if(cloned_action.context.active_ids) {
                    report_xlsx_url += '/' + cloned_action.context.active_ids.join(',');
                }
            } else {
                report_xlsx_url += '?options=' + encodeURIComponent(JSON.stringify(cloned_action.data));
                report_xlsx_url += '&context=' + encodeURIComponent(JSON.stringify(cloned_action.context));
            }
            self.getSession().get_file({
                url: report_xlsx_url,
                data: {data: JSON.stringify([
                    report_xlsx_url,
                    cloned_action.report_type
                ])},
                error: crash_manager.rpc_error.bind(crash_manager),
                success: function (){
                    if(cloned_action && options && !cloned_action.dialog){
                        options.on_close();
                    }
                }
            });
            framework.unblockUI();
            return;
        }
        return self._super(action, options);
    }
});
});
```
Which roughly mimics make_report_url without the map function https://github.com/odoo/odoo/blob/5dc180a3ea50083c88399a58ee07cd130fa8289d/addons/web/static/src/js/report/qwebactionmanager.js#L43

The only real change is to check for data before defaulting to context.active_ids.

I'm really lost on js so haven't made PR but can do.

gdgellatly:
Resolves #237 by checking for data first in calling action typically `report_action` from wizard button.

Note report_xml will need the same fix for at least qwebactionmanager.js

### DIFF
--- a/report_xlsx/__manifest__.py
+++ b/report_xlsx/__manifest__.py
@@ -7,9 +7,9 @@
     'author': 'ACSONE SA/NV,'
               'Creu Blanca,'
               'Odoo Community Association (OCA)',
-    'website': "http://github.com/oca/reporting-engine",
+    'website': "https://github.com/oca/reporting-engine",
     'category': 'Reporting',
-    'version': '11.0.1.0.2',
+    'version': '11.0.1.0.3',
     'license': 'AGPL-3',
     'external_dependencies': {
         'python': [
--- a/report_xlsx/report/report_xlsx.py
+++ b/report_xlsx/report/report_xlsx.py
@@ -17,8 +17,30 @@ except ImportError:
 class ReportXlsxAbstract(models.AbstractModel):
     _name = 'report.report_xlsx.abstract'
 
+    def _get_objs_for_report(self, docids, data):
+        """
+        Returns objects for xlx report.  From WebUI these
+        are either as docids taken from context.active_ids or
+        in the case of wizard are in data.  Manual calls may rely
+        on regular context, setting docids, or setting data.
+
+        :param docids: list of integers, typically provided by
+            qwebactionmanager for regular Models.
+        :param data: dictionary of data, if present typically provided
+            by qwebactionmanager for TransientModels.
+        :param ids: list of integers, provided by overrides.
+        :return: recordset of active model for ids.
+        """
+        if docids:
+            ids = docids
+        elif data and 'context' in data:
+            ids = data["context"].get('active_ids', [])
+        else:
+            ids = self.env.context.get('active_ids', [])
+        return self.env[self.env.context.get('active_model')].browse(ids)
+
     def create_xlsx_report(self, docids, data):
-        objs = self.env[self.env.context.get('active_model')].browse(docids)
+        objs = self._get_objs_for_report(docids, data)
         file_data = BytesIO()
         workbook = xlsxwriter.Workbook(file_data, self.get_workbook_options())
         self.generate_xlsx_report(workbook, data, objs)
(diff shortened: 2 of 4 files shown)


report_xlsx/__manifest__.py
```diff
@@ -9,7 +9,7 @@
               'Odoo Community Association (OCA)',
     'website': "http://github.com/oca/reporting-engine",
```
elicoidal: https

elicoidal (approved):
LGTM

report_xlsx/report/report_xlsx.py
```diff
@@ -17,8 +17,37 @@
 class ReportXlsxAbstract(models.AbstractModel):
     _name = 'report.report_xlsx.abstract'
 
+    def _get_objs_for_report(self, docids, data, ids=None):
```
pedrobaeza: Why the extra `ids` arg?

gdgellatly: The only reason was to allow for a clean inheritance chain if a wizard in particular wanted to use something other than data["context"]["active_ids"] to determine the active records for a report, it could pass ids and skip the logic in this functions.

I have no personal use for it.

EDIT: Come to think of it, it could just set docids anyway and it would stay clean, so what do you think? Remove it?

gdgellatly: @pedrobaeza I removed it, you are right, not really needed.

report_xlsx/__manifest__.py
gdgellatly: @elicoidal Wasn't part of my commit but had to update for Pedro so just did this as well

pedrobaeza (approved):

report_xlsx/report/report_xlsx.py
```diff
@@ -17,8 +17,30 @@
...
+            qwebactionmanager for regular Models.
+        :param data: dictionary of data, if present typically provided
+            by qwebactionmanager for TransientModels.
+        :param ids: list of integers, provided by overrides.
+        :return: recordset of active model for ids.
+        """
+        if docids:
+            ids = docids
+        elif data and 'context' in data:
+            ids = data["context"].get('active_ids', [])
```
sbidoul, line 37: would it make sense to get `active_model` from `data['context']` instead of `self.env.context` in this case

pedrobaeza: Agh, sorry, I have already merged it.

gdgellatly:
@sbidoul @pedrobaeza You would think so, but actually no.  When you call report_action it sets the active_model context to the model of the report and doesn't actually populate data["context"].  It is the way it is handled upstream.  I'll find the reference code I used and link here later.

Ahh I took the ids logic from upstream, actually it is render_xlsx of this module that sets active_model.  Other renderings e.g. render_qweb_pdf use self.model of the report name to determine model, which is probably better, but a behaviour change.

sbidoul:
Ok, thanks! :+1:

PR #243, opened 2018-09-24
[11.0][MIG]report_xlsx_helper

luc-demeyer:


report_xlsx_helper/report/test_partner_report_xlsx.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009-2018 Noviat.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class TestPartnerXlsx(models.AbstractModel):
```
grindtildeath, line 12: Is this class only required for the test ? As I understand, this shouldn't be loaded when installing the module ?
Moreover this looks like this is a duplicate from  report_xlsx_helper_demo/report/partner_export_xlsx.py so I'm not sure this should be loaded all the time.
Anyway files and classnames beginning with test* should be reserved for test files (https://docs.pytest.org/en/2.9.1/goodpractices.html)

```
================================================================================== test session starts ==================================================================================
platform linux -- Python 3.5.3, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /opt/odoo, inifile:
plugins: odoo-0.4.2, cov-2.5.1
collected 1 item                                                                                                                                                                        

odoo/external-src/reporting-engine/report_xlsx_helper/tests/test_report_xlsx_helper.py .

=================================================================================== warnings summary ====================================================================================
external-src/reporting-engine/report_xlsx_helper/tests/test_partner_report_xlsx.py::TestPartnerXlsx
  cannot collect test class 'TestPartnerXlsx' because it has a __init__ constructor

-- Docs: http://doc.pytest.org/en/latest/warnings.html
========================================================================= 1 passed, 1 warnings in 0.56 seconds ==========================================================================
```

luc-demeyer: @grindtildeath 
I fully agree with your comment but you'll see that the underlying report_xlsx module does exactly the same.
In order to execute the test the TestPartnerXlsx class must be loaded and added to the registry and this doesn't happen when calling the method only from within the tests folder.

I would need to dig into the Odoo kernel in order to figure out how to do this and it's clear that our friend from Acsone have taken the same approach as me when adding a unit test to the 11.0 report_xlsx module.

I propose that we move forward in two steps;
Step 1) merge this module into the 11.0 reporting-engine repo so that we move forward on the port of modules that depend on it.
Step 2) make a PR that moves the test classes of boty report_xlsx and report_xlsx_helper to the tests folder and adds these classes dynamically to the Odoo registry as test execution time.

tschanzt (approved):
Looks good to me.

report_xlsx_helper/report/test_partner_report_xlsx.py
grindtildeath: @simahawk Maybe you have a solution for this ?

simahawk: if this is the original code and it works I agree w/ Luc: let's unlock the rest.
I'd add a big TODO mentioning the refactoring related to tests and then I'd propose another PR that register models on the fly only for tests. We have a couple of examples out there. I'll provide them.

luc-demeyer: TODO added.

grindtildeath (approved):

PR #247, opened 2018-09-28
[IMP] report_qweb_element_page_visibility: Also be able to use the class in the report body

thomaspaulb:
Not only in header and footer of the report.

[subst is not called in body](https://github.com/OCA/OCB/blob/8.0/addons/report/models/report.py#L256)

### DIFF
--- a/report_qweb_element_page_visibility/views/layouts.xml
+++ b/report_qweb_element_page_visibility/views/layouts.xml
@@ -13,7 +13,7 @@
                             subst();
                         }
                         var vars = {};
-                        var x = document.location.search.substring(1).split('&amp;amp;');
+                        var x = document.location.search.substring(1).split('&amp;');
                         for (var i in x) {
                             var z = x[i].split('=', 2);
                             vars[z[0]] = unescape(z[1]);
@@ -38,8 +38,9 @@
                         };
                         for (var _class in operations) {
                             var y = document.getElementsByClassName(_class);
-                            for (var j=0; j&lt;y.length; ++j)
-                            operations[_class](y[j]);
+                            for (var j=0; j&lt;y.length; ++j) {
+                                operations[_class](y[j]);
+                            }
                         }
                     }
                 </script>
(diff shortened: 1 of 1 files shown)


thomaspaulb:
@hbrunn If you can, please review.

report_qweb_element_page_visibility/views/layouts.xml
```diff
@@ -1,48 +1,50 @@
...
+                <attribute name="onload">subst_wrapper();</attribute>
+            </xpath>
+            <xpath expr="//t[@t-if='subst is True']" position="after">
+                <script>
+                    function subst_wrapper() {
+                        if (typeof subst === "function") {
+                            subst();
+                        }
+                        var vars = {};
+                        var x = document.location.search.substring(1).split('&amp;amp;');
```
hbrunn: you can avoid this kind of stuff by wrapping all javascript code in a [CDATA](https://en.wikipedia.org/wiki/CDATA) section

hbrunn (approved):
refactoring looks good. But it seems I'm overlooking something, your PR title looks like you add functionality, but I don't see this?

report_qweb_element_page_visibility/views/layouts.xml
thomaspaulb: Unfortunately, that made "wkhtmltopdf" not recognize the Javascript function anymore and thus not calling it.

thomaspaulb:
@hbrunn The added functionality is that "subst_wrapper()" is called for header, body and footer, whereas the before modification of "subst()" was only called for header and footer. (Because the subst function is included in the HTML only when subst=True, and subst=False for body rendering, see also the code link referenced above).

hbrunn:
ah, thanks. Positive review stands

pedrobaeza:
Can you check Travis and runbot status?

thomaspaulb:
Rebased. Seems that Travis/runbot errors are not mine

silverf0x0:
@thomaspaulb Hi Thomas, I was looking for this kind of improvement, as I really need to show elements in the body report, rather then header and/or footer. I've tested your changes, but unfortunately it still show the element only in the header or in the footer.

silverf0x0:
@hbrunn You have given positive review to this commit, which means it works. Am I missing something to make it work in the body of my report?

Rad0van:
@thomaspaulb nice addition. Why didn't this go through? Is it worth adapting to 12.0/13.0? Thanx.

PR #248, opened 2018-10-08
[MIG] 12.0 report xlsx

ovnicraft:
TODO: 
* [x] refactor to be compatible with new action manager on v12
* [x] test module


ovnicraft:
Error on travis is not related to this.
Full log on runbot: https://runbot3.odoo-community.org/runbot/static/build/3341394-248-1e72f4/logs/job_20_test_all.txt

report_xlsx/static/src/js/report/qwebactionmanager.js
```diff
@@ -0,0 +1,47 @@
...
+// License AGPL-3.0 or later (https://www.gnuorg/licenses/agpl.html).
+odoo.define('report_xlsx.report', function(require){
+'use strict';
+
+var ActionManager= require('web.ActionManager');
+var crash_manager = require('web.crash_manager');
+var framework = require('web.framework');
+
+ActionManager.include({
+    ir_actions_report: function (action, options){
```
gdgellatly: Are we sure about this migration? AFAICT this function was completely refactored out.  There is no way this module should work straight from 11.0 to 12.0 without major changes here.  I would have thought at least intercepting _executeReportAction, and to do cleanly also _triggerDownload and _downloadReport rather than just contain the action in one monolithic function. 

But I wonder if we can't just intercept at _makeReportURLs and set an url there and then intercept again at _triggerDownload and pass to a custom _downloadXLSX function or something and ignore a lot of this.

pedrobaeza:
Alternative migration in #249

RickHunterEc:
> Alternative migration in #249

Hi @ovnicraft, please, can you check this branch for avoid duplicate effort. 

Thanks.

pedrobaeza:
@RickHunterEc take into account that this migration was before yours, so unless something dramatically, this one has priority.

report_xlsx/static/src/js/report/qwebactionmanager.js
ovnicraft: @gdgellatly Thanks for your review, so i will try to intercept `_makeReportURLs` and next functions at first try.

RickHunterEc:
For us is high priority. All reports and documents by four american countries localization in v 12 need this module. The module was tested with 90 reports and any problem found.

ovnicraft:
@rickHunterEc in your migration take in care @gdgellatly about _makeReportUrl ?

gdgellatly:
@ovnicraft @RickHunterEc the current action is a hangover from v10, since then Odoo has modularised a lot of the webclient especially the actionmanager.  So while a monolithic _executeReportAction will work, its not really the right way anymore.  So while @RickHunterEc  PR will work fine it isn't written in the way v12 web client is written. I haven't looked too closely but actually I think it is entirely feasible to not override _executeReportAction at all.  Just the make urls, maybe triggerDownload to call a downloadxlsx and then a small function to do just that base off downloadReport.  But we should use the first *report* PR to get things right as most of them modify actionmanager very similarly.

ovnicraft:
@gdgellatly i am working on considering the refactora on v12

ovnicraft:
@gdgellatly New ActionManager looks really good, so `_makeReportUrls` goes fine to add new type, `_triggerDownload()` has type arg and it allow to call custom XLSX :-) 

Let see tests and runbot

ovnicraft:
@gdgellatly just a Q: on controller, why goes directly on `report_routes()` and not on `report_download()` ?

ovnicraft:
@RickHunterEc Please test this with your reports to validate it

RickHunterEc:
> @RickHunterEc Please test this with your reports to validate it

Done and all Ok. Thanks for you help @ovnicraft . 
@elicoidal  and @gdgellatly, thanks for you reviews.
@pedrobaeza , close my PR.

ovnicraft:
@RickHunterEc thank you, waiting for another good review and merge 👍

ovnicraft:
@gdgellatly @elicoidal your help for review this

ovnicraft:
any updates here ?

gdgellatly (approved):
Static review of changes to actionmanager.  Comfortable with straight porting of rest.

ovnicraft:
@pedrobaeza possible merge this ?

report_xlsx/models/ir_report.py
```diff
@@ -0,0 +1,33 @@
...
+    report_type = fields.Selection(selection_add=[("xlsx", "XLSX")])
+
+    @api.model
+    def render_xlsx(self, docids, data):
+        report_model_name = 'report.%s' % self.report_name
+        report_model = self.env.get(report_model_name)
+        if report_model is None:
+            raise UserError(_('%s model was not found' % report_model_name))
+        return report_model.with_context({
+            'active_model': self.model
```
RickHunterEc: @ovnicraft , need add this:
'lang': self.env.user.partner_id.lang

RickHunterEc: @ovnicraft, please, can yo add this code:
```suggestion
            'active_model': self.model,
            'lang': self.env.user.partner_id.lang
```
because, translation report is wrong. Need pass language by context. Thanks.

gdgellatly: @RickHunterEc I am less than convinced by this.  lang is usually already set in context in UI or the caller may have specifically set context lang to something other than user lang (e.g. partner.lang in an email attachment). I believe the correct way to do this is 

`report_model.with_context(active_model=self.model)`

to stop the existing context from being overwritten.

ovnicraft: Yes, language its always on context.
Force the lang on user could break the normal use of this.

gdgellatly: @ovnicraft You will still need to make the change.

`with_context(dict())` sets a new context effectively removing lang from the dict..
`with_context(**kwargs)` updates current context so either pass the dict with ** or make the args active_model=self.model etc

ovnicraft: > @RickHunterEc I am less than convinced by this. lang is usually already set in context in UI or the caller may have specifically set context lang to something other than user lang (e.g. partner.lang in an email attachment). I believe the correct way to do this is
> 
> `report_model.with_context(active_model=self.model)`

This is already done.
> 
> to stop the existing context from being overwritten.

gdgellatly: weird, its still showing the old code when I look at the file.

ovnicraft: Check it https://github.com/ovnicraft/reporting-engine/blob/12.0-mig-report_xlsx/report_xlsx/models/ir_report.py#L20

gdgellatly: @ovnicraft It still shows the same using the dict constructor method instead of kwargs.  The dict constructor overwrites the entire context, keyword args updates the existing.  So using dict constructor as shown there, you effectively remove any context lang from the dict.

I had the same issue on base_report_to_printer.

RickHunterEc: @ovnicraft  and @gdgellatly , I agree with you.

ovnicraft: I'll fix it on this weekend

fmdl (approved):

pedrobaeza:
@gdgellatly do you confirm everything is OK?

report_xlsx/models/ir_report.py
anhvu-sg: Hello, is there any update about this task?

pedrobaeza:
I still don't see the final confirmation from @gdgellatly for merging.

gdgellatly:
@pedrobaeza It still requires fixing IMO. Otherwise any passed context will be lost.

pedrobaeza:
Perfect, that was what I'm afraid of. Maybe you can make a PR fixing this as @ovnicraft is not answering.

gdgellatly:
I cant this week. Away until 20th.

pedrobaeza:
OK, any of the other reviewers of this PR can superseed this PR for fixing the problem, or wait for the final solution by @ovnicraft or @gdgellatly .

report_xlsx/models/ir_report.py
ovnicraft: I recently push a commit with `dict()` as @gdgellatly recommended, BTW i am sure if this change fix problem so i am not passing active model, maybe i need call `with_context(active_model=self.model)` after `dict()` constructor call

report_xlsx/models/ir_report.py
```diff
@@ -0,0 +1,31 @@
...
+
+    report_type = fields.Selection(selection_add=[("xlsx", "XLSX")])
+
+    @api.model
+    def render_xlsx(self, docids, data):
+        report_model_name = 'report.%s' % self.report_name
+        report_model = self.env.get(report_model_name)
+        if report_model is None:
+            raise UserError(_('%s model was not found' % report_model_name))
+        return report_model.with_context(dict(active_model=self.model)).create_xlsx_report(docids, data)
```
gdgellatly: `report_model.with_context(dict(active_model=self.model))` -> `report_model.with_context(active_model=self.model)`

@ovnicraft this is the part I meant

ovnicraft: done

alexey-pelykh (changes requested):
@ovnicraft please fix the lint error
```
report_xlsx/models/ir_report.py:19:80: E501 line too long (98 > 79 characters)
```

alexey-pelykh (approved):

alexey-pelykh:
@ovnicraft also please:
 * split readme as per https://github.com/OCA/maintainer-tools/tree/master/template/module/readme
 * squash some migration-related commits where there's no sense of keeping changes separate

ovnicraft:
@alexey-pelykh i can split readme later by now and allow continue migrations for other i consider ready to merge.

alexey-pelykh:
But this is a requirement from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0 and it takes like 5-10 minutes to do that

pedrobaeza:
I have squashed properly commit history and add the README by fragments. Waiting CIs for merging.

alexey-pelykh:
@pedrobaeza one last click left 👍

anhvu-sg:
Thanks all (Y)

PR #252, opened 2018-11-06
[ADD] Module report_controller

hmrodrigues:
Reimplemented the controller (pdf) report type, missing since Odoo 11.0
It duplicates some upstream code, since upstream has the base types hardcoded,
if the PR https://github.com/odoo/odoo/pull/28365 is merged and backported
into 12.0, we will need to refactor in order to remove the duplicated
code.

### DIFF
--- a/report_controller/README.rst
+++ b/report_controller/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Report Controller
+=====================
+
+This modules brings back the controller type for reports missing since Odoo 11.0
+
+
+Usage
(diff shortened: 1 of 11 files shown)


pedrobaeza:
I don't get the goal of this module. Why do you want to reimplement it? What do you gain with this? All of these questions should be included in the README.

hmrodrigues:
> I don't get the goal of this module. Why do you want to reimplement it? What do you gain with this? All of these questions should be included in the README.

Hello Pedro, the goal of this module is to print reports based on data from a external service.

For example, in Portugal, all customer invoices must be certified and the PDF must be printed by the certified software. Since only the product owners can certify their software, in Odoo case is Odoo SA, the certification software is usually external to Odoo, making the Odoo report legally invalid. To print the external PDF using the print menu, we could to call a controller to fetch the PDF from the external service.

pedrobaeza:
OK, thanks for the explanation.

PR #186, opened 2018-01-08
[11.0] [MIG] report_qweb_element_page_visibility

ghost:

PR #257, opened 2018-11-26
[MIG] Migration of report_qweb_element_page_visibility to odoo 11

BT-dherreros:
Migration of the module from version 10 to version 11 of odoo


pedrobaeza:
There's already other migration in #186, and yours is very dirty. Closing. Please review that one instead or superseed it.

PR #217, opened 2018-05-15
[11.0][MIG] report_py3o (Work in Progress)

BT-nfrei:
Hello everyone,
This PR is just a starting point for the migration of this module. At brain-tec, we like this module and we will contribute to its migration to 11.0.
After looking at the technical implementation in 10.0, we still have some doubts about the parser part (report_sxw/rml_parse, contexts). So it would be nice to get some inputs from the initial creators like @lmignon, @alexis-via or @guewen.
In 11.0, it seems quite clear that we have to move away from the SXW/RML technology. So I'm not sure if we should try to keep as much as possible from the current implementation or maybe get some inspiration from report_xlsx which is already migrated to 11.0.
So if you have inputs or advice, please let us know.

PR #258, opened 2018-11-26
[12.0][MIG] report_py3o and report_py3o_fusion_server

lmignon:
This migration is based on the files of version 10.0 in order to include the latest developments made on the latest stable version of these modules.
It was not possible to duplicate some of the work done in #217 for migration to v11 because the history of this PR is not usable. In addition, #217 does not include the changes done in 10.0 to split report_py3o and report_py3o_fusion_server.
In version 10.0, the context accessible from a template was based on a `rml_parse` object. In 12.0, the context is a simple dictionary with the minimum amount of information necessary to render a template. I chose not to use the code and therefore the methods provided by `rml_parse` in 12.0 because it seems to me that they no longer made any sense in v12.0. If necessary, it will always be possible to extend the default context provided by default according to the needs that will be detected in future implementation of the module (hooks are however available to extend it to your own needs)


lmignon:
cc @hbrunn @alexis-via @sbidoul 
Once reviewed, I'll rework the commit history to only keep one commit for the migration.

sbidoul:
@lmignon I pushed a fix with the setup directory

lmignon:
Thank you @sbidoul

sbejaoui:
@lmignon , I pushed an update of i18n. Got some issues with --i18n-overwrite.

sbidoul (approved):
Works fine (functional test).

report_py3o/README.rst
```diff
@@ -0,0 +1,217 @@
...
+format_multiline_value(string)
+    Generate the ODF equivalent of ``<br/>`` and ``&nbsp;`` for multiline fields (ODF is XML internally, so those would be skipped otherwise)
+html_sanitize(string)
+    Sanitize HTML string
+time
+    Python's ``time`` module
+display_address(partner)
+    Return a formatted string of the partner's address
+formatLang(value, digits=None, date=False, date_time=False, grouping=True, monetary=False, dp=False, currency_obj=False)
+    Return a formatted numeric, monetary, date or time value according to the context language and timezone
```
sbidoul, line 168: @lmignon would it make sence to document `o_format_date`, `o_format_lang` here?

njeudy (approved):
Functionnal test and works like it should :)

report_py3o/models/py3o_report.py
```diff
@@ -0,0 +1,360 @@
...
+    @api.multi
+    def _merge_results(self, reports_path):
+        self.ensure_one()
+        filetype = self.ir_actions_report_id.py3o_filetype
+        if not reports_path:
+            return False, False
+        if len(reports_path) == 1:
+            return reports_path[0], filetype
+        if filetype == formats.FORMAT_PDF:
+            return self._merge_pdf(reports_path), formats.FORMAT_PDF
```
BT-astauder, line 338: Great job! Thank you very much.

This function does not exists anymore, but was removed in 11.0.
Function in 10.0:
https://github.com/odoo/odoo/blob/e1767490e939233c050f796f4ae4f6db627d53c3/addons/report/models/report.py#L559
Similar functionality wor QWeb starting from 11.0:
https://github.com/odoo/odoo/blob/a7cbac305ed8645ce8ce67210d2ca5d723a891f0/odoo/addons/base/ir/ir_actions_report.py#L481

As long as this is not implemented the printing for multiple records in a listview will not work.

nicoclavier:
Hello and congrats on the good work, 
Any way we know when is this PR going to be merged ?
Thank you,
Nicolas

report_py3o_fusion_server/models/ir_actions_report.py
```diff
@@ -0,0 +1,44 @@
...
+    _inherit = 'ir.actions.report'
+
+    @api.multi
+    @api.constrains("py3o_is_local_fusion", "py3o_server_id", "py3o_filetype")
+    def _check_py3o_server_id(self):
+        for report in self:
+            if report.report_type != "py3o":
+                continue
+            is_native = Formats().get_format(report.py3o_filetype).native
+            if ((not is_native or not report.py3o_is_local_fusion) and
```
sbidoul:
```suggestion
            if (not is_native and not report.py3o_is_local_fusion and
```

We need to change this condition, otherwise it is impossible to create new PDF reports with local fusion when `report_py3o_fusion_server` is installed. Put another way the previous condition means all new reports created must have  a fusion server defined when `report_py3o_fusion_server` which is not practical.

sbidoul: @lmignon @hbrunn may I summon your opinion on this one? Thanks in advance.

hbrunn: in 10 it looks like this: https://github.com/OCA/reporting-engine/blob/10.0/report_py3o_fusion_server/models/ir_actions_report_xml.py#L27 - so at least it's consistent ;-)

Can't we just leave out the is_native check entirely? If it's ie a `.doc` and there's no rendering server, the libreoffice code path will take care of it I think, right? How about some tests for the different cases?

hbrunn: ah, now I understand my own code again. Check out https://github.com/OCA/reporting-engine/blob/b81219abd33d56e6068212d5bb10587b06d3db95/report_py3o_fusion_server/models/py3o_report.py#L35 - this is where those things are relevant, and the idea is that you tick the local fusion box when you have the server module installed but you want to force libreoffice conversion instead of the server. So I rectify and now this the code is fine as it is

sbidoul: @hbrunn I recall now that local fusion and local format conversion is two different things. You can do local fusion, and use the py3o server to do the format conversion. Or delegate fusion to the py3o server too.

So yeah, the code is probably fine as is.

Yet we still have a problem. Let me explain our situation.
- we have an Odoo database with report_py3o_fusion_server installed
- we try to install a new module that attempts to create a new py3o PDF report; that module does not know about report_py3o_fusion_server, because it is normally used with local conversion, so it does not attach the report to a fusion server => that installation fails because the report is not native and there is no server

Not sure what to do... may be have a notion of default fusion server that is automatically set when creating new non native report when report_py3o_fusion_server is installed?

hbrunn: hmmm. Make it an onchange warning maybe? Or default local_fusion to true?

lmignon: @hbrunn @sbidoul The default for `py3o_is_local_fusion` is `True`. The problem is that before the split, report_py3o/report_py3O_fusion_server,  we assumed that there was no libreoffice runtime available on the machine. Therefore a py3o_server was required to transform the result of the fusion into the expected format if not native. IMO, we should have a way to detect into py3o_report that the libreoffice runtime is available and raise this error only if the runtime is not available.

alexis-via:
@lmignon Thanks for this great work. I'm currently working on porting https://github.com/akretion/odoo-py3o-report-templates to v12 and I have a problem: I can't use "user.something" in an expression of my template. It is important for headers and footers, where you can't access "objects" (so, in this case, I use user.company_id.partner_id.image for get the logo of the company for example)

```
 File "<string>", line 1, in <Expression "format_multiline_value(user.company_id._display_report_header([REMOVED])">
  File "/usr/local/lib/python3.6/dist-packages/genshi/template/eval.py", line 311, in lookup_name
    val = cls.undefined(name)
  File "/usr/local/lib/python3.6/dist-packages/genshi/template/eval.py", line 414, in undefined
    raise UndefinedError(key, owner=owner)
genshi.template.eval.UndefinedError: "user" not defined
```

lmignon:
@alexis-via I'm working on it to fix the remaining issues. 

* [X] Merge PDF  https://github.com/OCA/reporting-engine/pull/258#pullrequestreview-192189986 implemented into https://github.com/OCA/reporting-engine/pull/258/commits/213895ed0c7983d2f50eceb89675aa60f6695b57 
* [X] Constrains when py3o_report_fusion_server is installed https://github.com/OCA/reporting-engine/pull/258#pullrequestreview-197672094 implemented into https://github.com/OCA/reporting-engine/pull/258/commits/876aa27ef219a7f57fafb2ee56061570645cc35b

I add your problem 

* [X] `user`must be available into the report context. implemented into https://github.com/OCA/reporting-engine/pull/258/commits/762c52118fca4f7e7dc7ca6fa990733092dbac20

alexis-via:
It seems there is a pb in your last commit
```
File "/home/alexis/new_boite/dev/reporting-engine2/report_py3o/models/_py3o_parser_context.py", line 35, in __init__
    'user': self.env.user,
AttributeError: 'Py3oParserContext' object has no attribute 'env'
```

lmignon:
@alexis-via Sorry. should be fixed

report_py3o/models/_py3o_parser_context.py
```diff
@@ -32,6 +32,8 @@ def __init__(self, env):
         self._env = env
 
         self.localcontext = {
+            'user': self._env.user,
+            'lane': self._env.lang,
```
sbidoul: s/lane/lang/

alexis-via:
Thanks !
On another topic, I'm having some difficulties using the new method o_format_lang() ; it's not clear to me how I should pass the first argument "env".

BTW, I would like to contribute a method for selection fields, to make it easy to display the label of a selection field in the language of the object. I'll try to find time to make a PR for that soon...

alexis-via:
OK, i see that you fixed o_format_lang() in your commit a few minutes ago... I can now use it. But I still have to understand how it read the lang, because it doesn't take the decimal separator that it is supposed to use...

lmignon:
@alexis-via The lang should be the one from the user....

report_py3o/models/_py3o_parser_context.py
lmignon: done. Thank you

lmignon: We should test all the methods provided by the parser context into the demo template to avoid regressions and also tests this part of code...

report_py3o_fusion_server/models/ir_actions_report.py
lmignon: @sbidoul and @hbrunn I've implemented a non blocking approach of this problem. In short:
* The constrains only fails if not local fusion and not fursion server
* We log at server startup an display a warning message on the report if the report should not work due to the unavailability of the libreoffice runtime or a fusion server
* We raise a proper error message when the libreoffice runtime is called to generate a report and is not available.

With these changes the installation of new py3o reports in non native format should work even if py3o_report_fusion_server is installed and a libreoffice runtime is not available...

alexis-via:
@lmignon Yes, it's what I currently have, the lang is the one of the user. Now, we want the lang to be the one of the supplier of the purchase order. Here is what I used to do:
at the beginning of the report:
```
py3o://for="obj in objects"
py3o://with="lang=obj.partner_id.lang or 'en_US'"
py3o://with="o=obj.with_context(lang=lang)"
```
then, in the report, I use "o.partner_id.name", etc..

To display the date, I use:
```
py3o://function="o_format_date(o.date_order, lang_code=lang)"
```
and it works well, I get the date in the lang of the partner of the document.

To display the total untaxed amount, I use:
```
py3o://function="o_format_lang(o.amount_untaxed, currency_obj=o.currency_id)"
```
and it displays the amount in the lang of the user... we need to fix that.

lmignon:
@alexis-via Take a look at my last commit ;-) 
@GillesTephaneMeyomesse Do you call the method `o_format_lang` with named parameters? (o_format_lang(value, x=1, y=Z) ... Otherwise my last commit will break your existing reports. I would like to keep a kind of symmetry in the signature of the methods `o_format_lang` and `o_format_date`

lmignon:
@nicoclavier the issue with `_merge_pdf` is now fixed. Thank you for the report.

report_py3o/models/py3o_report.py
```diff
@@ -302,6 +305,23 @@ def _zip_results(self, reports_path):
...
         return result_path
 
+    @api.model
+    def _merge_pdf(self, reports_path):
+        """ Merge PDF files into one.
+
+        :param reports_path: list of path of pdf files
+        :returns: path of the merged pdf
+        """
+        writer = PdfFileWriter()
```
ghost, line 319: Hi. Thanks for your work!
if you did not know: there is a function in odoo core for pdf merge: https://github.com/odoo/odoo/blob/12.0/odoo/tools/pdf.py#L7
However it expects a pdf file and not a path. Feel free to include it or leave it as is. :+1:

alexis-via:
```
File "/home/alexis/new_boite/dev/reporting-engine2/report_py3o/models/_py3o_parser_context.py", line 57, in _format_lang
    env = env.with_context(lang=lang_code)
AttributeError: 'Environment' object has no attribute 'with_context'
```

report_py3o/models/py3o_report.py
lmignon: @BT-kaberer Thank you for the reference, I'm aware of this method but I don't like it since it requires to have all the documents into memory. If you take a look at the implementation of py3o_report you'll see that we try to minimize the memory footprint of the generation process by working with files and streams ... https://github.com/OCA/reporting-engine/pull/258/files#diff-718f50f1e1aa99232a24143065702955R376

lmignon:
@alexis-via I'll fix it ASAP, Sorry, not tested

lmignon:
@alexis-via Should be fixed.
We should put a call to all the methods provided into the parser context into the demo template to be sure that these are tested by travis and avoid this kind of error....

alexis-via:
It works like a charm, thanks. I use:
```
py3o://function="o_format_lang(l.product_qty, dp='Product Unit of Measure', lang_code=lang)"
```

alexis-via:
FYI, i ported purchase_report_py3o to v12, cf https://github.com/akretion/odoo-py3o-report-templates

report_py3o/models/ir_actions_report.py
```diff
@@ -70,6 +77,77 @@ def _get_py3o_filetypes(self):
...
+        This method log a warning message into the logs for each report
+        that should not work.
+        """
+        for report in self.search([("report_type", "=", "py3o")]):
+            if report.is_py3o_report_not_available:
+                logger.warning(report.msg_py3o_report_not_available)
+
+    @api.model
+    def _get_lo_bin(self):
+        lo_bin = self.env['ir.config_parameter'].get_param(
```
sbidoul: @lmignon a sudo is needed here. cc/ @sbejaoui

lmignon: Good catch @sbidoul!
fixed

report_py3o/models/py3o_report.py
```diff
@@ -0,0 +1,384 @@
...
+            suffix='.ods', prefix='p3o.report.tmp.')
+        tmpl_data = self.get_template(model_instance)
+
+        in_stream = BytesIO(tmpl_data)
+        with closing(os.fdopen(result_fd, 'wb+')) as out_stream:
+            template = Template(in_stream, out_stream, escape_false=True)
+            localcontext = self._get_parser_context(model_instance, data)
+            template.render(localcontext)
+            out_stream.seek(0)
+            tmpl_data = out_stream.read()
```
sbidoul, line 231: @lmignon this seek and read seems useless?

sbidoul:
@lmignon one more thing: the demo report is broken:
- for users with no image it produces a stack trace on `b64decode` trying to decode `False`
- for users with an image, the image does not render correctly (read error shown in the resulting file, at least with my libreoffice 6.1)

abdelhamid-attaby:
@lmignon Any updates in this PR?

lmignon:
@abdelhamid-attaby I need to find time to understand what has changed in libreoffice6 and what breaks the rendering of images. This problem seems to be a bug in the py3.template library and not related to this PR. If you do not need to insert images from Odoo data, this PR is OK.

faide:
> OK, i see that you fixed o_format_lang() in your commit a few minutes ago... I can now use it. But I still have to understand how it read the lang, because it doesn't take the decimal separator that it is supposed to use...

did you see the latest commits and especially this one : https://bitbucket.org/faide/py3o.template/commits/76c2209c8eaff0001a8af207c486d267959db354 that introduces babel as a way to really control your langs and formatting ?

We are still testing it because we have some differences from local tests and CI tests, but it should address your question...

And more important it deprecates the old date and number formatting techniques and will be the preferred (and soon only) way

abdelhamid-attaby:
@lmignon Thank you for your reply. Rendering images is very important for any templating engine. I hope you find the bugs quickly and fix it. I wish I could help but I don't have enough experience in templating engines. Sorry for that.

report_py3o_fusion_server/models/py3o_report.py
```diff
@@ -0,0 +1,110 @@
...
+            for chunk in r.iter_content(chunk_size):
+                fd.write(chunk)
+        end_chrono = datetime.now()
+        convert_seconds = (end_chrono - start_chrono).total_seconds()
+        logger.info(
+            'Report %s converted to %s in %s seconds',
+            report_xml.report_name, filetype, convert_seconds)
+        if len(model_instance) == 1:
+            self._postprocess_report(
+                result_path, model_instance.id)
```
houzefa-abba: This does not follow the signature in `report_py3o/models/py3o_report.py`:
```python
    @api.multi
    def _postprocess_report(self, model_instance, result_path):
```

Attachments don't get saved as a result. Instead the following works fine:
```suggestion
                model_instance, result_path)
```

lmignon: Thank you @houzefa-abba! I've included your changes.

lmignon:
>     * for users with an image, the image does not render correctly (read error shown in the resulting file, at least with my libreoffice 6.1)

The problem is into the py3o.template lib with libreoffice 6.1. I've reported the issue here https://bitbucket.org/faide/py3o.template/issues/25/images-are-no-longer-supported-with

sbidoul:
The image issue with libreoffice 6.1 is not blocking this PR IMO.

lmignon:
The image issue should be fixed by https://bitbucket.org/faide/py3o.template/pull-requests/49 (at least it work on my machine now...)
@faide Can you take a look at this fix PLZ?

lmignon:
The image issue seems to be a bug into libreoffice... https://bugs.documentfoundation.org/show_bug.cgi?id=123983, PR on py3o.template is a work-around.

sbidoul:
It looks like this PR is now good to merge. Thanks @lmignon for this huge effort.

PR #259, opened 2018-12-04
report_xlsx: fix report context propagation on records

### LINKED ISSUE #326: report_xlsx: Can not pass context values to xlsx reports
The context is overridden with this one with that only key, is this intended?
https://github.com/OCA/reporting-engine/blob/79166e71122b1691f67a7820bc95fa38fd2e2dcc/report_xlsx/models/ir_report.py#L19-L21
Because I need to pass some value in context and read it from report, had I miss something?
I can submit a patch in case it's a bug.

simahawk:
Without this change If you return a custom action that sets a custom context -> you lose it -> if your report depends on it... it's broken :sob: 

@adrienpeiffer @sbidoul this fixes XLS export on `mis.report.instance` :wink: 

Fixes #326

### DIFF
--- a/report_xlsx/models/ir_report.py
+++ b/report_xlsx/models/ir_report.py
@@ -16,9 +16,9 @@ class ReportAction(models.Model):
         report_model = self.env.get(report_model_name)
         if report_model is None:
             raise UserError(_('%s model was not found' % report_model_name))
-        return report_model.with_context({
-            'active_model': self.model
-        }).create_xlsx_report(docids, data)
+        return report_model.with_context(
+            active_model=self.model,
+        ).create_xlsx_report(docids, data)
 
     @api.model
     def _get_report_from_name(self, report_name):
--- a/report_xlsx/report/report_xlsx.py
+++ b/report_xlsx/report/report_xlsx.py
@@ -37,7 +37,9 @@ class ReportXlsxAbstract(models.AbstractModel):
             ids = data["context"].get('active_ids', [])
         else:
             ids = self.env.context.get('active_ids', [])
-        return self.env[self.env.context.get('active_model')].browse(ids)
+        # propagate context anyway
+        # let it fail explicitely if `active_model` is not there
+        return self.env[self.env.context['active_model']].browse(ids)
 
     def create_xlsx_report(self, docids, data):
         objs = self._get_objs_for_report(docids, data)


simahawk:
digging into this issue turns out that if you don't pass `data` [in the action you return](https://github.com/OCA/mis-builder/blob/11.0/mis_builder/models/mis_report_instance.py#L567), you won't get anything into `data` [in the controller ](https://github.com/OCA/reporting-engine/blob/11.0/report_xlsx/controllers/main.py#L12) :thinking:

sbidoul:
@simahawk in mis builder 10, we have this https://github.com/OCA/mis-builder/blob/be28ba9de4c67bbc4339522bdb1686b4e942ada3/mis_builder/models/mis_report_instance.py#L618-L627 and it works fine.

simahawk:
@sbidoul but that code has been migrated already to v11 since there's not `get_action`.
Anyway, the issue is not there. What you have in mis_builder now is:
```py
return {
            'name': 'MIS report instance XLSX report',
            'model': 'mis.report.instance',
            'type': 'ir.actions.report',
            'report_name': 'mis_builder.mis_report_instance_xlsx',
            'report_type': 'xlsx',
            'report_file': 'mis_report_instance',
            'context': context,
        }
```
and this does not work.
If I apply this patch, it does not work neither, unless I change the dict above like:
```py
[...]
            'context': context,
            'data': {'__make_report_controller_work__': 1},
```
by adding `data` key. However, its content doesn't matter...

sbidoul:
@simahawk is this still WIP?

simahawk:
@sbidoul I'm not working on this since ages :sweat_smile: 
Actually I was waiting for feedback, especially on the fact that we have to add a fake `data` dict anyhow to make it work :man_shrugging:

sbidoul:
All I can say is the fake data dict is needed to propagate the context for pdf reports too. So at least the hack is consistent ;) Why it is so is beyond me...

simahawk:
> Why it is so is beyond me...

It's because the controller relies on the automatic conversion of the params in the request and if the param is not in the request they ignore it completely :man_facepalming:

simahawk:
ok then I'll update this PR w/ a mention in the docs

sbidoul:
@simahawk I just debugged a bit, and I think the context is actually lost here:

https://github.com/OCA/reporting-engine/blob/9eb9a4248c4a08649ca6b8d6708f1fe2917f25df/report_xlsx/models/ir_report.py#L19-L21

So replacing that with `with_context(active_model=self.model)` (plus the data hack in the report action), it seems to work.

simahawk:
I removed context manipulation and fixed up the places where the context was replaced instead of updated. I cannot test it now, I'll see if someone else can.

report_xlsx/controllers/main.py
```diff
@@ -26,7 +26,7 @@ def report_routes(self, reportname, docids=None, converter=None, **data):
                 if data['context'].get('lang'):
                     del data['context']['lang']
                 context.update(data['context'])
-            xlsx = report.with_context(context).render_xlsx(
+            xlsx = report.with_context(**context).render_xlsx(
```
jffernandez: Why is this needed? It works without that change.
This context is created from self.env.context, then updated.
No need to update the actual context again with all the dict from the context var.

report_xlsx/models/ir_report.py
```diff
@@ -30,4 +30,4 @@ def _get_report_from_name(self, report_name):
         conditions = [('report_type', 'in', qwebtypes),
                       ('report_name', '=', report_name)]
         context = self.env['res.users'].context_get()
-        return report_obj.with_context(context).search(conditions, limit=1)
+        return report_obj.with_context(**context).search(conditions, limit=1)
```
jffernandez: Why is this needed? It works without that change.
You don't need to update the context for searching that object (to be consistent with other controllers, too) because you will update the context before using it.

report_xlsx_helper/controllers/main.py
```diff
@@ -33,7 +33,7 @@ def report_routes(self, reportname, docids=None, converter=None, **data):
                 context.update(data['context'])
             context['report_name'] = reportname
 
-            xlsx = report.with_context(context).render_xlsx(
+            xlsx = report.with_context(**context).render_xlsx(
```
jffernandez: Why is this needed? It works without that change.
This context is created from self.env.context, then updated.
No need to update the actual context again with all the dict from the context var.

report_xlsx/report/report_xlsx.py
```diff
@@ -37,7 +37,9 @@ def _get_objs_for_report(self, docids, data):
             ids = data["context"].get('active_ids', [])
         else:
             ids = self.env.context.get('active_ids', [])
-        return self.env[self.env.context.get('active_model')].browse(ids)
+        # propagate context anyway
+        # let it fail explicitely if `active_model` is not there
+        return self.env[self.env.context['active_model']].browse(ids)
```
jffernandez, line 42: Not really needed for the fix you're addressing. If the active model name is not there, it will fail anyway.
Keep the changes to the minimun needed.

jffernandez (changes requested):
I think that to fix the context propagation only is needed one change, with that it works.
See inline comments.
Thank you for your PR!

report_xlsx/report/report_xlsx.py
sbidoul: Here the error message will be more explicit (`KeyError`) so it's worthwhile even if not strictly necessary to fix the context propagation bug.

sbidoul (approved):
@simahawk if you'd have 2 minutes to revert the `**context` changes which are not really necessary?

For the rest I tested and the fix works fine. Thanks!

simahawk:
I changed "blindly" what have you asked for, no time to test, sorry guys :/

sbidoul:
Thanks!

/ocabot merge patch

fixes #326

PR #263, opened 2018-12-12
[IMP] do not set views ids on sql automatic win action

jjscarafia:
## Actual situation

- You create an sql query with "bi_sql_editor"
- you want to modify default views, available options are:

1. Modify created view, if you need to reset the sql query then you loose the changes
2. Modify views by extending. Then you can not reset sql quer because of constraint on ir.ui.view with parent view
3. You can create new views with priority lower than 16. Great option! but....

## Problem found
The window action created by bi_sql_editor explicitly refers to view_id and view_search_id while those two fields are optional and really no needed. Better to let odoo choose the right view by priority

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -437,20 +437,11 @@ class BiSQLView(models.Model):
     def _prepare_action(self):
         self.ensure_one()
         view_mode = self.view_order
-        first_view = view_mode.split(',')[0]
-        if first_view == 'tree':
-            view_id = self.tree_view_id.id
-        elif first_view == 'pivot':
-            view_id = self.pivot_view_id.id
-        else:
-            view_id = self.graph_view_id.id
         return {
             'name': self._prepare_action_name(),
             'res_model': self.model_id.model,
             'type': 'ir.actions.act_window',
             'view_mode': view_mode,
-            'view_id': view_id,
-            'search_view_id': self.search_view_id.id,
         }
 
     @api.multi


ValentinoDefelice (approved):

zaoral (approved):
+1

pedrobaeza:
@aheficent any of you that are users of this module can confirm this patch?

AaronHForgeFlow:
IMO view_order field allows the user to select what view comes first, if we remove this then we are losing a feature. I am not sure. Ping @legalsylvain

legalsylvain (changes requested):
I'm really not in favour of this patch.
indeed, as @aheficent said, this allow user to select the view that will be used by default. So we are loosing a feature.

I didn't faced the case where we have to make manual changes of the generated view. could you detail your use case ? just to understand ? 

But mainly, I don't understand what your patch is solving. (but maybe I missed something). because when you reset the sql (to draft) the function button_set_draft will unlink all the model and the fields and all the UI (menu, views, action). So changes will be dropped anyway.

Thanks for your explanations.

jjscarafia:
Hi @aheficent @legalsylvain !
Thanks for the review.
Please check [this video](https://drive.google.com/file/d/1qko-nk5PB3-LcGgR-3mio1sbVAt1xXXF/view) with a real use case and explanation on why we are not loosing anything and why I believe we are winning.

Sorry for the ugly video :)

jjscarafia:
@aheficent @legalsylvain . Can you take a look to the explanation on previous comment?
Thanks!

legalsylvain:
Hi. I'm on hollidays until 8, January. i'll take a look the next week. please ping me if I forget !
kind regards, and happy new years !

jjscarafia:
Thanks @legalsylvain ! pinging right now ;)

legalsylvain (approved):
Well, in fact, ``view_mode`` that define the order is not removed by this patch.

So, :+1: with that patch. (my bad for my first bad analysis)

And happy new year !

PR #264, opened 2018-12-14
[12.0][MIG] bi_sql_editor

AdriaGForgeFlow:
Migration of module bi_sql_editor to v12

PR #269, opened 2019-01-15
[FIX] bi_sql_editor: solves the update all break

MiquelRForgeFlow:
This PR updates the readme of bi_sql_editor to inform how to solve the update all break.

### DIFF
--- a/bi_sql_editor/README.rst
+++ b/bi_sql_editor/README.rst
@@ -47,6 +47,12 @@ this module is interesting for the following use cases
 * You have a lot of data, and classical SQL Views have very bad performance.
   In that case, MATERIALIZED VIEW will be a good solution to reduce display duration
 
+Installation
+============
+
+* You must put this module as `server_wide_modules` in your odoo configuration file
+  or add '--load=bi_sql_editor' if you start odoo in command line.
+
 Configuration
 =============
 


HviorForgeFlow (approved):
LGTM :+1:

JordiBForgeFlow (approved):
Tested in a broken environment, and fixes the issue. Works like a charm!

MiquelRForgeFlow:
NOTE: For reference, added here the error that this PR solves.

> Selection_632
> in https://github.com/OCA/reporting-engine/pull/264#issuecomment-470890648

legalsylvain:
Hum. Thanks for the comment. 

My question is : I use this module since years (version 8 and 12), and I realize update all each month. I never faced that issue. 
Do you know in which context ?

MiquelRForgeFlow:
I don't remember. Perhaps it has to do if you have bi sql views for materialized views, not sure. @HviorForgeFlow do you remember?

legalsylvain:
(I only use materialized views).

legalsylvain:
Well, I removed that comment in #670 as it seems to work. If there is a problem, we could always add it again, with a more detailled text of what the --load solves.

thanks for your answers.

PR #271, opened 2019-01-20
[FIX] add slash in url in qwebactionmanager.js

ruben-tonetto:
For me, it does work with added slash
if not i receive an error indicating that the url ...:8069report/... is not found


pedrobaeza:
This is working as expected in other addons. Can you specify how you reproduce the error?

ruben-tonetto:
Sorry, you are right!
The problem was in another module
I should have tested before with empty database.

ruben-tonetto:
ps: i've tried the module report_xlsx in runbot and it gives an error when trying to export two records; this is the reason that made me sure of the problem

pedrobaeza:
The module is not usable without another module that makes use of their possibilities, so I'm curious how do you reproduce the problem.

PR #272, opened 2019-02-05
[ADD] report csv

etobella:
This module allows to create csv reports directly on odoo on a similar way than `report_xlsx`


pedrobaeza:
@alexis-via didn't you develop something similar?

alexis-via:
@pedrobaeza Maybe report_qweb_txt https://github.com/OCA/reporting-engine/tree/10.0/report_qweb_txt

pedrobaeza:
Yeah, that's the one I remember. It's not the same then.

JordiBForgeFlow (approved):
Code review and functional tests

LoisRForgeFlow (approved):
Great module! :+1:

jarroyomorales (approved):
LGTM

pedrobaeza:
The problem I see with this module (and the `report_xlsx` one) is the need of adding code for using it, instead of using data for generating reports. On `report_xml`, a data file serves as "template" for generating the XML. I know it's not the same target, but maybe there's some way of doing this in a similar manner.

JordiBForgeFlow:
@pedrobaeza I think that it's convinient for some use cases. And for some others it won't. I see that the module is formally correct, and solves a use case.

etobella:
If we want to make the CSV from the XML we should create a parser and define the logic somewhere. It shouldn't be hard, but I think it is out of the scope of this module. Maybe, I could create an upgrade after this module. What do you think @pedrobaeza ?

pedrobaeza:
Yes, I'm not blocking this. Simply thinking out loud for a better design.

PR #274, opened 2019-02-18
[ADD] reporting-context

jarroyomorales:
This module adds a context variable to reports. A possible use for this
context could be hiding some fields or many other configuration options.

To configure this module, you need to:

* Enter Odoo in debug mode.
* Go to Settings -> Parameters -> System Parameters. On the system
  parameters list, look for report.default.context, which is a Python
  dictionary variable where you can add a context that will be common for
  all reports.
* To add a specific context to a report, you should go to Settings ->
  Reporting -> Reporting and look for the report you want to edit on the
  list. You will see that now they contain a new field called Context Value
  , where you will be able to add all the desired context parameters.


jarroyomorales:
This is the addon that we were talking about on OCA/l10n-spain#1014

report_context/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
+
+from . import test_report_context
```
etobella: Missing new line

report_context/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
```
etobella: It is not necessary if it is in the __init__

etobella (approved):
Just a few comments. But it looks great for me

JordiBForgeFlow (approved):
:+1: Tested and reviewed code. Works as expected. Notice that in order to work effectively in some cases it need this fix in Odoo https://github.com/odoo/odoo/pull/31150

JordiBForgeFlow:
@jarroyomorales can you squash your commits?

pedrobaeza:
Why not adding the context in the report object in a similar way that action windows have a context?

etobella:
Well, it can be edited like 
```xml
<record id="model_name.report_id" model="ir.actions.report">
    <field name="context">YOUR CONTEXT HERE</field>
</record>
```
It is hard to edit the standard function where it reads it, as it is on `odoo.tools`.

pedrobaeza:
Yeah, that's the idea. It's for having the information in the same place, instead of using some kind of black magic in names for fetching from system parameters the context.

etobella:
This module allows to use both systems. A generic parameter and specific report parameters. Maybe it should be explained on the readme.

pedrobaeza:
I only saw the part of the system parameter, but for me the most important one should be the other. Even more, I would remove the system parameter feature, as it's redundant, can produce permission problems, and you already have the other mechanism.

etobella:
Well, the system parameter might be important as sometimes you have a standard behaviour that should be found on all reports. But it is not the interesting part of the module.

pedrobaeza:
OK, I see. Maybe giving more importance to the other part in the README avoids confusions like mine.

pedrobaeza:
But I admit the problem was to not reading full README. If you don't want to spend more time, I can merge it.

etobella:
@jarroyomorales can you improve the README and rebase all in order to merge it, please?

jarroyomorales:
Done, I changed a little bit the readme and squashed the commits!

JordiBForgeFlow:
Yes, it is better understood now!

PR #275, opened 2019-02-18
[FIX] Fix issue on returning a bi_sql_view to draft and add some tests

daramousk:


daramousk:
@hbrunn This will not turn green. Let me know what you think

hbrunn:
the [linting](https://travis-ci.org/OCA/reporting-engine/jobs/502085536#L1097) issues are an easy win (fix the ones for report_signer in a separate commit), and as you don't test anything after your `view.unlink()`, you can just as well not do it at all - a test is only a proper test if you assert something sensible about the result of the code you run.

For fixing the actual problem, you'll have to dive deep into Odoo's [caching/update code](https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L601), from the stack trace it looks like there are triggers generated that aren't cleaned up, you'd have to do this manually.

daramousk:
While the issues you mentioned have been fixed the linting issues mentioned by the runner have nothing to do with this module yet they turn the build red.

hbrunn:
that's how it works, a build is only green if the whole repo is fine. That's why I proposed to fix them in a separate commit, that's nice for the community and increases the chances someone actually reviews and merges your PR tremendously

.travis.yml
```diff
@@ -13,6 +13,7 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
+  postgresql: "9.3"
```
MiquelRForgeFlow:
```suggestion
  postgresql: "9.6"
```

PR #283, opened 2019-04-10
[WIP] Replace resolution parameter with size

daramousk:
cc: @hbrunn

### DIFF
--- a/report_qweb_pdf_watermark/models/report.py
+++ b/report_qweb_pdf_watermark/models/report.py
@@ -17,6 +17,10 @@ try:
     from PIL import PdfImagePlugin  # noqa: F401
 except ImportError:
     pass
+try:
+    from papersize import parse_couple, SIZES
+except ImportError:
+    raise
 from openerp import api, models, tools
 logger = getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


hbrunn:
please explain which problem this is supposed to fix, and check travis.

Note that the watermark is not supposed to be scaled, if this is what you need, then you should scale your watermark to the size you need beforehand.

daramousk:
I think that this module needs some further refactoring:

1) It accepts any file while it can only process pdfs and images.
2) If the watermark is not supposed to be scaled then the user should not be allowed to set a watermark that has different dimensions than the page, because without this PR I have a pdf which contains a part of the watermark as it's background. And what I did here is actually scaling the image in order to achieve the a proper result. Or at least give a boolean option of whether to scale the image or not.

hbrunn:
1. sounds good, now people actually have to [read the documentation](https://github.com/OCA/reporting-engine/blob/f9a45afa12bae57b0191b53d98f491df17bb8438/report_qweb_pdf_watermark/README.rst#usage) to find that out. Implement by moving the watermark loading code to its own function and calling this in your constraint, not by sniffing the mimetype or similar.
2. adding a checkbox for scaling with default `False` seems fine to me, this way you also don't destroy existing functionality. Note that this will only work for non-pdf watermarks.

None of the above is a [refactoring](https://en.wikipedia.org/wiki/Code_refactoring) in the literal sense btw.

gfcapalbo:
@daramousk  some of your image degradation problems will be addressed if you extend this improvement with a proportion check ORIGINAL vs Resized. to ensure you are actually scaling the image and not deforming. A simple extension of what done.  do a check on the actual image proportions (a boolean to switch this on/off is good.)
if scale is true, one of the resize params is calculated to keep ratio H/W intact.

daramousk:
pushed my latest work on this, tests to come a little bit later

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
-            context=context
-        )
-        report = self._get_report_from_name(cr, uid, report_name)
-        watermark = None
+    @api.multi
+    def _scale_watermark(self, report, image):
+        size = parse_couple(
+            SIZES[report.paperformat_id.format.lower()])
+        size = int(size[0]), int(size[1])
+        return image.resize(size, resample=Image.LANCZOS)
```
hbrunn, line 46: you actually stretch the watermark, which distorts proportions if the watermark doesn't happen to have the same proportions as the report. either make this configurable, or decide for using either the x or y side for scaling

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
-        result = super(Report, self).get_pdf(
-            cr, uid, ids, report_name, html=html, data=data,
-            context=context
-        )
-        report = self._get_report_from_name(cr, uid, report_name)
-        watermark = None
+    @api.multi
+    def _scale_watermark(self, report, image):
+        size = parse_couple(
+            SIZES[report.paperformat_id.format.lower()])
```
hbrunn: as you pointed out in a comment before, this could fail. So handle this by logging an error and not scaling

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
+        size = int(size[0]), int(size[1])
+        return image.resize(size, resample=Image.LANCZOS)
+
+    @api.multi
+    def _read_watermark(self, report):
         if report.pdf_watermark:
             watermark = b64decode(report.pdf_watermark)
         else:
```
hbrunn, line 52: this branch shouldn't be in this function, as it will most likely fail when called without ids. this function should only be concerned with creating the watermark from the binary given.

hbrunn, line 52: this branch shouldn't be in this function, as it will most likely fail when called without ids. this function should only be concerned with creating the watermark from the binary given.

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
-                'No usable watermark found, got %s...', watermark[:100]
-            )
-            return result
-
-        if pdf_watermark.numPages < 1:
-            logger.error('Your watermark pdf does not contain any pages')
-            return result
-        if pdf_watermark.numPages > 1:
-            logger.debug('Your watermark pdf contains more than one page, '
-                         'all but the first one will be ignored')
```
hbrunn, line 85: those handlings should be in your constraint then

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
+            image = Image.open(StringIO(watermark))
+            pdf_buffer = StringIO()
+            if image.mode != 'RGB':
+                image = image.convert('RGB')
+            if report.paperformat_id.format and \
+                    report.paperformat_id.format.lower() in SIZES and \
+                    report.pdf_watermark_scale:
+                image = self._scale_watermark(report, image)
+            image.save(pdf_buffer, 'pdf', quality=100)
```
hbrunn, line 75: check out what the `resolution` parameter does for pdfs and reconsider

report_qweb_pdf_watermark/models/report.py
```diff
@@ -17,73 +17,66 @@
...
+            pdf_watermark = PdfFileReader(pdf_buffer)
+        return pdf_watermark
 
+    @api.model
```
hbrunn, line 79: read this in the `report` module, and reconsider

hbrunn (changes requested):

report_qweb_pdf_watermark/models/report.py
daramousk: What is wrong with this? The function seems to be called properly, and the `api.model` is the almost the functional equivalent of `api.v7` and `api.v8`.

report_qweb_pdf_watermark/models/report.py
daramousk: I am not sure what you want me to do with `resolution`.

Also, as I understand with my limited knowledge on the picture data structures and their operations, changing the resolution should not change the size (width, length) of an image/pdf. Yet this is what happens when I set `resolution=report.paperformat_id.dpi`. 

The watermark as applied right know is great, but my issue here is that while the image I provide has _higher_ resolution and I lower its resolution it appears "foggy". But I assume since we do not have to do with SVG files this is normal.

report_qweb_pdf_watermark/models/report.py
daramousk: > you actually stretch the watermark, which distorts proportions if the watermark doesn't happen to have the same proportions as the report.

We have to be specific with terminology so that we can understand each other. The operation I apply here is `scaling` or `resizing`. Because the image file is PNG which is a raster graphics file format, when **scaling** there will be a loss of quality because the resolution stays the same.

By giving the option to `scale` here, I mean to bring the watermark to the same size (**not** resolution) as the already existing report.

I am already giving the opportunity to the user not to scale at https://github.com/OCA/reporting-engine/pull/283/files#diff-0f06fef7d1f6955c05bfff02a178967cR11. Should they choose this they should know the consequences (or they will quickly be able to see them).

Having said that, what troubles me is that while I understand that when you scale **up** an image there will be a loss of quality, when you scale **down** it should not, which is what happens here.

report_qweb_pdf_watermark/models/report.py
hbrunn: we don't need an almost functional equivalent, but the correct override, otherwise the function doesn't do the right thing wenn called from v7 or the other way around

report_qweb_pdf_watermark/models/report.py
hbrunn: 400x400 -> 800x800: scaling
400x400 -> 600x800: stretching, distorting proportions and that's what you do here

report_qweb_pdf_watermark/models/report.py
daramousk: See this https://github.com/OCA/reporting-engine/pull/283/files#diff-ea48341e81bea8a710c5781b95d7ff8dR83. I call the function both ways and it works correctly. Also, I am not sure with the linter fails and the tests do not run but this module does not seem to be the issue.

PR #285, opened 2019-05-01
[MIG][11.0] report_py3o backport of v12

tbaden:


lmignon:
@tbaden Thank you for taking care of this backport.  Can I ask you to preserve the commit history and authorship of changes when backporting the addons from Odoo 12. 
Here it's a technical method to backport the addons from 12.0...
* On a shell command: 
```
$ git clone https://github.com/OCA/reporting-engine -b 11.0
$ git checkout -b 11.0-mig-reporting-engine origin/11.0
$ git format-patch --keep-subject --stdout origin/11.0..origin/12.0 -- report_py3o report_py3o_fusion_server | git am -3 --keep
```
* Adapt the module to the 11.0 version.
* On a shell command: 
```
$ git add --all
$ git commit -m "[MIG] $MODULE: Migration to 11.0"
$ git remote add $USER_ORG git@github.com:$USER_ORG/$REPO.git # This mode requires an SSH key in the GitHub account
$ ... or ....
$ git remote add $USER_ORG https://github.com/$USER_ORG/$REPO.git # This will required to enter user/password each time
$ git push $USER_ORG 11.0-mig-$MODULE --set-upstream
```

(same method as for normal migration process but adapted to takes the 12.0 branch as source. This method also takes care to migrate report_py3o **and**  report_py3o_fusion_server at same time. https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#technical-method-to-migrate-a-module-from-110-to-120-branch)

tbaden:
@lmignon I'll take care of this.

lmignon:
> @lmignon I'll take care of this.

Thank you to you for this work :smirk:

tbaden:
@lmignon maybe you can give me a hint. I got all tests on green now except for test_reports_merge_zip.

getting the error: 
<string>:1: in <Expression '__py3o_frame(b64decode(objects.image), \'png\', width=\'2cm\', height=\'2cm\', origin_attrib={\'{urn:oasis:names:tc:opendocument:xmlns:text:1.0}anchor-type\': \'paragraph\', \'{urn:oasis:names:tc:opendocument:xmlns:svg-compatible:1.0}width\': \'0.7902in\', \'{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}name\': "py3o.image(b64decode(objects.image), \'png\', width=\'2cm\', height=\'2cm\')", \'{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}style-name\': \'fr1\', \'{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}z-index\': \'0\'})'>
    ???
/usr/lib/python3.5/base64.py:81: in b64decode
    s = _bytes_from_decode_data(s)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 TypeError: argument should be a bytes-like object or ASCII string, not 'bool'

If you need it, I can push my current code to github.

report_xlsx/static/src/js/report/qwebactionmanager.js
```diff
@@ -3,10 +3,14 @@
 odoo.define('report_xlsx.report', function(require){
 'use strict';
 
-var ActionManager= require('web.ActionManager');
+var ActionManager = require('web.ActionManager');
+var core = require('web.core');
```
lmignon: @tbaden `core, _t, _lt` are never used. This file should stay untouched.

report_py3o_fusion_server/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,7 @@
+* Thore Baden <thorebaden@gmail.com>
```
lmignon: @tbaden Can you put your name at the end of the list. This list is chronologically ordered.

report_py3o/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,7 @@
+* Thore Baden <thorebaden@gmail.com>
```
lmignon: @tbaden Can you put your name at the end of the list. This list is chronologically ordered.

report_py3o/models/ir_actions_report.py
```diff
@@ -194,3 +194,28 @@ def _get_attachments(self, res_ids):
...
                         save_in_attachment[record_id.id] = attachment_id
         return save_in_attachment
+
+    @api.model
+    def _get_rendering_context_model(self):
+        report_model_name = 'report.%s' % self.report_name
+        return self.env.get(report_model_name)
+
+    @api.model
+    def _get_rendering_context(self, docids, data):
```
lmignon, line 204: @tbaden Why do we need these methods? These are never called by the current addon. (and seems not to be part of the odoo api since they don't call `super`)

report_py3o/static/src/js/py3oactionmanager.js
```diff
@@ -1,38 +1,75 @@
...
+    var make_report_url = function (action) {
+        var report_urls = {
+            'py3o': '/report/py3o/' + action.report_name,
+        };
         // We may have to build a query string with `action.data`. It's the place
         // were report's using a wizard to customize the output traditionally put
         // their options.
-        if (_.isUndefined(action.data) || _.isNull(action.data) ||
-            (_.isObject(action.data) && _.isEmpty(action.data))) {
+        if (_.isUndefined(action.data) || _.isNull(action.data) || (_.isObject(action.data) && _.isEmpty(action.data))) {
```
lmignon: @tbaden PLZ avoid to reformat the code.  Code reformatting adds burden to the review process without added value. Moreover, if the code looks better to you when formatted in this way, for me that reduce de readability...

report_py3o/static/src/js/py3oactionmanager.js
```diff
@@ -1,38 +1,75 @@
...
             if (action.context.active_ids) {
-                var activeIDsPath = '/' + action.context.active_ids.join(',');
-                reportUrls.py3o += activeIDsPath;;
+                var active_ids_path = '/' + action.context.active_ids.join(',');
```
lmignon: @tbaden PLZ avoid cosmetic changes... it becomes hard to compare between versions if the variable are named in a different way.

report_py3o/static/src/js/py3oactionmanager.js
```diff
@@ -1,38 +1,75 @@
...
-    _executeReportAction: function (action, options) {
-        // Py3o reports
-        if ('report_type' in action && action.report_type === 'py3o' ) {
-            return this._triggerDownload(action, options, 'py3o');
-        } else {
-            return this._super.apply(this, arguments);
-        }
-    },
+    var ActionManager = require('web.ActionManager');
+    var core = require('web.core');
```
lmignon: @tbaden `core` is not used... to remove

lmignon (changes requested):

lmignon:
> @lmignon maybe you can give me a hint. I got all tests on green now except for test_reports_merge_zip.
> 
> getting the error:
> :1: in <Expression '__py3o_frame(b64decode(objects.image), 'png', width='2cm', height='2cm', origin_attrib={'{urn:oasis:names:tc:opendocument:xmlns:text:1.0}anchor-type': 'paragraph', '{urn:oasis:names:tc:opendocument:xmlns:svg-compatible:1.0}width': '0.7902in', '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}name': "py3o.image(b64decode(objects.image), 'png', width='2cm', height='2cm')", '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}style-name': 'fr1', '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}z-index': '0'})'>
> ???
> /usr/lib/python3.5/base64.py:81: in b64decode
> s = _bytes_from_decode_data(s)
> 
> TypeError: argument should be a bytes-like object or ASCII string, not 'bool'
> 
> If you need it, I can push my current code to github.

`objects.image` is False... That means that field image is False on the object used into the report...
In fact into the demo template, we don't check if the image field is set before the call to b64decode... We should fix the template to avoid this kind of error.

report_py3o/static/src/js/py3oactionmanager.js
tbaden: copied from v11 core, will remove

report_py3o/static/src/js/py3oactionmanager.js
tbaden: copied from v11 core, will refractor

report_py3o/static/src/js/py3oactionmanager.js
tbaden: copied from v11 core, will refractor

report_py3o/models/ir_actions_report.py
tbaden: this is called here:

https://github.com/OCA/reporting-engine/blob/4ce795a6d0ef9ea40c0fd00a6aa83d592e362e7b/report_py3o/models/py3o_report.py#L197-L199

in v12 these are functions in odoo core. v11 don't have these. in v10 this module had a custom function using rml_parse.

report_py3o/readme/CONTRIBUTORS.rst
tbaden: will change it

report_py3o_fusion_server/readme/CONTRIBUTORS.rst
tbaden: will change it

report_xlsx/static/src/js/report/qwebactionmanager.js
tbaden: copied from v11 core, will remove

tbaden (commented):
When I checked the object neither of the entries had images and only the last one failed. I'll check again.

report_py3o/static/src/js/py3oactionmanager.js
tbaden: copied from v11 core, will refractor

tbaden: or is it more consistent to keep acrive_ids_path as it's used everywhere in odoo11

tbaden:
@lmignon @pedrobaeza 
how should I handle this test?
```
    def test_py3o_report_availability(self):
        # This test could fails if libreoffice is not available on the server
```

locally it pass, because of libreoffice installation, but travis will fail.

tbaden:
> > @lmignon maybe you can give me a hint. I got all tests on green now except for test_reports_merge_zip.
> > getting the error:
> > :1: in <Expression '__py3o_frame(b64decode(objects.image), 'png', width='2cm', height='2cm', origin_attrib={'{urn:oasis:names:tc:opendocument:xmlns:text:1.0}anchor-type': 'paragraph', '{urn:oasis:names:tc:opendocument:xmlns:svg-compatible:1.0}width': '0.7902in', '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}name': "py3o.image(b64decode(objects.image), 'png', width='2cm', height='2cm')", '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}style-name': 'fr1', '{urn:oasis:names:tc:opendocument:xmlns:drawing:1.0}z-index': '0'})'>
> > ???
> > /usr/lib/python3.5/base64.py:81: in b64decode
> > s = _bytes_from_decode_data(s)
> > TypeError: argument should be a bytes-like object or ASCII string, not 'bool'
> > If you need it, I can push my current code to github.
> 
> `objects.image` is False... That means that field image is False on the object used into the report...
> In fact into the demo template, we don't check if the image field is set before the call to b64decode... We should fix the template to avoid this kind of error.

seems like I dived to deep... found it and catching this issue in the context parser.

on local all tests are green

lmignon:
> @lmignon @pedrobaeza
> how should I handle this test?
> 
> ```
>     def test_py3o_report_availability(self):
>         # This test could fails if libreoffice is not available on the server
> ```
> 
> locally it pass, because of libreoffice installation, but travis will fail.

@tbaden add this commit by cherry-pick :smirk: https://github.com/OCA/reporting-engine/commit/8d3b61a6e10e242a3bdd54f6b6983f5482e94554

report_py3o/models/py3o_report.py
```diff
@@ -198,6 +198,9 @@ def _get_parser_context(self, model_instance, data):
             report_xml._get_rendering_context(model_instance.ids, data)
         )
         context['objects'] = model_instance
+        # image false will crash base64 decoding
```
lmignon: @tbaden The image field does not exist on all models. It is in the template that you must add an `if' instruction before using this field (it is a demo template)

report_py3o/static/src/js/py3oactionmanager.js
lmignon: It's to avoid difficulties if we need to backport changes/fixes from V12 to V11.

lmignon (changes requested):

report_py3o/models/py3o_report.py
tbaden: how would this instruction look like in the odt template?
additional: wouldn't it be good to still check if image exists in objects and when false to set it to empty bytestring. otherwise you'll get an error if an user forgets this if instruction.

like this:
```
        if (
            hasattr(context['objects'], 'image') and
            not context['objects'].image
        ):
            context['objects'].image = b''
```

lmignon: @tbaden A report can be defined for any model and the name of a field used to store an image is not always 'image'. 
It's important to manage this case into the template to also avoid to display an empty square if the image is missing into the document.
You can use a control structure into the template to avoid to output an image if not set https://py3otemplate.readthedocs.io/en/latest/templating.html#use-control-structures

tbaden: @lmignon  done

lmignon: @tbaden The demo template seems to no more work....

tbaden: how did you test this or whats the error?

lmignon: I functionally tests your changes thanks to ci/runbot... 
There is no more image into the generated report

tbaden: Ah I see. Do you have a working example for an if condition in a image frame option? I checked the reference you send, but the are only "normal" if conditions and not a single one for options. tried brackets for a single line if, but didn't work.

lmignon: @tbaden This template provides you some example on how to design a template .. https://bitbucket.org/faide/py3o.template/src/default/example/py3o_example_template.odt

tbaden: I think I finally figured it out. getting the right files with just py3o installed. with fusion_server an error. will check this

tbaden: @lmignon is it really needed to dump the dict as there is no encoding involved?
https://github.com/OCA/reporting-engine/blob/47dc23626025b991388434e3bdb8add82854bcde/report_py3o_fusion_server/models/py3o_report.py#L72

json.dumps fails on bytestrings (images)

lmignon: IMO the problem is not there. This code works into production (V12) with complex reports involving images etc...

tbaden: I just cloned the v12 version of this module into my v12 odoo and get the same error when running tests:

> E       TypeError: b'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVQI12P4//8/AAX+Av7czFnnAAAAAElFTkSuQmCC' is not JSON serializable

If it's working on your odoo, maybe a missing package or something similar?

lmignon: strange, even tests on travis for V12 are OK. Should be investigated but unfortunately I've no time for it for the moment.

tbaden: @lmignon I'll investigate. Thanks a lot for the support so far.

tbaden: @lmignon the use of simplejson could fix this. what's your opinion on this?

like: 

```
 try:
         import simplejson as json
except ImportError:
         import json
```

lmignon: @tbaden Before introducing a new dependency I prefer to understand what's the root cause of the problem...

tbaden: I'll reset my dev env and install it again. Maybe I messed it up somewhere for other developments.

lmignon: I think we should try to stay with the standard python libraries. It doesn't seem normal to me that the result of serialization to json works with one library and not another when it shouldn't make any difference. I have the feeling that the problem encountered is elsewhere.

tbaden: @lmignon I found the "issue". In v11 the portal User has no image, in v12 the portal user has an image. If you delete the image in v12 you get the exact same error:

>  File "/usr/lib/python3.5/base64.py", line 46, in _bytes_from_decode_data
>     "string, not %r" % s.__class__.__name__) from None
> TypeError: argument should be a bytes-like object or ASCII string, not 'bool'

If I use my if structure template I get the same error:

> E TypeError: b'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVQI12P4//8/AAX+Av7czFnnAAAAAElFTkSuQmCC' is not JSON serializable

So it's a general issue in v11 and v12. As you said, this should be caught in the odt template with an if structure. Maybe my if structure is still faulty. Could you please check my odt [template](https://github.com/OCA/reporting-engine/raw/4bf56dfa1be1b861f4251f89cd8074614c0c7326/report_py3o/demo/res_user.odt)

cataclysmic:
What is the current situation with the backport? I'm using odoo 11 CE and would really like to try this module. I'd be willing to help. However, I'm a noob in odoo development.

cataclysmic:
When I try to create a report from the admin interface I get the following error:

`Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 949, in __get__
    value = record.env.cache.get(record, self)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 977, in get
    value = self._data[key][field][record._ids[0]]
KeyError: <odoo.models.NewId object at 0x7f0a7c5e8768>

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 949, in __get__
    value = record.env.cache.get(record, self)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 977, in get
    value = self._data[key][field][record._ids[0]]
KeyError: <odoo.models.NewId object at 0x7f0a7c5e8768>

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/odoo/http.py", line 653, in _handle_exception
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 312, in _handle_exception
  File "/usr/lib/python3/dist-packages/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 695, in dispatch
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 344, in _call_function
  File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 97, in wrapper
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 337, in checked_call
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 939, in __call__
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 517, in response_wrap
    response = f(*args, **kw)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 934, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 926, in _call_kw
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 699, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 690, in call_kw_multi
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 5245, in onchange
    for name in nametree
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 5246, in <listcomp>
    if name not in done and snapshot0.has_changed(name)
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 5148, in has_changed
    return self[name] != record[name]
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 4820, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 955, in __get__
    self.determine_draft_value(record)
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1077, in determine_draft_value
    self._compute_value(record)
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1011, in _compute_value
    getattr(records, self.compute)()
  File "/usr/lib/python3/dist-packages/odoo/addons/report_py3o_fusion_server/models/ir_actions_report.py", line 46, in _compute_py3o_report_not_available
    if (not rec.is_py3o_native_format and
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 955, in __get__
    self.determine_draft_value(record)
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1077, in determine_draft_value
    self._compute_value(record)
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1011, in _compute_value
    getattr(records, self.compute)()
  File "/usr/lib/python3/dist-packages/odoo/addons/report_py3o/models/ir_actions_report.py", line 126, in _compute_is_py3o_native_format
    rec.is_py3o_native_format = format.get_format(filetype).native
  File "/usr/local/lib/python3.5/dist-packages/py3o/formats/main.py", line 67, in get_format
    raise UnkownFormatException("Format {} is unkown".format(name))
py3o.formats.main.UnkownFormatException: Format False is unkown`

PR #287, opened 2019-05-08
[FIX] UI Not being blocked when exporting xlsx

Maartincm:
* `framework.unblockUI` should be called after `get_file` has finished
(now its unblocking the UI just after it is blocked)
* The Response from the server should set a *fileToken* cookie as
requested in the docstring of `get_file` in order to trigger the
callbacks

`get_file` func can be found in /web/static/src/js/core/ajax.js :: 262


dmsergio (approved):

dmsergio:
This change works fine. Its ok for me, and very useful for user.

PR #296, opened 2019-06-11
[11.0] [FIX] report_xlsx: context is overwritten when creating report recordset

dalonsod:
This error implies losing user language in report context, so traslatable fields are shown in base language instead of user one.

The bug was already fixed in v12 during addon migration, I simply made the same changes on `with_context()` call during XLSX report rendering.


pedrobaeza:
Isn't this the same as #259?

dalonsod:
No doubt, it's clear I didn't properly search across open PRs :-(

Sorry. I'll close this PR and start working on the right PR.

PR #306, opened 2019-07-03
Update README for report_py3o and report_py3o_fusion_server

alexis-via:
Convert README to the new readme subdirectory


sbidoul:
@alexis-via the README is broken. Can you try generating it manually to fix the error?

Normally such errors are caught when merging with /ocabot.

PR #307, opened 2019-07-19
[12.0][ADD] - New module: report substitute

sbejaoui:
This module allows you to create substitution rules for report actions. A typical use case is to replace a standard report by alternative reports when some conditions are met. For instance, it allows to configure alternate reports for different companies.


sbidoul:
Suggested DESCRIPTION.rst:

```This module allows you to create substitution rules for report actions. A typical use case is to replace a standard report by alternative reports when some conditions are met. For instance, it allows to configure alternate reports for different companies.```

File: report_substitute/models/ir_actions_report.py [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+                )
+        return False
+
+    @api.multi
+    def render(self, res_ids, data=None):
+        substitution_report = self._get_substitution_report(
+            self.model, res_ids
+        )
+        if substitution_report:
+            return substitution_report.render(res_ids)
```
etobella: Just wondering one thing, with this change, you may be substituting recursively... Maybe you should make:
`return super(IrActionReport, substitution_report).render(res_ids)`

etobella (changes requested):
I think that you have a recursive issue.

However, it is really interesting!!!

report_substitute/models/ir_actions_report.py
sbejaoui: yes it is a desired behavior. 
we return super only when there is no possible substitution. otherwise we continue the substitution recursively.

```python
if substitution_report:
    return substitution_report.render(res_ids)
return super().render(res_ids, data)
```

sbidoul: @etobella it's true that a user could create loops, and it might be nice to add a better error than a stack overflow if it happens. But that should not be an issue in practice, what do you think?

sbidoul:
Last thing we'll add is an option to let the user decide what to do if there is no matching substitution: either raise an error, or fallback on the base report.

report_substitute/models/ir_actions_report.py
```diff
@@ -0,0 +1,74 @@
...
+        substitution_report = action_report
+        while substitution_report:
+            action_report = substitution_report
+            substitution_report = action_report._get_substitution_report(
+                action_report.model, active_ids
+            )
+        return action_report
+
+    @api.model
+    def get_substitution_report_dict(self, action_report_dict, active_ids):
```
sbidoul:
```suggestion
    def get_substitution_report_action(self, action):
```

Since `active_ids` is in the `action` dict anyway.

A short test for this method would be nice.

report_substitute/static/src/js/action_manager.js
```diff
@@ -0,0 +1,37 @@
...
+    ActionManager.include({
+
+        /**
+         * Intercept action handling substitute the report action
+         * @override
+         */
+
+        _handleAction: function (action, options) {
+            if (action.type === "ir.actions.report" &&
+                action.context.active_ids) {
```
sbidoul: Is there some indication in action that there are subsitution rules for this report? To save a server round-trip is there are no substitution rules.

report_substitute/static/src/js/action_manager.js
```diff
@@ -0,0 +1,37 @@
...
+                var active_ids = action.context.active_ids;
+                var self = this;
+                var _super = this._super;
+                var callersArguments = arguments;
+                return this._rpc({
+                    model: "ir.actions.report",
+                    method: "get_substitution_report_dict",
+                    args: [action, active_ids]
+                }).then(function (action_id) {
```
sbidoul: is it an action_id or an action dict?

report_substitute/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,28 @@
...
+            and self.env.context.get('active_ids')
+        ):
+            active_ids = self.env.context.get('active_ids')
+            old_report_template = self.template_id.report_template
+            self.template_id.report_template = (
+                old_report_template.get_substitution_report(active_ids)
+            )
+        res = super().onchange_template_id_wrapper()
+        if old_report_template:
+            self.template_id.report_template = old_report_template
```
sbidoul:
```suggestion
            self.template_id.report_template = old_report_template
```

I would put the 3 lines above inside the if, for readability.

report_substitute/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,28 @@
...
+    _inherit = 'mail.compose.message'
+
+    @api.multi
+    @api.onchange('template_id')
+    def onchange_template_id_wrapper(self):
+        old_report_template = False
+        if (
+            self.template_id
+            and self.template_id.report_template
+            and self.env.context.get('active_ids')
```
sbidoul: ... and `self.template_id.report_template.action_report_substitution_rule_ids` ?

report_substitute/static/src/js/action_manager.js
sbejaoui: action dict

sbidoul:
@sbejaoui can you check travis?

report_substitute/README.rst
```diff
@@ -0,0 +1,100 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/143/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module allows you to create substitution rules for report actions.
+A typical use case is to replace a standard report by alternative reports
+when some conditions are met. For instance, it allows to configure alternate
+reports for different companies.
```
astirpe, line 31: To manage alternative reports for different companies, I find the existing OCA module [ir_actions_report_multi_company](https://github.com/OCA/multi-company/tree/11.0/ir_actions_report_multi_company) much easier for a non-technical user.

I would change this line in the readme: instead of mentioning reports for different companies, I would expect some other use case. A real case with some code examples would be here very appreciated.

report_substitute/models/ir_actions_report.py
```diff
@@ -0,0 +1,74 @@
...
+                substitution_report = action_report._get_substitution_report(
+                    action_report.model, active_ids
+                )
+            action.update(action_report.read()[0])
+        return action
+
+    @api.multi
+    def render(self, res_ids, data=None):
+        substitution_report = self.get_substitution_report(res_ids)
+        return super(IrActionReport, substitution_report).render(res_ids, data)
```
astirpe, line 59: No need to pass parameters to the `super()`

report_substitute/views/ir_actions_report.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="ir_actions_report_form_view">
+        <field name="name">ir.actions.report.form (in report_dispatch_base)
```
astirpe: No need to specify the `<field name="name"> ... ` since you are inheriting. Also `report_dispatch_base` is probably some leftover from old code.

report_substitute/models/ir_actions_report.py
```diff
@@ -0,0 +1,74 @@
...
+            if set(model.search(domain).ids) == set(active_ids):
+                return substitution_report_rule.substitution_action_report_id
+        return False
+
+    @api.multi
+    def get_substitution_report(self, active_ids):
+        self.ensure_one()
+        action_report = self
+        substitution_report = action_report
+        while substitution_report:
```
astirpe, line 36: With this while, the matching rule with higher value of the sequence will be executed instead of the previous matching rules. Is this the desired behavior?

astirpe (changes requested):

report_substitute/README.rst
sbejaoui: In standard modules, many report calls are hard-coded by their xml ID. An example, the print quotation action in the sale module.

https://github.com/odoo/odoo/blob/48b3893b84cbd51f880548b38d036db5ebc1e256/addons/sale/models/sale.py#L489-L494

Adding an ir_rule to the report action will not resolve the use case here because this type of action will raise an access error for other companies if a company is set.

the alternative solutions that can be made in this case and that we avoid with this module:
- override the print action to improve the way odoo get the report action.
- override the report template to manage all the layout differences in the same template.

So the idea is to define some substitution rules on the standard report to return the desired one if some condition are met.

sbejaoui: IMO, with the domain widget used in the UI, the the substitution rules is easy to use for non-technical user

report_substitute/models/ir_actions_report.py
sbejaoui: the super is called on a different object then self. have to be specified.

report_substitute/models/ir_actions_report.py
sbejaoui: The purpose of this while is to continue the substitution recursively. A substitution report can have a matching on a one of its substitution rules.

report_substitute/views/ir_actions_report.xml
sbejaoui: The name is required even in the case of inheritance.

good catch for the leftover.

astirpe: If you don't specify the name, Odoo will automatically create one.

report_substitute/models/ir_actions_report.py
rousseldenis: > With this while, the matching rule with higher value of the sequence will be executed instead of the previous matching rules. Is this the desired behavior?

@astirpe Due to this, it's ok https://github.com/OCA/reporting-engine/pull/307/files#diff-53c1b43ed6b9be73de53ef31d1261b38R11

rousseldenis (approved):
LGTM. Code review

rousseldenis:
@astirpe Could you update your review?

astirpe (approved):

report_substitute/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    'name': 'Report Substitute',
+    'summary': """
+        This module allows to create substitution rules for report actions.
+        """,
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/acsone/reporting-engine',
```
etobella: URL should be https://github.com/OCA/reporting-engine

report_substitute/models/ir_actions_report.py
```diff
@@ -0,0 +1,74 @@
...
+    _inherit = 'ir.actions.report'
+
+    action_report_substitution_rule_ids = fields.One2many(
+        comodel_name="ir.actions.report.substitution.rule",
+        inverse_name="action_report_id",
+        string="Substitution Rules",
+    )
+
+    @api.multi
+    def _get_substitution_report(self, model, active_ids):
```
etobella, line 19: Can you add a limit=1, increase every time and once it has reached 50 raise a Exception?
This way at least we know if we have recursion issue. What do you think?
It is not blocking.

etobella (commented):
For me it's good, just a nitpicking and a comment about the recursion issue.
It is not blocking

report_substitute/models/ir_actions_report.py
sbejaoui: I added a check for the substitution infinite loop when creating the rules. can you check it please.

sbidoul:
Looks like all comments have been handled here.

/ocabot merge, please

PR #311, opened 2019-08-07
[12.0][IMP]date to datetime conversion when autotyping

luc-demeyer:
This PR simplifies the syntax in the templates when using ORM date fields:

E;g. self._render("aml.date") will now be converted automatically to datetime before passing the value to xlsxwriter.


sebastienbeau (approved):
LGTM

PR #313, opened 2019-08-08
[12.0][ADD] report_excel, report_excel_demo

kittiu:
This module add new report action of report_type = "excel", which will allow user to create excel form or
report with ease (no coding necessary in generating excel file).

This module leverage capability to generate excel file from module **OCA/server-tools/excel_import_export**

**Report Action** which pass an object + **Dict Instruction** on how to fill data + **Excel template** --> **Output Excel**

For example, creating a print out form will consist of,

1. Create report action for the required model with new report_type = 'excel', and name = <excel_template.xlsx>
2. Create an instruction in XML (or manually set it in XLSX Template). The instruction will tell Odoo how data will be display in the excel file
3. Create a blank excel_template.xlsx, with all heading, labels, formatting or formula as required.

Note:

* This module can ease the development for most reports, but XML instruction can be limited.
  Use **report_xlsx** with total control (by coding)


kittiu:
The easiest way to see how this works is by installing report_excel_demo

pedrobaeza:
There's already `report_xlsx`, so IMO this shouldn't be called `report_excel`, but something more inline with their specific utility. But why using this instead of regular export option? I don't get the module utility. Also you mustn't include a demo module, but put data in demo of the main module.

kittiu:
@pedrobaeza thanks for your comment. May I clarify my view  first :).

In fact, for us, this module is used largely as reporting tools (excel and csv), and very few as import/export.

Without having a **Report Action**, it need to store file in wizard first to click on file to download (vs just click) It also can't work along side with other reports like HTML or QWeb.

So, technically, we see strong use of this report action. And so, as it become another type of report action, my idea to have this as separated module so user can see it as alternative reporting tools, and not the import/export tools.

I think there are 2 options,

1. Have this as alternative reporting tools, and keep it here.
2. Drop this, but include this feature back into excel_import_export.

As I clarify it, I am ok with both. You can make the call? :+1:

pedrobaeza:
Again, I think this is redundant with export tool having profiles already stored, which serves as well for reporting purposes. Which feature is not included in core export that it is here?

kittiu:
> Again, I think this is redundant with export tool having profiles already stored, which serves as well for reporting purposes. Which feature is not included in core export that it is here?

Nothing except being another report action. It is the main purpose though.

pedrobaeza:
So that's why I think this is not adding any value, and it's only 2 click less (on usual flow, click on action, then on export, then select profile, and export).

What you can do is to create a module `report_export_profile` that allows to define a report action for exporting directly using an export profile. That's something with added value and not too much code burden.

kittiu:
@pedrobaeza sorry, I am a bit confused. We didn't have the report action to use yet, that's why I create it. When you say,
`report_export_profile that allows to define a report action `, you mean what report action?

If you exclude the demo file, there are only 3 files for this (I use sample from report_xlsx).
* action_manager_report.js
* main.py
* ir_report.py

Once we have this new action, we can set the profile (template to use) in the action.

pedrobaeza:
@kittiu the question is that IMO you are reinventing the wheel with `excel_import_export` and specially with this one, having the core exporting option. What I'm proposing is to extend core export option (Action > Export one), for being able to be put as report action, but not creating a new engine like `excel_import_Export` does.

kittiu:
@pedrobaeza 

I think I understand you now. When you say "regular export option", I thought you talk about `excel_import_export` all along. :laughing: Really sorry to have confused you.

So....please allow me to assume that you never use the `excel_import_export` for now (as sorry if you already knew it).

> Which feature is not included in core export that it is here?

The `excel_import_export` is totally different in technical concept and serve very different use cases. Normally, we use Odoo's export in **list view**, to export/import like this, which is mostly raw data w/o any color format or formulas, correct?

image

In contrast, the `excel_import_export` is used for import/export **a document** (not list of documents). The excel file exported can be as colorful, with formula, marco objects, etc. (as designed by Excel) as following,

image

The flexibility of how data is presented is achieved through **dictionary instruction** (templates.xml), which is also updated into Odoo's XLSX Template window, and also configurable through UI. This window keeps both excel template file and the instruction how to import/export.

image

**What is the main difference?**
* Odoo's Export --> it write the data into a **newly created excel** file (with 1 sheet only).
* `excel_import_export` --> it lay down data into the designed excel template (as many sheet as desired), **without wipe out whatever excel has been designed for**. When excel is intacted, we get excel at its best -- formulars, macro, data validation, dropdown list (from master data in another sheet), you name it.

In fact, I got to do `excel_import_export` because of 2 challenges I faced in my previous project.
1. In Budget Planing window, user not able to plan in Odoo's UI, and want to export the plan into excel worksheet (with data validation, dropdown list, etc) and then import it back to Odoo (which can trigger another validation, we can has the POST_IMPORT validation method call in xml too)
2. We have too many excel reports (> 50 reports) to do, at first we also use report_xlsx to write each one. It is too much to code. Also as it changes often, having to rewrite the code is not efficient at all. `excel_import_export` allow non-programmer to changes report's looks and feel and also values easily (just open excel to edit the file, and load it back in).

We use it to create good looking form / reports with minimum (wizard and model view at most) or no coding involved.

-------------------------------------------------------------------------------------

I hope this explain a bit. I am not too sure, but from what I know, it is not capable with Odoo's import/export. If it is, it would be new knowledge to me also.

Thank you @pedrobaeza, please feel free to comment.

pedrobaeza:
OK, indeed I haven't tried that module, and I still see it as a complication over the concept of importing things. For exporting things in "rich" format, it might work though, but I'm still reducing any reporting "paper style" for all my customers. It requires more time to format than to obtain information.

Going back to this PR, why not including this inside `excel_import_export` module itself? It's like the button of `mass_editing` that adds the action.

kittiu:
I hope you have chance test it, it also make import easy too. You gives valuable feedback.

I will include it to excel_import_export later on. Thanks!

kittiu:
Moved to https://github.com/OCA/server-tools/pull/1641

PR #317, opened 2019-08-15
[FIX] report_custom_filename: Use model specified on the report to br…

lmignon:
…owse the active records

The active_model into the context is not always the model for which the report is generated. For exemple if you open an invoice from the smart button on a sale order, the active model is 'sale.order' and not 'account.invoice'.

### DIFF
--- a/report_custom_filename/controllers/reports.py
+++ b/report_custom_filename/controllers/reports.py
@@ -24,7 +24,7 @@ class Reports(main.Reports):
             ('report_name', '=', action['report_name']),
             ('download_filename', '!=', False)])
         for report in reports:
-            objects = http.request.session.model(context['active_model'])\
+            objects = http.request.session.model(report.model)\
                 .browse(context['active_ids'])
             generated_filename = mail_template.mako_template_env\
                 .from_string(report.download_filename)\


pedrobaeza:
But if the `active_model` is not correct, why `active_ids` should be then?

lmignon:
@pedrobaeza Here it's what I've observed: active_id and active_model refers the sale.order. In the same time active_ids is the id of the account.invoice. This sound like a bug into the js part of Odoo but Odoo 9 is no more supported and I don't want to spend time to debug Odoo. Moreover the proposed fix is more consistent since we take the model name on the report definition. (same logic as in the report_controller https://github.com/OCA/reporting-engine/blob/9.0/report_custom_filename/controllers/report_controller.py#L27)

pedrobaeza:
And isn't possible to use `docids` or similar for avoiding `active_ids` as well? Any way, I'm not blocking this, but please check Travis fail.

lmignon:
@pedrobaeza Travis is failing for errors outside this addon. I don't know the failing addon since I've never used it before.
I'm in a delivery rush... no time to check this error at this moment...

pedrobaeza:
OK, I will try to check later if this is something in main branch, although I'm not using 9.0 anymore.

PR #318, opened 2019-08-20
fix(bi.sql.view.field): fix decorator for create

agb80:
Create function must be decorated with @api.model and not with
@api.multi as is now

### DIFF
--- a/bi_sql_editor/models/bi_sql_view_field.py
+++ b/bi_sql_editor/models/bi_sql_view_field.py
@@ -108,7 +108,8 @@ class BiSQLViewField(models.Model):
                 sql_field.bi_sql_view_id.view_name, sql_field.name)
 
     # Overload Section
-    @api.multi
+    @api.model
+    @api.returns('self', lambda value:value.id)
     def create(self, vals):
         field_without_prefix = vals['name'][2:]
         # guess field description


astirpe (approved):
LGTM

bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -108,7 +108,8 @@ def _compute_index_name(self):
                 sql_field.bi_sql_view_id.view_name, sql_field.name)
 
     # Overload Section
-    @api.multi
+    @api.model
+    @api.returns('self', lambda value:value.id)
```
legalsylvain, line 112: i don't understand why adding the second decorator. (@api.returns). Should works without dont you think ?

other wise, LGTM.

astirpe: @legalsylvain looking at how the create is used in the code, it seems the return value must be an id, not a recordset

agb80: @legalsylvain just as @astirpe saids, that is how the code is decorated on main models.py odoo function

legalsylvain: Sorry, I don't understand. The function currently return a recordset.

see : ``return super(BiSQLViewField, self).create(vals)``

So, I think that @api.model will works.

legalsylvain (commented):
Could you check travis & runbot ? 

thanks.

Issue #319, opened 2019-08-28
[RFC] Header & Footer for report_xlsx


apineux:
The idea is to be able to define a excel header and a footer on the model `ir.actions.report.xml` that have the `report_type == 'xlsx'`.
The header and footer will use the xlsxwriter syntax (see example here https://xlsxwriter.readthedocs.io/example_headers_footers.html)
We will create a separate model to store the headers and footers so that we just have to select them as a many2one on the report.
This feature will therefore be available for mis_builder or account_financial_report as they both use the report_xlsx.

This is a simple development as we only use the xlsxwriter syntax but feel free to share your opinion and suggestions.

PR #320, opened 2019-08-28
[10.0][ADD] Header and Footer configuration for XLSX report

apineux:
PR for issue https://github.com/OCA/reporting-engine/issues/319

**How to configure Header / Footer:**
On the menu `Settings > Technical > Reports > XLSX Header/Footer` you can create Header or Footer following the documentation https://xlsxwriter.readthedocs.io/page_setup.html#set_header
header

You can add a Header and a Footer on the reports of type 'xlsx' such as here:
report_xlsx

Once you exported your excel file, when you print it or do a print preview, you will see the Header and or Footer you configured.
print_xlsx

### DIFF
--- a/report_xlsx/models/header_footer.py
+++ b/report_xlsx/models/header_footer.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import io
+import ast
+from odoo import fields, models, api, _
+from odoo.exceptions import ValidationError
+
+
+class ReportHeaderFooter(models.Model):
+    _name = 'report.xlsx.hf'
+
+    name = fields.Char(string="Name", required=True)
+    hf_type = fields.Selection(
+        [('header', 'Header'), ('footer', 'Footer')],
+        string="Type",
+        required=True)
+    value = fields.Char(string="Value")
+    manual_options = fields.Char(string="Options")
+    image_left = fields.Binary(string='Image left')
+    image_left_name = fields.Char('File Name')
+    image_center = fields.Binary(string='Image center')
+    image_center_name = fields.Char('File Name')
+    image_right = fields.Binary(string='Image right')
+    image_right_name = fields.Char('File Name')
+    header_report_ids = fields.One2many(
+        'ir.actions.report.xml',
+        'header_id',
+        string="Associated report(s)")
+    footer_report_ids = fields.One2many(
+        'ir.actions.report.xml',
+        'footer_id',
+        string="Associated report(s)")
+
+    @api.multi
+    @api.constrains('manual_options')
+    def _check_manual_options(self):
+        for rec in self:
+            if rec.manual_options:
+                options = ast.literal_eval(rec.manual_options)
+                if not isinstance(options, dict):
+                    raise ValidationError(
+                        _('The Header/Footer is not configured properly.\
+                                     Options must be a dictionary.'))
+
+    @api.multi
+    @api.constrains('image_left', 'image_center', 'image_right')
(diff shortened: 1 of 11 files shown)


report_xlsx/models/header_footer.py
```diff
@@ -0,0 +1,64 @@
...
+    footer_report_ids = fields.One2many(
+        'ir.actions.report.xml',
+        'footer_id',
+        string="Associated report(s)")
+
+    @api.multi
+    @api.constrains('manual_options')
+    def _check_manual_options(self):
+        for rec in self:
+            options = safe_eval(rec.manual_options)
```
sbidoul: In this case, `ast.literal_eval` is probably sufficient, since there are no context variables to evaluate. It's safer.
Also, can you handle the case where manual_options is not set? (currently it errors)

report_xlsx/models/header_footer.py
```diff
@@ -0,0 +1,64 @@
...
+import io
+from odoo import fields, models, api, _
+from odoo.tools.safe_eval import safe_eval
+from odoo.exceptions import ValidationError
+
+
+class ReportHeaderFooter(models.Model):
+    _name = 'report.xlsx.hf'
+
+    name = fields.Char(string="Name")
```
sbidoul: Would it be better to make the name required?

report_xlsx/views/ir_report.xml
```diff
@@ -0,0 +1,20 @@
...
+    <record id="act_report_xml_view_inherit" model="ir.ui.view">
+        <field name="name">ir.actions.report.xml (xlsx header footer)</field>
+        <field name="model">ir.actions.report.xml</field>
+        <field name="inherit_id" ref="base.act_report_xml_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page string="Header/Footer" attrs="{'invisible': [('report_type', '!=', 'xlsx')]}">
+                    <group>
+                        <field name="header_id"/>
+                        <field name="footer_id"/>
```
sbidoul: I suggest adding default_hf_type in context so when the user creates from here it defaults to the correct value

sbidoul (commented):
Looks very nice.

Could you add a couple of screenshots for reviewers?

Also I did not succeed adding an image. I got the following error in the log, but to user visible error except the header did not display.
```
UserWarning: Number of header images (1) doesn't match placeholder count (0) in string: &LHello &CWorld
  % (image_count, placeholder_count, header_orig))
```

I did not investigate yet. Could you perhaps add some more help on how to add an image, and look if it's possible to raise an exception in such cases?

apineux:
@sbidoul Thanks for the review, I'll do the corrections.
Concerning images, you must follow the syntax from https://xlsxwriter.readthedocs.io/page_setup.html#set_header
In the form view of the header / footer I already added the link to that documentation and I also already add the following message "If you select an image, you don't have to specify any image options (image_left, image_data_left, ...) but you need to add the control character &G in the value."
I guess that you didn't follow the instructions and forgot to put the placeholder &G as indicated ;)

sbidoul:
Sure I don't read the instructions :) Or rather I did, but I was confused by the fact there were buttons to add images named left, right, center and it was not obvious a placeholder was needed in addition. Maybe some additional help text in the form or a warning would have helped. Especially since it failed silently and, as a user who can't read the logs, I had no way to understand what was happening.

apineux:
@sbidoul A description has been added on the top of this topic. 
More checks and helpers have been added while configuring the Header / Footer.
Thanks for the review

sbidoul (approved):
Code review ok. Functional test ok, including images. The explanation in the configuration form is easy enough to understand now. Thanks!

lmignon (approved):
Thank you @apineux For this new functionality. LGTM (Code review)

sbidoul:
squashed

/ocabot merge minor

PR #315, opened 2019-08-09
[12.0][IMP] report_xlsx, make sure print_report_name works

kittiu:
This improve make sure `print_report_name` field in report action will works, as it not work previously.

Note: do I need to set minor version to 12.0.1.0.1 ? Is it updated auto by Bot?

PR #325, opened 2019-09-23
report_xlsx: add custom filename to response headers

jffernandez:


jffernandez:
Sorry, duplicate of #315, that can be backported.
I will use my version while not backported.
Closing.

PR #327, opened 2019-09-25
[11.0][IMP] report_xlsx, make sure print_report_name works

kittiu:
@jffernandez would you help test again? :)

### DIFF
--- a/report_xlsx/controllers/main.py
+++ b/report_xlsx/controllers/main.py
@@ -3,8 +3,10 @@
 
 from odoo.addons.web.controllers import main as report
 from odoo.http import content_disposition, route, request
+from odoo.tools.safe_eval import safe_eval
 
 import json
+import time
 
 
 class ReportController(report.ReportController):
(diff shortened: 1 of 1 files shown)


report_xlsx/controllers/main.py
```diff
@@ -29,13 +31,18 @@ def report_routes(self, reportname, docids=None, converter=None, **data):
             xlsx = report.with_context(context).render_xlsx(
                 docids, data=data
             )[0]
+            report_name = report.report_file
+            if report.print_report_name and not len(docids) > 1:
```
jffernandez: Here you should check that docids is set:
`if report.print_report_name and docids and not len(docids) > 1:`
Just realized that sometimes a report is launched without docids, and it will fail.
In that case is fair to not personalize the name because a main object is not given and the data property would contain the details for rendering the report.

jffernandez (changes requested):
It's working, but could fail in special cases, look inline comment.
Please, patch the v12 with this too. Sorry for not seeing before.
Thank you!

report_xlsx/controllers/main.py
kittiu: Thank you for noticing. I fixed per your request here and also in v12

jffernandez (approved):
Now is safe and working correctly.
Thank you!

carlosdauden (approved):

pedrobaeza:
Is this needed for 12.0?

carlosdauden:
It seems that in v12 is already:
https://github.com/OCA/reporting-engine/blob/12.0/report_xlsx/controllers/main.py#L34-L38

kittiu:
@pedrobaeza
https://github.com/OCA/reporting-engine/pull/315
It was merged.

pedrobaeza:
Ok, thanks for confirming

PR #333, opened 2019-10-17
[IMP] report_xlsx: print_report_name field support

carlosdauden:
@Tecnativa


pedrobaeza:
Isn't the same as #327?

carlosdauden:
You are right, It is the same as https://github.com/OCA/reporting-engine/pull/327

PR #337, opened 2019-10-22
[11.0][ADD] Header and Footer configuration for XLSX report

apineux:
Migration in 11.0 of https://github.com/OCA/reporting-engine/pull/320

### DIFF
--- a/report_xlsx/README.rst
+++ b/report_xlsx/README.rst
@@ -55,6 +55,15 @@ A report XML record ::
         attachment_use="False"
     />
 
+**XLSX Header & Footer**
+
+You can configure them on the menu *Settings > Technical > Reports > XLSX Header/Footer* following the syntax from 
+`xlsxwriter documentation <https://xlsxwriter.readthedocs.io/page_setup.html#set_header>`_.
+
+Example of Header / Footer syntax : ``&LPage &P of &N &CFilename: &F &RSheetname: &A``
+
+On a report XML with ``report_type == 'xlsx'`` you can specified the Header and Footer you configured.
+
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/143/11.0
@@ -75,6 +84,7 @@ Contributors
 
 * Adrien Peiffer <adrien.peiffer@acsone.eu>
 * Enric Tobella <etobella@creublanca.es>
+* Arnaud Pineux <arnaud.pineux@acsone.eu>
 
 Maintainer
 ----------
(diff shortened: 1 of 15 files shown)


report_xlsx/README.rst
```diff
@@ -55,6 +55,15 @@ A report XML record ::
...
 
+**XLSX Header & Footer**
+
+You can configure them on the menu *Settings > Technical > Reports > XLSX Header/Footer* following the syntax from 
+`xlsxwriter documentation <https://xlsxwriter.readthedocs.io/page_setup.html#set_header>`_.
+
+Example of Header / Footer syntax : ``&LPage &P of &N &CFilename: &F &RSheetname: &A``
+
+On a report XML with ``report_type == 'xlsx'`` you can specified the Header and Footer you configured.
+
```
sbidoul, line 66: This part must go in `readme/USAGE.rst`, from which `README.rst` is generated.

sbidoul (dismissed):
Functional test + code review. Thanks.

Could you squash your commits?

lmignon (approved):
LGTM (Code review + Functional test)

report_xlsx/report/report_xlsx.py
```diff
@@ -58,4 +69,4 @@ def get_workbook_options(self):
         return {}
 
     def generate_xlsx_report(self, workbook, data, objs):
-        raise NotImplementedError()
+        return
```
sbidoul: @apineux why changing this?

apineux: @sbidoul Because if I keep the raise, the tests will failed. In the test we call create_workbook directly on the ReportXlsxAbstract model instead of create a new model based on ReportXlsxAbstract where we implement the method generate_xlsx_report.

sbidoul: I would prefer if the method would be mocked in the test, as raising NotImplementedError is a good way to communicate that the abstract method needs to be implemented. @lmignon would you have an example at hand?

lmignon: @sbidoul @apineux

To mock a method on a model, the best approach is to mock the class from the registry...
```python
report_abstract = self.env["report.report_xlsx.abstract"]
with mock.patch.object(report_abstract.__class__, "generate_xlsx_report") as mocked_method:
    # the test
    ...
    report_abstract. ...

    # and you can check that the method has been called//
    self.assertEqual(mocked_method.call_count, 1)
    
    # if you want you can also check the params received by the method
    mocked_method.assert_called_once_with(.....)
    # or
    mocked_method.assert_called_with(.....)

```

By mocking the method you can let the `NotImplementedError` as default implementation of this abstract method

apineux:
@sbidoul The latest requested changes are made.

andreampiovesana (approved):
merge?

PR #338, opened 2019-10-26
[10.0][FIX] report_xlsx: missing dependency

sbidoul:


sbidoul:
fast tracking this trivial patch to greenify the mis builder test suite

/ocabot merge patch

PR #344, opened 2019-11-14
[11.0][ENH] sql view - allow definition of group operator

richard-willdooit:
Adds the possibility, for float and integer columns, to
apply a group operator (average, min, max).

### DIFF
--- a/bi_sql_editor/__manifest__.py
+++ b/bi_sql_editor/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'BI SQL Editor',
     'summary': 'BI Views builder, based on Materialized or Normal SQL Views',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.1.0',
     'license': 'AGPL-3',
     'category': 'Reporting',
     'author': 'GRAP,Odoo Community Association (OCA)',
--- a/bi_sql_editor/models/__init__.py
+++ b/bi_sql_editor/models/__init__.py
@@ -2,3 +2,4 @@
 
 from . import bi_sql_view
 from . import bi_sql_view_field
+from . import ir_model
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -8,7 +8,7 @@ from psycopg2 import ProgrammingError
 
 from odoo import _, api, fields, models, SUPERUSER_ID
 from odoo.exceptions import UserError
-from odoo.tools import pycompat, sql
+from odoo.tools import pycompat, sql, table_columns
 from odoo.addons.base.ir.ir_model import IrModel
 
 _logger = logging.getLogger(__name__)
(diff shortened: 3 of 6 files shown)


richard-willdooit:
@pedrobaeza I think this is a very useful extension to the module.

pedrobaeza:
Good, but I'm not a direct user of this module, so you should wait for appropriate reviewers. You can encourage them reviewing other PRs in this repo and asking the review of yours in return.

CasVissers-360ERP:
@richard-willdooit 
Very nice addition to the module!

I get:
```
  File "/opt/odoo/odoo/addons/base/ir/ir_cron.py", line 222, in _process_jobs
    registry = odoo.registry(db_name)
  File "/opt/odoo/odoo/__init__.py", line 76, in registry
    return modules.registry.Registry(database_name)
  File "/opt/odoo/odoo/modules/registry.py", line 61, in __new__
    return cls.new(db_name)
  File "/opt/odoo/odoo/modules/registry.py", line 85, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/odoo/modules/loading.py", line 383, in load_modules
    registry.setup_models(cr)
  File "/opt/odoo/odoo/modules/registry.py", line 274, in setup_models
    model._setup_base()
  File "/opt/odoo/odoo/models.py", line 2401, in _setup_base
    self.env['ir.model.fields']._add_manual_fields(self)
  File "/opt/cust/community/bi_sql_editor/models/ir_model.py", line 14, in _add_manual_fields
    Sql.check_manual_fields(model)
  File "/opt/cust/community/bi_sql_editor/models/bi_sql_view.py", line 656, in check_manual_fields
    self.search([('model_name', '=', model._name)]
  File "/opt/odoo/odoo/fields.py", line 952, in __get__
    self.determine_value(record)
  File "/opt/odoo/odoo/fields.py", line 1055, in determine_value
    record._prefetch_field(self)
  File "/opt/odoo/odoo/models.py", line 2663, in _prefetch_field
    result = records.read([f.name for f in fs], load='_classic_write')
  File "/opt/odoo/odoo/models.py", line 2601, in read
    self._read_from_database(stored, inherited)
  File "/opt/odoo/odoo/models.py", line 2762, in _read_from_database
    field.read(fetched)
  File "/opt/odoo/odoo/fields.py", line 2277, in read
    get_id = (lambda rec: rec.id) if comodel._fields[inverse].type == 'many2one' else int
KeyError: 'bi_sql_view_id'
```

Am I doing something wrong?

richard-willdooit:
@CasVissers I think this is only happening if you are upgrading?  Can you at least try this on a fresh install?  
I am thinking of pushing a refactored version which leaves bi_sql_editor module intact, and adds this functionality as a newly installable module which depends on it.  This will allow for clean addition of the feature in to environments which already have the other module installled.

CasVissers-360ERP:
@richard-willdooit 
True, this happens when upgrading. Tested on OCA runbot, functionality is great!

But I think the upgrading issue needs to be solved before this improvement can be rolled out.

HviorForgeFlow:
@CasVissers are you applying this configuration?
https://github.com/OCA/reporting-engine/tree/12.0/bi_sql_editor#installation

CasVissers-360ERP:
@hveficent 
I use:
server_wide_modules = web,sentry,queue_job,bi_sql_editor, should also do the trick right?

HviorForgeFlow:
@CasVissers it should allow you to update the module. Still failing with this configuration?

CasVissers-360ERP:
@hveficent 
Tried couple of things.

1:
 ./odoo-bin -c /etc/odoo-test-server.conf -u bi_sql_editor -d xxx --stop-after-init
with bi_sql_editor in server_wide_modules

Run it the first time: Traceback
Ran it a second time: Updated module, but the grouping didn't work.

2:
./odoo-bin -c /etc/odoo-test-server.conf -u bi_sql_editor -d xxx --load=bi_sql_editor
bi_sql_editor not in config server_wide_modules
Back to traceback

richard-willdooit:
@CasVissers @hveficent Yes, I agree - due to the inheritance of the low level methods, I cannot quite get it right so it will work in an upgrade AND a fresh install.

I have now defined it as a separate module - and will push that change when I iron 1 or 2 things out in my production environment - and then it will not break for existing installs, and works correctly by installing the additional module...

Stay tuned....

richard-willdooit:
@CasVissers @hveficent See #351 and #352

PR #345, opened 2019-11-14
[12.0][ENH] sql view - allow definition of group operator

richard-willdooit:


richard-willdooit:
@pedrobaeza And for version 12

bi_sql_editor/models/ir_model.py
```diff
@@ -0,0 +1,14 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class IrModelFields(models.Model):
+    _inherit = 'ir.model.fields'
+
+    def _add_manual_fields(self, model):
+        super(IrModelFields, self)._add_manual_fields(model)
```
bealdav, line 10:
```suggestion
        super()._add_manual_fields(model)
```

richard-willdooit: > 

Cool - I have not seen that use of super in Odoo.  Just had a look and saw a half dozen so uses in v12 addons - is that something that changed since 12 only?

pedrobaeza: This is something of Python 3, that allows this simplified syntax. It's not mandatory though.

richard-willdooit:
Refactored as additional module.  #352

PR #352, opened 2019-11-28
[12.0][ENH] sql view - allow group operator

richard-willdooit:
Replaces #345

Adds the possibility, for float and integer columns, to
apply a group operator (average, min, max).


bealdav:
@legalsylvain

bi_sql_editor_aggregate/models/bi_sql_view_field.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = 'bi.sql.view.field'
+
+    _GROUP_OPERATOR_SELECTION = [
+        ('sum', 'Sum'),
+        ('avg', 'Average'),
+        ('min', 'Minimum'),
+        ('max', 'Maximum'),
+    ]
+
+    _GRAPH_TYPE_SELECTION = [
```
legalsylvain: where it is used ?

bi_sql_editor_aggregate/models/ir_model.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class IrModelFields(models.Model):
+    _inherit = 'ir.model.fields'
+
+    def _add_manual_fields(self, model):
+        super()._add_manual_fields(model)
+        if 'bi.sql.view' in self.env:
+            Sql = self.env['bi.sql.view']
+            if hasattr(Sql, 'check_manual_fields'):
```
legalsylvain, line 13: why this check ? 
I guess that it is allways true, if the module is installed ? or did I missed something ?

legalsylvain (commented):
Hi @richard-willdooit 
Thanks a LOT, for this improvment !
That is for sure a big limitation of the current module ``bi_sql_editor``.
What about to integrate this code to the main module. I don't see the interest to add a very important feature in an extra module. What do you think ?

(partial code review for the time being).

bi_sql_editor_aggregate/models/ir_model.py
richard-willdooit: @legalsylvain  This check MAY be unnecessary now, but see my comment below....

richard-willdooit:
@legalsylvain I first wrote this as an integrated part of the module - see #344 and #345 - however, due to the override of add_manual_fields, upgrading when adding the module turned in to a nightmare - there were times during the instanciation of the models when certain fields and attributes were not yet available or visible.  As it is, we need to be careful if it gets merged in to the one module.

I was much happier integrating it to the single module, but I had trouble upgrading two production instances for different issues (it depended on the order that models and fields were being added) - in the worse of the two, I had a maximum recursion every time the odoo server started, and I was only able to get the module installed by manipulating column definitions directly with PostgreSQL - not pretty.  @CasVissers had a similar problem on #344 

I am happy to re-propose the previous two PRs, but I cannot work out a clean and always reproducable upgrade :confused:

richard-willdooit:
@legalsylvain  Do you know what we need to do to get this merged?

legalsylvain (approved):
/ocabot merge

bealdav (approved):

PR #355, opened 2019-12-10
[12.0][ADD] report_async

kittiu:
The new menu "Report Center" is the central place to host all your reports in one place.
From here, there are 2 ways to launch the report,

1. Run Now - run report immediately as per normal.
2. Run Background - put the report execution to queue job.

report_async


kittiu:
@Saran440 can you review this?

kittiu:
This module is ready for review / merge

Note: --test-enable can result in fallback to type html (it is in Odoo core). So, when you test by hand, make sure you don't use --test-enable on your environment.

sbidoul:
Very interesting.

Regarding the mechanism to make the file available to the user, I have a preference for the one in [base_export_async](https://github.com/OCA/queue/tree/13.0/base_export_async) which sends an email with a link to the report that only the user can download. It also has a mechanism to remove these files after a while.

kittiu:
@sbidoul we have both, do you think it is ok? In fact, base_export_async was my example for this.
https://github.com/ecosoft-odoo/reporting-engine/blob/12.0-add-report_async/report_async/models/report_async.py#L158

report_async/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+{
+    'name': 'Report Async',
+    'summary': 'Central place to run reports live or async',
+    'version': '12.0.1.0.0',
+    'author': 'Ecosoft, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'website': 'https://github.com/OCA/queue',
```
rousseldenis:
```suggestion
    'website': 'https://github.com/OCA/reporting-engine',
```

report_async/models/report_async.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+    file_ids = fields.Many2many(
+        comodel_name='ir.attachment',
+        compute='_compute_file',
+        help="List all files created by this report background process",
+    )
+
+    @api.multi
+    def _compute_job(self):
+        for rec in self:
```
rousseldenis, line 75:
```suggestion
        for rec in self.sudo():
```
At each call of sudo(), a new environment is created and can lead to performance issues.

report_async/models/report_async.py
```diff
@@ -0,0 +1,181 @@
...
+                order='id desc')
+            rec.job_status = (rec.job_ids[0].sudo().state
+                              if rec.job_ids else False)
+            rec.job_info = (rec.job_ids[0].sudo().exc_info
+                            if rec.job_ids else False)
+
+    @api.multi
+    def _compute_file(self):
+        for rec in self:
+            rec.file_ids = self.env['ir.attachment'].search(
```
rousseldenis: Move the search before the loop with ('res_id', 'in', self.ids) and then filter.

It can avoid many sql queries.

report_async/models/report_async.py
```diff
@@ -0,0 +1,181 @@
...
+        odoo_bot = self.sudo().env.ref('base.partner_root')
+        email_from = odoo_bot.email
+        description = attachment.name
+        self.env['mail.mail'].create({
+            'email_from': email_from,
+            'reply_to': email_from,
+            'recipient_ids': [(6, 0, [user.partner_id.id])],
+            'subject': _("Export {} {}").format(
+                description, fields.Date.to_string(fields.Date.today())),
+            'body_html': _("""
```
rousseldenis: Mmmmh, maybe a mail.template ?

rousseldenis (changes requested):
Code review. Apart of that, this is great !

rousseldenis:
@OCA/core-maintainers

moylop260 (approved):

report_async/models/report_async.py
kittiu: Or should I remove all the .sudo() from _compute_job()
and use `compute_sudo=True` in all fields that call _compute_job instead?
WDYT?

report_async/models/report_async.py
kittiu: Great idea. Fixed.

report_async/models/report_async.py
kittiu: Done using a template.

image

dong-z: Add a widget in the navigation drawer to centrally notify the report print results.
How about this suggestion？

dong-z: This is great anyway

jarroyomorales (approved):
Tested locally and LGTM!
Just two questions. Is the dashboards app the best place for the report center? And should we create a security group to better control who accesses this?
Thank you for this awesome module!

etobella (commented):
Thanks for the effort!!! I think this is really a good implementation!!!
About the widget on the bar, that could be a great idea, but we can do this on a new PR, this way we can proceed. On the other hand, I agree about the need of a group.

kittiu:
About groups, the first idea is allow everyone to create his/her own report async. And have someone (group_no_one) to see them all.
I still not super sure what kind of security structure to best works in real lift, except anyone can help me decide. Otherwise I suggest we merge this first (given it is good enough). Then, people can use it and propose what best fit.

WDYT?

etobella:
Good for me.

etobella (approved):

jarroyomorales:
@kittiu Okay, I agree with you, we can merge this and discuss it later.

JordiBForgeFlow (approved):

PR #365, opened 2020-01-23
[12.0][FIX] - report_py3o: run libreoffice in an isolated user installation

sbejaoui:
Bug when more than one conversion print is launched within the same libreoffice instance.

The standard behavior of libreoffice when a user open it while another instance is
running is to show a new window and throw an error if a new instance is forced within
the same user installation [see](https://bugs.documentfoundation.org/show_bug.cgi?id=37531).

This implies a bug in report_py3o module when we call libreoffice at the same time for
different documents.

To reproduce this bug:

**Case 1:**
1.  Simultaneously print two documents.

**Case 2:**
1.  Run print jobs using job_queue module
2.  Manually print another document

**Case 3:**
2.  Open libreoffice
3.  Print a py3o report

This PR creates a temporary user installation for each libreoffice conversion to bypass this limitation.

cc/ @sbidoul

### DIFF
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -243,26 +243,32 @@ class Py3oReport(models.TransientModel):
     def _convert_single_report(self, result_path, model_instance, data):
         """Run a command to convert to our target format"""
         if not self.ir_actions_report_id.is_py3o_native_format:
-            command = self._convert_single_report_cmd(
-                result_path, model_instance, data,
-            )
-            logger.debug('Running command %s', command)
-            output = subprocess.check_output(
-                command, cwd=os.path.dirname(result_path),
-            )
-            logger.debug('Output was %s', output)
-            self._cleanup_tempfiles([result_path])
-            result_path, result_filename = os.path.split(result_path)
-            result_path = os.path.join(
-                result_path, '%s.%s' % (
-                    os.path.splitext(result_filename)[0],
-                    self.ir_actions_report_id.py3o_filetype
+            with tempfile.TemporaryDirectory() as tmp_user_installation:
+                command = self._convert_single_report_cmd(
+                    result_path,
+                    model_instance,
+                    data,
+                    user_installation=tmp_user_installation,
                 )
-            )
+                logger.debug('Running command %s', command)
+                output = subprocess.check_output(
+                    command, cwd=os.path.dirname(result_path),
+                )
+                logger.debug('Output was %s', output)
+                self._cleanup_tempfiles([result_path])
+                result_path, result_filename = os.path.split(result_path)
+                result_path = os.path.join(
+                    result_path, '%s.%s' % (
+                        os.path.splitext(result_filename)[0],
+                        self.ir_actions_report_id.py3o_filetype
+                    )
+                )
         return result_path
 
     @api.multi
-    def _convert_single_report_cmd(self, result_path, model_instance, data):
+    def _convert_single_report_cmd(
+        self, result_path, model_instance, data, user_installation=None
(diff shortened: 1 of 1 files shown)


report_py3o/models/py3o_report.py
```diff
@@ -265,18 +267,23 @@ def _convert_single_report(self, result_path, model_instance, data):
...
+        return (
+            [
+                lo_bin,
+                '-env:UserInstallation=file:%s' % tmp_user_installation,
+                '--headless',
+                '--convert-to',
+                self.ir_actions_report_id.py3o_filetype,
+                result_path,
+            ],
+            tmp_user_installation,
```
sbidoul: Why do we need to return tmp_user_installation to the caller? Can't we clean it here?

sbidoul:
@hbrunn @lmignon do you have an opinion about this? Have you ever encountered this issue?

report_py3o/models/py3o_report.py
hbrunn: no, we need this to exist for the actual call to libreoffice to succeed. But I'm also concerned about breaking the API here, I'd prefer to see the the temporary directory passed in an optional keyword argument. If passed, we add the parameter to the command, and the caller is responsible for cleaning up.

hbrunn:
I'm a little bit surprised by the repeated mention of the fusion server, as this is exactly the code path where we don't use the fusion server.

Myself I never encountered the issue described, but then we also use this only for a handful of reports that are very unlikely to be run anywhere close to parallel. That said, the bug description makes sense to me, I immediately believe this can be a problem. So as soon as this doesn't break the API, I'd be fine with the change.

report_py3o/models/py3o_report.py
sbidoul: right, I skimmed on it too quickly. So yes, better doing a `with TemporaryDirectory()` in the caller and pass the temp dir to `_convert_single_report`.

sbejaoui:
> I'm a little bit surprised by the repeated mention of the fusion server, as this is exactly the code path where we don't use the fusion server.

Thank you for your review,

I misunderstood the usage of the fusion server as I am not familiar with the module. I updated the PR description.

report_py3o/models/py3o_report.py
sbejaoui: I pushed a new version to not break the API

lmignon (approved):
Thank you for this fix @sbejaoui. All this makes sens according to https://ask.libreoffice.org/en/question/14841/how-can-i-use-soffice-from-the-command-line-when-quickstarter-is-running/

report_py3o/models/py3o_report.py
```diff
@@ -243,40 +243,49 @@ def _create_single_report(self, model_instance, data):
...
+    ):
         """Return a command list suitable for use in subprocess.call"""
         lo_bin = self.ir_actions_report_id.lo_bin_path
         if not lo_bin:
             raise RuntimeError(
                 _("Libreoffice runtime not available. "
                   "Please contact your administrator.")
             )
-        return [
+        params = [
```
sbidoul: nit: I'd name this variable cmd to match the method name.

sbidoul (approved):

hbrunn (approved):

report_py3o/models/py3o_report.py
hbrunn: if you do this nit, please also add a comma after `None`, but I don't insist on it. Who does the honors?

sbidoul: actually, black does not add a comma in this particular case

PR #369, opened 2020-02-12
[13.0][MIG] report batch

Khalid-SerpentCS:
This module will allow to to print multiple QWeb reports in a single batch.
For example, picking, packing, reports needs to be printed together
in a single batch.


report_batch/models/ir_action_report.py
```diff
@@ -2,86 +2,98 @@
...
+        )
 
     def generate_bottom_part(self):
         return """\n
         \t\t</t>\n\t\t"""
 
     def generate_custom_content(self, report_name):
-        return """\n
-        \t<t t-call="%s"/>""" % report_name
+        return (
```
dreispt, line 46: Why the braces? I don't think they're necessary.

dreispt (commented):
LGTM, but can you restore the name of the original commits - they now have "PATCH" added to them.

dreispt:
@Khalid-SerpentCS Please fix conflicts. A rebase should help.

PR #375, opened 2020-03-04
[11.0][FIX] use render as other modules may be modifying it

jarroyomorales:
Useful in order to integrate it with report_context


jarroyomorales:
This happens, for example if you install first the report_xlsx and then the report_context. report_context will not add the context, as the funcion of report_xlsx will be executed before and not execute super

jarroyomorales:
The other option would be to add a sequence on the report_context

etobella (approved):
LGTM

PR #387, opened 2020-04-14
[12.0][ADD] kpi_dashboard

etobella:
This module adds new kinds of dashboards on a specific new type of view.

Configuration
=============

Configure KPIs
------------

#. Access `Dashboards > Configuration > KPI Dashboards > Configure KPI`
#. Create a new KPI specifying the computation method and the kpi type

   #. Number: result must contain a `value` and, if needed, a `previous`
   #. Meter: result must contain `value`, `min` and `max`
   #. Graph: result must contain a list on `graphs` containing `values`, `title` and `key`

Configure dashboards
-----------------

#. Access `Dashboards > Configuration > KPI Dashboards > Configure Dashboards`
#. Create a new dashboard and specify all the standard parameters on `Widget configuration`
#. Append elements on KPIs
#. You can preview the element using the dashboard view
#. You can create the menu entry directly using the `Generate menu` button

Here, we can see the demo using the bus with the automatic update

Peek 14-04-2020 09-51

@JordiBForgeFlow @LoisRForgeFlow @jarroyomorales


AaronHForgeFlow:
Looks awesome! :heart: 

I was trying to play I bit on runbot. But I could not get the dashboard for the demo kpi that are defined.

When I go to the KPI Dashboard and I click on "view dashboard" (see screenshot below) nothing happens.
image

Same thing if I create specific menu for the dashboard.

Do I need to configure anything else? 

BTW It would be nice to have a "help section" in the KPI configurations menu. 

Thank you!

etobella:
Now it works. There were some missing libraries :wink:

AdriaGForgeFlow (approved):
Awesome work!! Tested on local with demo data and works perfectly!

Just a nitpicking. Maybe limiting the size of the kpi would be a good idea because currently if you set a big size on one of the kpi it will stop showing on the dashboard and a white space will be rendered instead.
Peek 2020-04-14 12-12

etobella:
@AdriaGForgeFlow Comments attended. I also added some constrains in order to avoid widgets crossing each other

etobella:
By the way, KPIs are updated automatically by Bus

AaronHForgeFlow (approved):
Tested successfully on local. Awesome.

LoisRForgeFlow (approved):
Functional review in local :+1:

pedrobaeza:
Not sure if this the best hosting, and it seems this is mixing a new view type with the KPI definition, isn't it? Does it make sense to split the view part from the KPI definition?

Functionally, in a very brief review showing the GIF, I have several points:

- Can this be added to the Odoo regular dashboards for integral view there? You may want to mix lists, graphs, and this KPI view.
- Is logical that users go first to the definition of the KPIs and switch to the KPI view? Shouldn't they be isolated? Or maybe is only some kind of "preview"?

etobella:
> Not sure if this the best hosting, and it seems this is mixing a new view type with the KPI definition, isn't it? Does it make sense to split the view part from the KPI definition?
> 
> Functionally, in a very brief review showing the GIF, I have several points:
> 
>     * Can this be added to the Odoo regular dashboards for integral view there? You may want to mix lists, graphs, and this KPI view.
Interesting!!! I will try to make this possible :wink: 
> 
>     * Is logical that users go first to the definition of the KPIs and switch to the KPI view? Shouldn't they be isolated? Or maybe is only some kind of "preview"?

This is just the preview, when you define the menu entry, you will be sent directly to the dashboard view. Also, I have added the possibility to add an action for a KPI. When defined, you can click on it and go directly there. Let me try an example:

- We define a Dashboard for your 10 KPIs. Some of them will be updated every day, other instantly and some of them just once a month.
- When you want to dive on your data, you select your KPI and access directly the to your view.

Let me try with an example:
Action

Menu

pedrobaeza:
Thanks for the extra info. I see the point in your approach, but as said I think it's more interesting to be added to regular dashboard if possible, but indeed this is a very huge improvement. Another one can be the graphical definition of the KPIs instead of doing it through formulas.

kpi_dashboard/models/kpi_dashboard.py
```diff
@@ -0,0 +1,179 @@
...
+    def _generate_menu(self, menu):
+        action = self.env["ir.actions.act_window"].create(
+            self._generate_action_vals(menu)
+        )
+        self.menu_id = self.env["ir.ui.menu"].create(
+            self._generate_menu_vals(menu, action)
+        )
+
+
+class KpiDashboardKpi(models.Model):
```
legalsylvain: should not be better to name this model kpi.dashboard.line ?

kpi_dashboard/models/kpi_kpi.py
```diff
@@ -0,0 +1,121 @@
...
+        kwargs = ast.literal_eval(self.kwargs or "{}")
+        return getattr(obj, self.function)(*args, **kwargs)
+
+    def test_demo_number(self):
+        return {
+            "value": random.random() * 10000,
+            "previous": random.random() * 10000,
+        }
+
+    def test_demo_meter(self):
```
legalsylvain: I think demo data should not be in model files.

kpi_dashboard/wizards/kpi_dashboard_menu.py
```diff
@@ -0,0 +1,17 @@
...
+class KpiDashboardMenu(models.TransientModel):
+
+    _name = "kpi.dashboard.menu"
+    _description = "Create a Menu for a Dashboard"
+
+    dashboard_id = fields.Many2one("kpi.dashboard", required=True)
+    menu_id = fields.Many2one("ir.ui.menu")
+
+    @api.multi
+    def doit(self):
```
legalsylvain: ``doit`` is not a very common name for a button function. ``generate_menu`` or ``button_generate_menu`` could be better.

legalsylvain (changes requested):
Hi.
I did a quick code review.
some remarks inline. (specially the presence of demo data in python code).

Thanks for this great job ! 

A general question. Don't understand why the models of this module, is not based on the model defined in the ``kpi`` module. Don't you think ? not sure if it's easy to do.
But when both modules are installed, the configuration menu has many time the same name.

kpi_dashboard/models/kpi_kpi.py
legalsylvain: If you don't have any other work around, you can create an extra module kpi_dashboard_test for that contains that part of code.

andreampiovesana:
Put version [12.0] in PR name
Super work ! You are the Best!

kpi_dashboard/wizards/kpi_dashboard_menu.py
etobella: True. Changed

kpi_dashboard/models/kpi_dashboard.py
etobella: Changed for kpi.dashboard.item. I don't like line, as gridster is a matrix, but the original name was not the best one :wink:

kpi_dashboard/models/kpi_kpi.py
etobella: It is better to add a second module. True.

etobella:
@legalsylvain Your comments have been attended. About the `kpi` module. At the begining, that was my intention, but kpis from that module are only numbers, You cannot create dashboards or define configuration. At the end, I decided to create a new structure. Maybe there are better names, I have no problems with it :wink: 
On the other hand, I have already prepared two other modules related with this one:

- kpi_dashboard_bokeh: Allows to create the item inside the dashboard on python using bokeh library @pedrobaeza 
- kpi_dashboard_formula: Used to define the computation directly on the configuration, not using modules @JordiBForgeFlow

etobella:
@pedrobaeza Now you can add the dashboard to the user board :wink:
imagen

peek dashboard gif

legalsylvain:
Hi @etobella. Thanks for the changes you made, and thanks also for this great module ! 

Well regarding, the existing modules, it seems that this new module ``kpi_dashboard`` shares some feature with other modules : 

### ``web_dashboard_tile``
[V10](https://github.com/OCA/web/tree/10.0/web_dashboard_tile) and [PR V11](https://github.com/OCA/web/pull/1355) and [PR V12](https://github.com/OCA/web/pull/1428)
Contributors : @OSguard, @legalsylvain, @ivantodorovich, @steigendit, @zaoral

**commons features :**
- have a dashboard (or many) with squares with KPI displayed
- possibility to set custom colors (font and background)

**missing features in ``web_dashboard_tile``**
- not the possibility to call a python function
- basic display. (the ``kpi_dashboard`` is very nice for end user, using cool JS lib)
- data are computed each time the dashboard is displayed, that can be slow if request are complex, or if there are a lot of KPI to display

**Features not present in ``kpi_dashboard``**
- possibility to create new items of the dashboard, on the fly, by any users. domain / context will we stored, and reused after
- possibility to make operation on a column of any model (sum / max / min / average)
- possibility to display count quantity.

**different design**
- possibility to limit the access for a user, or for members of group(s) for a kpi. (in ``kpi_dashboard`` restriction are done per dashboard)

### ``kpi``
(same repository, previously in server-tools)
Contributors : @AdriaGForgeFlow, @joserprieto, @JordiBForgeFlow, @naousse, @dreispt, @max3903

don't know very good this module, but it seems that : 

**commons features :**
- call function of models

**missing features in ``kpi``**
- no dashboard
- no possibility to have many values. (only one single value)

**Features not present in ``kpi_dashboard``**
- possibility to make SQL query for local and on external database
 - possibility to handle threshold and range to mention if the value is bad or correct
- save history of old values.

My questions are : 
- regarding ``kpi`` module : is it relevant to merge models improving existing module ``kpi`` or not. (or maybe in a second time for a V13 / V14 migration). or do we accept to have two differents implementation of KPI ? 

-> I don't have any point of view. Maybe initial developpers and contributors of ``kpi`` modules could give their point of view ?

- regarding ``web_dashboard_tile``, Is it relevant to merge modules.

-> I think that ``kpi_dashboard`` is better, even if some features are missing. So, I think I'll will stop to work on the migration of ``web_dashboard_tile`` for V12, and make a PR to add missing feature in ``kpi_dashboard``. Certainly adding one key for the field ``computation_method``.

etobella:
Thanks @legalsylvain I didn't know `web_dashboard_tile`. Interesting module 😄 
I think it might be easy to migrate it to kpi_dashboard. So let me know if you need something in order to integrate it 😉

ivantodorovich (approved):
Nice! This looks incredible!!

Quick code review :+1: LGTM, functional test as well.

Just found a clumsy bug when reorganizing tiles, sometimes they overlap each other:

hidden

But overall this looks much much better than `web_dashboard_tile`.
As @legalsylvain said there are some missing features here, but it's much better to concentrate efforts in this new module, it looks much nicer

etobella:
Thanks @ivantodorovich I will try to remove the option, as the intial configuration is not overwritten by this, so it is confusing

etobella:
Drag and drop has been disabled @ivantodorovich

legalsylvain:
@etobella : My review is done. I just whant to have the Point of View of ``kpi`` module developpers, to avoid to have in the OCA two concurrent module, if other ways are possible.
Regarding ``web_dashboard_tile``, I'll make a PR in a second time. (I think I'll work on that topic in June of July). Not a blocking point to merge this PR, can be done later.

kind regards.

etobella:
Ok!!! I was just asking because github is blocking because you asked for changes :wink:

kpi_dashboard/static/src/scss/kpi_dashboard.scss
```diff
@@ -0,0 +1,112 @@
...
+    >.gridster {
+        margin: 0 auto;
+        >ul {
+            >li {
+                text-align: center;
+                list-style: none outside none;
+            }
+        }
+    }
+    .updated_at {
```
legalsylvain, line 14: This class is not displayed correctly for lines without kpi. (title) I think that it should be hidden for non kpi lines.

Tested on runbot : 

image

Do you have the same problem ? 

Trouble on Ubuntu 18.04 : Firefox 75.0 ; Chromium : 80.0.3987

etobella: Someone made it contain a kpi Number 01. If you set no kpi_id it works correctly. Did it happen on runbot or in local (maybe demo data is wrong, but I see it right)

legalsylvain: on runbot. maybe data was corrupted.

legalsylvain:
> Ok!!! I was just asking because github is blocking because you asked for changes wink

Well, I'm just asking reviews for ``kpi`` developpers.

andreampiovesana (approved):
good work

bealdav:
Really nice module, thanks a lot @etobella 

> - kpi_dashboard_bokeh: Allows to create the item inside the dashboard on python using bokeh library @pedrobaeza

FYI, I heard about Altair, as pythonic Declarative Visualization in Python
https://youtu.be/vTingdk_pVM?t=1121



Did you considered Altair ?

kpi_dashboard/security/ir.model.access.csv
```diff
@@ -0,0 +1,9 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_kpi_dashboard,access_kpi_dashboard,model_kpi_dashboard,base.group_user,1,0,0,0
+access_kpi_dashboard_kpi,access_kpi_dashboard_kpi,model_kpi_dashboard_item,base.group_user,1,0,0,0
+access_kpi_kpi,access_kpi_kpi,model_kpi_kpi,base.group_user,1,0,0,0
+access_kpi_kpi_action,access_kpi_kpi_action,model_kpi_kpi_action,base.group_user,1,0,0,0
+manage_kpi_dashboard,manage_kpi_dashboard,model_kpi_dashboard,group_kpi_dashboard_manager,1,1,1,0
+manage_kpi_dashboard_kpi,manage_kpi_dashboard_kpi,model_kpi_dashboard_item,group_kpi_dashboard_manager,1,1,1,1
+manage_kpi_kpi,manage_kpi_kpi,model_kpi_kpi,group_kpi_dashboard_manager,1,1,1,0
```
AaronHForgeFlow: IMHO we should be able to delete KPIs

etobella: No problem for me

etobella:
@bealdav I proposed bokeh because there is a module for bokeh on OCA. It is no problem for me to review it. At the end, if it can export the graph to a SVG, PNG or something that can be displayed on HTML, it will be fine.

etobella:
@bealdav I have been testing altair and it can be configurated easily into dashboard, following a simple configuration like bokeh, I can propose a PR for each method as soon as this module is in OCA :wink:

etobella:
If someone is interested, the modules I talked can be found here:
https://github.com/tegin/bi/tree/12.0-add-kpi_dashboard
I started on a new repo at the begining, but all will be ported to reporting-engine as soon as possible.

JordiBForgeFlow:
@legalsylvain I think that it's not worth to try to merge all modules together, but let them coexist until one wins. It's too much effort to try to satisfy everyone into a single module, and then the end result is often a very complicated module. So +1 to continue with this one as @etobella proposed.

etobella:
@legalsylvain The problem I see is that kpi already exists on 12.0. So maybe we could try to create a migration script in order to simplify the process but this could be hard if it has some data interface or something similar.

legalsylvain (approved):

legalsylvain:
Hi all. My question was just : is it possible to converge both designs ? but it doesn't seem easy.

Let's go having two kpi modules, and as @JordiBForgeFlow said, see which module wins.
Maybe it could be great to mention in the readme of each module the existence of the other module. Don-t you think ?

Thanks for this great work.

bealdav:
Maybe it can be merged now ? Thanks

LoisRForgeFlow:
I will do the honors :grinning: 

/ocabot merge patch

PR #391, opened 2020-04-23
[12.0][FIX] bi_view_editor: window functions without ORDER BY clause do not have reliable behavior

### LINKED ISSUE #388: [12.0] bi_view_editor: row_number over () does not guarantee ordering
Hi,
given the following view created using module `bi_view_editor`:
```
CREATE OR REPLACE VIEW public.x_bve_commissioni
AS SELECT row_number() OVER ()::integer AS id,
    t1.id AS x_bve_t1_id, -- manually added for analysis
    t1.agent_id AS x_bve_t1_agent_id,
    t1.partner_id AS x_bve_t1_partner_id,
    t1.date_invoice AS x_bve_t1_date_invoice,
    t1.product_id AS x_bve_t1_product_id,
    t1.price_unit AS x_bve_t1_price_unit,
    t1.commission_id AS x_bve_t1_commission_id,
    t1.price_subtotal_signed AS x_bve_t1_price_subtotal_signed,
    t1.amount AS x_bve_t1_amount
   FROM sale_commission_analysis_report t1; -- defined in https://github.com/OCA/commission/blob/befb9470a352edc9a2946bbf1c3b5a57e02badeb/sale_commission/report/sale_commission_analysis_report.py#L7
```
I am currently encountering the following strange behavior:
```
prod_db=# select * from x_bve_commissioni where id = 5130;
  id  | x_bve_t1_id | x_bve_t1_agent_id | x_bve_t1_partner_id | x_bve_t1_date_invoice | x_bve_t1_product_id |  x_bve_t1_price_unit   | x_bve_t1_commission_id | x_bve_t1_price_subtotal_signed | x_bve_t1_amount 
------+-------------+-------------------+---------------------+-----------------------+---------------------+------------------------+------------------------+--------------------------------+-----------------
 5130 |       36303 |             14343 |               19833 | 2020-03-16            |                   2 | 0.00000000000000000000 |                     11 |                           0.00 |            0.00
(1 row)

prod_db=# select id, x_bve_t1_id,  x_bve_t1_date_invoice from x_bve_commissioni where id = 5130;
  id  | x_bve_t1_id | x_bve_t1_date_invoice 
------+-------------+-----------------------
 5130 |       19991 | 2020-01-27
(1 row)
```
I expect the column `x_bve_t1_id` to be the same in both query results.

I'm no PostgreSQL expert but I did some research and the solution I came up with is adding an `order by t1.id` clause in the `OVER` of  column `row_number() OVER ()::integer AS id`.
I've seen that the column's creation is hardcoded in https://github.com/OCA/reporting-engine/blob/124d1db26b3f13f3b4669896699a4b9f3e16eb49/bi_view_editor/models/bve_view.py#L292 so I don't really know how this could be fixed.

Any ideas?
Thanks!

SimoRubi:
Fix #388.
An order by clause is highly recommended when using window functions:
> Without ORDER BY, rows are processed in an unspecified order.

(from https://www.postgresql.org/docs/10/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS)

This modification allows a (technical) user to specify such condition and others, like PARTITION BY.

### DIFF
--- a/bi_view_editor/models/bve_view.py
+++ b/bi_view_editor/models/bve_view.py
@@ -98,6 +98,19 @@ class BveView(models.Model):
         compute='_compute_users',
         store=True)
     query = fields.Text(compute='_compute_sql_query')
+    over_condition = fields.Text(
+        states={
+            'draft': [
+                ('readonly', False),
+            ],
+        },
+        readonly=True,
+        help="Condition to be inserted in the OVER part "
+             "of the ID's row_number function.\n"
+             "For instance, 'ORDER BY t1.id' would create "
+             "IDs ordered in the same way as t1's IDs; otherwise "
+             "IDs are assigned with no specific order.",
+    )
     er_diagram_image = fields.Binary(compute='_compute_er_diagram_image')
 
     _sql_constraints = [
@@ -285,11 +298,12 @@ class BveView(models.Model):
         self.env.cr.execute('CREATE or REPLACE VIEW %s as (%s)', (
             AsIs(view_name), AsIs(query), ))
 
-    @api.depends('line_ids', 'state')
+    @api.depends('line_ids', 'state', 'over_condition')
     def _compute_sql_query(self):
         for bve_view in self:
             tables_map = {}
-            select_str = '\n CAST(row_number() OVER () as integer) AS id'
+            select_str = '\n CAST(row_number() OVER ({}) as integer) AS id' \
+                .format(bve_view.over_condition or '')
             for line in bve_view.field_ids:
                 table = line.table_alias
                 select = line.field_id.name
--- a/bi_view_editor/models/bve_view.py
+++ b/bi_view_editor/models/bve_view.py
@@ -295,8 +295,16 @@ class BveView(models.Model):
         self._cr.execute('DROP TABLE IF EXISTS %s', (AsIs(view_name), ))
 
         # create postgres view
-        self.env.cr.execute('CREATE or REPLACE VIEW %s as (%s)', (
-            AsIs(view_name), AsIs(query), ))
+        try:
+            with self.env.cr.savepoint():
+                self.env.cr.execute('CREATE or REPLACE VIEW %s as (%s)', (
+                    AsIs(view_name), AsIs(query), ))
+        except Exception as e:
+            raise UserError(
+                _("Error creating the view '{query}':\n{error}")
+                .format(
+                    query=query,
+                    error=e))
 
     @api.depends('line_ids', 'state', 'over_condition')
     def _compute_sql_query(self):
--- a/bi_view_editor/readme/ROADMAP.rst
+++ b/bi_view_editor/readme/ROADMAP.rst
@@ -5,4 +5,3 @@
 * Data the user has no access to (e.g. in a multi company situation) can be
   viewed by making a view. Would be nice if models available to select when
   creating a view are limited to the ones that have intersecting groups.
-* Raise a warning in case the SQL query is not correct.
--- a/bi_view_editor/readme/ROADMAP.rst
+++ b/bi_view_editor/readme/ROADMAP.rst
@@ -5,3 +5,4 @@
 * Data the user has no access to (e.g. in a multi company situation) can be
   viewed by making a view. Would be nice if models available to select when
   creating a view are limited to the ones that have intersecting groups.
+* Raise a warning in case the SQL query is not correct.
--- a/bi_view_editor/readme/USAGE.rst
+++ b/bi_view_editor/readme/USAGE.rst
@@ -21,3 +21,6 @@ To access the created BI View with a dedicated menu:
 A more advanced UI is also available under the "Details" tab. It provides extra
 possibilities for more advanced users, like to use LEFT JOIN instead of the
 default INNER JOIN.
+
+It also possible to improve the IDs generation for new views by adding an `Over Condition` in the "SQL" tab, see https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS for further details.
+For instance, an ORDER BY clause helps preventing unreliable behavior when filtering the generated views.
--- a/bi_view_editor/tests/test_bi_view.py
+++ b/bi_view_editor/tests/test_bi_view.py
@@ -5,6 +5,7 @@ import json
 
 import odoo
 from odoo.tests.common import TransactionCase
+from odoo.tools import mute_logger
 from odoo.exceptions import UserError, ValidationError
 from ..hooks import post_load, uninstall_hook
 
@@ -407,3 +408,24 @@ class TestBiViewEditor(TransactionCase):
         bi_view1 = self.env['bve.view'].create(vals)
         bi_view1.action_create()
         self.assertEqual(len(bi_view1.line_ids), 4)
+
+    @mute_logger('odoo.sql_db')
+    def test_20_broken_view(self):
+        """
+        Create a broken query, a nice UserError should be raised.
+        odoo.sql_db logger is muted to avoid the
+        ERROR: bad_query line in the logs.
+        """
+        vals = self.bi_view1_vals
+        vals.update({
+            'name': 'Test View broken',
+            'over_condition': 'bad SQL code',
+        })
+        bi_view = self.env['bve.view'].create(vals)
+        with self.assertRaises(UserError) as ue:
+            bi_view.action_create()
+
+        self.assertEqual(bi_view.state, 'draft')
+        self.assertIn(bi_view.over_condition, str(ue.exception))
+        # remove view
+        bi_view.unlink()
--- a/bi_view_editor/views/bve_view.xml
+++ b/bi_view_editor/views/bve_view.xml
@@ -85,8 +85,9 @@
                             </group>
                         </page>
                         <page string="SQL" groups="base.group_no_one">
+                            <field name="query"/>
                             <group>
-                                <field name="query" nolabel="1" />
+                                <field name="over_condition"/>
                             </group>
                         </page>
                         <page string="Security">


legalsylvain (approved):
make sense. 
Thanks.

bi_view_editor/models/bve_view.py
```diff
@@ -98,6 +98,11 @@ def _inverse_serialized_data(self):
         compute='_compute_users',
         store=True)
     query = fields.Text(compute='_compute_sql_query')
+    over_condition = fields.Text(
+        string="Over condition",
```
astirpe: You can remove this line `string="Over condition",` since Odoo creates the field name automatically.

bi_view_editor/models/bve_view.py
```diff
@@ -98,6 +98,11 @@ def _inverse_serialized_data(self):
...
+        help="Condition to be inserted in the OVER part "
+             "of the ID's row_number function.",
```
astirpe: Could you add extra information regarding how to use this functionality? Or maybe some example of use?

astirpe (changes requested):
@SimoRubi Thank you for this!

bi_view_editor/models/bve_view.py
SimoRubi: Thanks for the review!
Yes I know that the `string` is automatically filled but I usually prefer to keep it explicit.

bi_view_editor/models/bve_view.py
SimoRubi: Added an example, generic extra information (PARTITION BY, RANGE, ...) would be way too long for an `help` attribute I think.

bi_view_editor/models/bve_view.py
astirpe: As per https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#fields, the string should be removed. It's also useful to not specify any string in some cases. For example, in case someone wants to manually set a different string directly in the field definition, this won't conflict when the module is updated.

bi_view_editor/models/bve_view.py
astirpe: Thanks. Any useful extra explanation could be added in the readme.

bi_view_editor/models/bve_view.py
SimoRubi: Removed

astirpe:
In case the user writes an invalid over condition, the entire query would fail. In this case can be raised any user friendly warning?

bi_view_editor/models/bve_view.py
SimoRubi: Added

SimoRubi:
This is a good idea for an improvement to the module, but I think it is beyond this PR's purpose.

astirpe:
Ok if you please add "Raise a warning in case the user writes an invalid over condition." in the `ROADMAP.rst` then I will merge immediately.

SimoRubi:
hah, that's blackmailing! :wink:
Added, and broadened the scope of the warning to "in case the SQL query is not correct.": let's not point fingers to the new arrived field :smile:

astirpe (approved):

astirpe:
I just noticed that I cannot merge since I'm not a maintainer of the module :thinking:

bi_view_editor/models/bve_view.py
```diff
@@ -285,11 +292,12 @@ def _create_sql_view(self):
         self.env.cr.execute('CREATE or REPLACE VIEW %s as (%s)', (
```
astirpe: What do you think about:


```python
from psycopg2.errors import Error, UndefinedObject

        # ...

        try:
            self.env.cr.execute(
                "CREATE or REPLACE VIEW %s as (%s)", (AsIs(view_name), AsIs(query),)
            )
        except UndefinedObject:
            raise ValidationError(_("The generated query is invalid. Is the OVER condition correct?"))
        except Error as e:
            raise ValidationError(_("The generated query is invalid:\n%s") % e.pgerror)
```

SimoRubi: Ahah still blaming my dear `over condition`? I know it is the weak spot because it is the only free text input in the query definition, but maybe it would be better to print also the original exception in both cases (maybe only the `except Error` should suffice).

Moreover, I rapidly checked your solution and `psycopg2.errors` is introduced in 2.8 (ref. https://www.psycopg.org/docs/errors.html) but minimal requirement is `2.7.3.1` (ref. https://github.com/odoo/odoo/blob/012e4a3eb11feabe5634ab806ec5dd2012ffda00/requirements.txt#L28)..

Anyway I still think these changes deserve a separate issue/PR

astirpe:
```python
        try:
            self.env.cr.execute(
                "CREATE or REPLACE VIEW %s as (%s)", (AsIs(view_name), AsIs(query),)
            )
        except psycopg2.Error as e:
            raise ValidationError(_("The generated query is invalid:\n%s") % e.pgerror)

```

SimoRubi: Haven't tried that out, but looks good

SimoneVagile (approved):

SimoRubi:
@OCA/reporting-engine-maintainers can this be merged? 
I'd like to implement the changes suggested in https://github.com/OCA/reporting-engine/pull/391#pullrequestreview-406343248 in a separate PR, after this one is merged to avoid conflicts.

astirpe:
@SimoRubi  the PRs approvals in OCA are usually long and debated. This PR already has 3 approvals. Would you consider to include that changes here? It would save a lot of time to all of us.

SimoRubi:
> @SimoRubi the PRs approvals in OCA are usually long and debated. This PR already has 3 approvals. Would you consider to include that changes here? It would save a lot of time to all of us.

The approvals are for the changes that are in the PR right now; if I added new changes, I should also re-request the reviews.

I have created https://github.com/OCA/reporting-engine/pull/397 for the new changes you requested

astirpe:
What I proposed was my review for this PR. Now we have two PRs for the same issue, and this one has 3 approvals, the other has 0 ...

SimoRubi:
That is the point, it's two separate issues:

1 (this PR) allows the user to specify something in the `OVER` condition
2 (the other PR) gives a better error message when the query is incorrect

Now, this PR as you said is ready for merge because the feature it adds has already been reviewed.

The other one might be independent from this one but the fastest way that I've found to break the query is using `over_condition` field, so that PR actually depends on this one.

astirpe (changes requested):
I'm reviewing this PR.

Just tested in Runbot.

Steps to reproduce:

1. Enter an invalid `over_condition`.

Result:


```
Error:
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 656, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 314, in _handle_exception
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 698, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 346, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 98, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
  File "/.repo_requirements/odoo/odoo/http.py", line 941, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-12.0/addons/web/controllers/main.py", line 966, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/OCB-12.0/addons/web/controllers/main.py", line 954, in _call_kw
  File "/.repo_requirements/odoo/odoo/api.py", line 759, in call_kw
    return _call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 746, in _call_kw_multi
  File "/home/odoo/build/OCA/reporting-engine/bi_view_editor/models/bve_view.py", line 381, in action_create
    self._create_sql_view()
  File "/home/odoo/build/OCA/reporting-engine/bi_view_editor/models/bve_view.py", line 293, in _create_sql_view
    AsIs(view_name), AsIs(query), ))
  File "/.repo_requirements/odoo/odoo/sql_db.py", line 148, in wrapper
    return f(self, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/sql_db.py", line 225, in execute
    res = self._obj.execute(query, params)
psycopg2.ProgrammingError: window "test" does not exist
LINE 1: ...EW x_bve_test as (SELECT   CAST(row_number() OVER (test) as ...
                                                             ^


```

SimoRubi:
Improving the error message is a general improvement for the module, that's why I added
```
* Raise a warning in case the SQL query is not correct.
```
in the roadmap.
Anyway that does look like the correct behavior: you write an invalid query and you get an error, the added changes only give a 'better' error.

In the end, since here we are getting into a loop I'll merge the changes from https://github.com/OCA/reporting-engine/pull/397 and request again all the reviews

astirpe (changes requested):
Thanks!

`over_condition` should be readonly when not in draft, otherwise it's possible to modify it while the bi view is already created. Just tested in Runbot.

astirpe (approved):
@legalsylvain would you merge this PR?

SimoRubi:
@OCA/reporting-engine-maintainers can this be merged? Thanks!

SimoRubi:
@OCA/reporting-engine-maintainers can this be merged?
2 revisions should be more than enough since this is a Beta module (ref. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst#2alpha-and-beta-modules).

SimoRubi:
@OCA/reporting-engine-maintainers can someone please take a look?
@legalsylvain there is a little change (https://github.com/OCA/reporting-engine/pull/397) from when you reviewed, can you update your review?

Thanks!

PR #393, opened 2020-04-28
[12.0][IMP] kpi_dashboard: Allow to use formula

etobella:
Kpi Dashboard Formula
=====================

Define KPIs for dashboard using python formulas directly.
### Usage

* Access `Dashboards > Configuration > KPI Dashboards > Configure KPI`
* Create a new KPI and choose as computation method `Code`
* Define the code directly. You can use:

   * `self` and `model` as the kpi element
   * You must return result as a dictionary that will be stored as data


legalsylvain (commented):
Thanks !
Lgtm. Code review, no test. 
However, i'm asking why this feature is not in the main module.
It's a,very generic feature, and more modular than the default one.

pedrobaeza:
Yeah, indeed it's not adding any extra dependency in Odoo modules nor libraries, and the base code is not a burden, as it's very few, so I would include it on the main module.

etobella:
Ok, I will include it then :wink:

etobella:
@legalsylvain @pedrobaeza Now it is integrated directly on kpi_dashboard. Easier now.

etobella:
Just a question, with this change I can probably delete the testing module. What do you think?

pedrobaeza:
If possible, of course I prefer to remove any testing module.

kpi_dashboard/readme/CONFIGURE.rst
```diff
@@ -8,6 +8,9 @@ Configure KPIs
    #. Meter: result must contain `value`, `min` and `max`
    #. Graph: result must contain a list on `graphs` containing `values`, `title` and `key`
 
+#. In order to compute the KPI you can use a predefined function from a model or
+   use the code to directly compute it.
```
pedrobaeza, line 12: Isn't this missing the explanation of what to return?

pedrobaeza (approved):

jarroyomorales (approved):

kpi_dashboard/readme/CONFIGURE.rst
etobella: Ups, Sorry, comments attended

kpi_dashboard/models/kpi_kpi.py
```diff
@@ -84,6 +87,30 @@ def write(self, vals):
...
 
+    def _get_code_input_dict(self):
+        return {
+            "self": self,
+            "model": self.browse(),
+        }
+
+    def _compute_value_code(self):
+        new_cr = Registry(self.env.cr.dbname).cursor()
+        try:
```
etobella: This tries to disallow something like self.unlink()

We are trying to find a better option...

pedrobaeza: Put it inside a savepoint and do rollback after:
```python
self.env.cr.execute("savepoint kpi_formula_%" % self.id)
safe_eval(...)
self.env.cr.execute("rollback to kpi_formula_%" % self.id)
```

etobella: Thanks @pedrobaeza your solution is good, however the savepoint is not perfect. I mean, if the user puts 
```
self.unlink()
self.env.cr.commit()
```
the unlink will be performed. However, the same happened with my solution and this is already cleaner, so I will keep it if nothing better can be found.

pedrobaeza: Why don't you simply look for `.commit()` in a constraint and forbids that string?

etobella: That will probably be the best combined with the savepoint

kpi_dashboard/models/kpi_kpi.py
```diff
@@ -84,6 +88,27 @@ def write(self, vals):
...
         return super().write(vals)
 
+    def _get_code_input_dict(self):
+        return {
+            "self": self,
+            "model": self.browse(),
+        }
+
+    def _forbidden_code(self):
+        return "(commit|rollback)"
```
jarroyomorales: Shouldnt you also forbid `execute` ?

bealdav (approved):

bealdav:
Hi @legalsylvain your wishes are anwsered. Thanks

kpi_dashboard/models/kpi_kpi.py
legalsylvain: Well, in that case, it will avoid to have a compute value code that makes an SQL query, that can be interesting.

etobella: Good Point, I have added execute to the forbidden words. I also added `getattr`, as you could do something like:
```
getattr(self.env.cr, 'exec'+'ute')('com'+'mit')
```

pedrobaeza: But it's the contrary. @legalsylvain is saying that this prevent manual SQL queries, and the only problem I see is to commit that, so no need to forbid it.

kpi_dashboard/models/kpi_kpi.py
```diff
@@ -84,6 +88,27 @@ def write(self, vals):
...
+            "model": self.browse(),
+        }
+
+    def _forbidden_code(self):
+        return "(commit|rollback|getattr|execute)"
+
+    def _compute_value_code(self):
+        if re.search(self._forbidden_code(), (self.code or "").lower()):
+            raise ValidationError(_(
+                "The code cannot contain rollback or commit inside."
```
bealdav: Maybe 
```suggestion
                "The code cannot contain these terms '%s' inside." % self._forbidden_code()[1:-1].replace("|", ", ")
```

etobella: I proposed another option. What do you think @bealdav ?

LoisRForgeFlow:
Then, do you think guys this is good to merge? We shouldn't stop momentum ;)

bealdav:
Yes I think it's good to merge. It doesn't prevent to improve it after

pedrobaeza:
Last change is not going to be in the merge. Should we stop the merge process for including it?

etobella:
I was adding myself as maintainer

LoisRForgeFlow:
cancelled, let's restart:

/ocabot merge minor

pedrobaeza:
Other thing: this topic about avoiding to abuse the `safe_eval` thing is in several modules (base_tier_validation_formula - or similar name -, etc). Maybe we should extract this logic in a base module that join all the needed techniques and make use of it. It can be named `base_super_safe_eval` :stuck_out_tongue_closed_eyes:

etobella:
Maybe it is interesting to create a module on server-tools or somewhere else. If we create the module I will integrate it here. What do you think @LoisRForgeFlow ?

LoisRForgeFlow:
Sounds good, I like security improvements :nerd_face:

PR #394, opened 2020-04-28
[12.0][ADD] kpi_dashboard_bokeh

etobella:
Kpi Dashboard Bokeh
===================

This module adds new kinds of dashboards on a specific new type of view.

#### Configuration

* Access `Dashboards > Configuration > KPI Dashboards > Configure KPI`
* Create a new KPI specifying with widget type `bokeh`

In order to define the value, you can must define a function like::

    def test_demo_bokeh(self):
        p = figure(width=1000, height=1000, sizing_mode="scale_both")
        # import that as `from bokeh.plotting import figure`
        p.line([0, 1, 2], [1, 10, random.random() * 10], line_width=5)
        # (...)
        # fill the record field with both markup and the script of a chart.
        script, div = components(p)
        return {"bokeh": "%s%s" % (div, script)}


etobella:
imagen

etobella:
By the way, once #393 is merged, I will add bokeh to the inputs you can use, so you can define the plot with code. With a test and more information to the readme, obviously :wink:

tonygalmiche (commented):
@etobella ,
I don't know bokeh
I failed to operate this function
What should you import?

```
def test_demo_bokeh(self):
    p = figure(width=1000, height=1000, sizing_mode="scale_both")
    # import that as `from bokeh.plotting import figure`
    p.line([0, 1, 2], [1, 10, random.random() * 10], line_width=5)
    # (...)
    # fill the record field with both markup and the script of a chart.
    script, div = components(p)
    return {"bokeh": "%s%s" % (div, script)}
```

Thank you in advance

tonygalmiche:
@etobella,
I found, we must add that
```
from bokeh.plotting import figure, output_file, show
from bokeh.embed import components
```
But now, I got this error when displaying the dashboard
```
Error: property Grid.axis wasn't declared
http://buster:8069/web_widget_bokeh_chart/static/src/lib/bokeh/bokeh-1.1.0.js:1563
```

etobella:
I have improved the readme and I also added the option to configure the graph directly on code. Also, a new theme has been added with no background that can be added using the simple_component function

jarroyomorales (approved):
Code Review and tested on Runbot and LGTM! :+1:

tonygalmiche:
I installed the module on my computer
when I want to display the Dashboard proposed by default, I have this error : 
`Error: QWeb2: Template 'FieldColor' not found`
Thanks in advance for the help

etobella:
This is weird, are you using the proper `web_widget_color` addon from OCA/web?

tonygalmiche:
I just restarted my odoo and now it works. Probably a cache problem.

But the marking of the axes (x and y) is not legible (too small)

image

thank you in advance

etobella:
This problem is related with how bokeh prints the data... At the end, bokeh is fixed in how to manage the plot :unamused: I tried to make the image with js but at the end I had no option... :cry: 
Right now I proposed two options, bokeh and altair, I think altair is better, but bokeh is already present on OCA. Maybe we should need to keep both knowing the limitations of bokeh

pedrobaeza:
I would go with https://frappe.io/charts (with a glue module). The advantage of both Altair and Bokeh is the gateway to declare in Python the result, but at the end, it has these disadvantages.

etobella:
Altair is more flexible on this point as the declaration of the graph is managed by vega on JS. Not the same with Bokeh. We might be able to see integration with frappe on the Spanish OCA Days

tonygalmiche (approved):
ok, I understand

bealdav:
2 500 projects use altair
661 use frappe/charts

according github

with data scientists behind the scene for altair
https://www.youtube.com/watch?v=OquQ6M7yoGU

About serious data, it's seems python is a leader

Altair has the reputation to be more pythonic than bokeh and is more recent.

Future'll learn a lot

pedrobaeza:
Data scientist is a field totally different from BI dashboards...

bealdav:
dashboard is dashboard whatever the subject: business, science, home data

dashboard aggregate kpi sharing filters and parameters

like https://redash.io do

@etobella modules allows us to hope to encroach features in these dashboards tools

But maybe not so easy with bokeh

pedrobaeza:
I think you are rambling a bit. I have only proposed an alternative that is very well built for JS and for showing BI graphs, which has nothing to do with scientific plots, due to the problems with current alternative. All the already done work is of course amazing and won't be dismissed by what I'm proposing.

etobella:
Well both systems should be compatible

LoisRForgeFlow (approved):
LGTM, nice work!

PR #397, opened 2020-05-07
[12.0] bi_view_editor: manage incorrect query

SimoRubi:
Better error message if the query is not correct.

Depends on https://github.com/OCA/reporting-engine/pull/391


SimoRubi:
Merged into #391

PR #400, opened 2020-05-11
12.0 IMP altair: complete demo

bealdav:
This PR is to test altair on more use cases.

I have a js bug when I create these additional demo data.

I wait for runbot


etobella:
Which bug?

bealdav:
No bug in fact on runbot. :-)

I encountered a bug on full installed modules and this demo, then OK for that.

Sorry for the noise

PR #414, opened 2020-06-25
[11.0][FIX] Fix error when downloading xlsx reports

sacostaylara:
The frontend was expecting a cookie with value 'fileToken' to be able to complete the ajax call, this lead to error in frontend since the ajax call was never completed.

Probably due to this error is that the method framework.unblockUI() was put outside of the ajax call at qwebactionmanager.js, so this made that the UI was unblocked immediately after clicking on download report and not when report was actually downloaded, so we wouldn't see the loading spinner while the report was being generated. This was fixed too.

### DIFF
--- a/report_xlsx/controllers/main.py
+++ b/report_xlsx/controllers/main.py
@@ -45,7 +45,7 @@ class ReportController(report.ReportController):
                     content_disposition(report_name + '.xlsx')
                 )
             ]
-            return request.make_response(xlsx, headers=xlsxhttpheaders)
+            return request.make_response(xlsx, headers=xlsxhttpheaders, cookies={'fileToken': request.params.get('token')})
         return super(ReportController, self).report_routes(
             reportname, docids, converter, **data
         )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I have `account_financial_report` module working with Odoo v11 enterprise without problems, so I doubt the problem is due to enterprise. It seems more something about your deployment or specific tweaks.

sacostaylara:
> I have `account_financial_report` module working with Odoo v11 enterprise without problems, so I doubt the problem is due to enterprise. It seems more something about your deployment or specific tweaks.

I seriously doubt that this module works in v11 without problems. Method get_file at web/static/src/js/core/ajax.js requires a cookie 'fileToken' to be set for being able to complete successfully, but the module report_xlsx never sets that cookie, so thats why the ajax call is never completed and generates the error.

I assume that this error is the reason why the method framework.unblockUI is invoked outside of the ajax call at file report_xlsx/static/src/js/report/qwebactionmanager.js, instead of doing it on the complete function of the ajax call as it should be done to make the loading work correctly. When I click on EXPORT XLSX the UI is unblocked right away, even if the report hasn't finished downloading yet, this is not ideal.

So for me this module is definitely bugged, and I don't see how it could work correctly on v11 without setting the cookie 'fileToken'.

pedrobaeza:
OK, let's see what others think, but I assure you that it was working in all my installations in both community and enterprise. Now we have barely migrated all of them to v12, but 3 remaining ones, 2 with CE and 1 with EE, are still working with latest code.

Issue #415, opened 2020-07-06
[RFC] kpi_dashboard vs web_dasboard_tile : take into account context


legalsylvain:
Hi all,

In previous version of Odoo, I was using ``web_dashboard_tile`` that generates tiles with computed values. Following [this discussion](https://github.com/OCA/reporting-engine/pull/387#issuecomment-614242900), we decided to abandon ``web_dashboard_tile``, and focus the work on ``kpi_dashboard``, that has more features.  

However, ``web_dashboard_tile`` compute values on the fly, and ``kpi_dashboard`` stores values. So ``kpi_dashboard`` is faster, but doesn't take into account the context. In a multi company context, if a kpi is "Sum of the opened invoices to pay", ``kpi_dashboard`` will display a unique global values to all users. It is expected to display a value depending on the context of the user. (display only the opened invoices of the current company).

How do see I could improve the ``kpi_dashboard`` module to take into account the context ? 

I was thinking to add a boolean ``compute_on_the_fly``, unchecked by default, on the ``kpi.kpi`` model. 
- if unchecked : existing behaviour.
- if checked : no cron is generated, and the value is computed on the fly. (slower, no bus, but take into account context).

What do you think ? 

Thanks !

CC : original author of ``kpi_dashboard`` : @etobella 
reviewers of the ``kpi_dashbaord`` module : @AaronHForgeFlow, @ivantodorovich, @andreampiovesana, @LoisRForgeFlow, @AdriaGForgeFlow. 
reviewers / coauthors of the ``web_dashboard_tile`` module : @steigendit, @zaoral, @OSguard, @ivantodorovich, @pedrobaeza

ivantodorovich:
I haven't dug into that module yet, but maybe you could create one kpi per
company:
"Sum of the opened invoices to pay (company A)"
"Sum of the opened invoices to pay (company B)"
etc..

That way you don't need to sacrifice performance

If you go for this approach, I think it would make sense to add an optional
company_id field, with a corresponding ir.rule, in kpi.kpi.

etobella:
Well, I think it is good to use this kind of new behavior, even in the base module, so it is good for me.

pedrobaeza:
The problem I see is that there's no historic storage of previous values, and proper visualization of them. For a kpi you want to see its trend, not only previous and current value.

etobella:
Well, trends can be stored, and there is a way to show trends. Also, in my experience, it has no sense to store the data. Let me show an example:

[...]

PR #416, opened 2020-07-07
[IMP] kpi_dashboard: Set history & Add compute on the fly & Show value on kpi form

etobella:
following the comments of #415  I made this PR. Also, the value of the kpi can be shown directly on the KPI form and history using the show value button

@pedrobaeza @legalsylvain

### DIFF
--- a/kpi_dashboard/models/kpi_dashboard.py
+++ b/kpi_dashboard/models/kpi_dashboard.py
@@ -167,10 +167,18 @@ class KpiDashboardItem(models.Model):
                     "kpi_id": self.kpi_id.id,
                     "suffix": self.kpi_id.suffix or "",
                     "prefix": self.kpi_id.prefix or "",
-                    "value": self.kpi_id.value,
-                    "value_last_update": self.kpi_id.value_last_update,
                 }
             )
+            if self.kpi_id.compute_on_fly:
+                vals.update({
+                    "value": self.kpi_id._compute_value(),
+                    "value_last_update": fields.Datetime.now(),
+                })
+            else:
+                vals.update({
+                    "value": self.kpi_id.value,
+                    "value_last_update": self.kpi_id.value_last_update,
+                })
             if self.kpi_id.action_ids:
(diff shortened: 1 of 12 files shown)


pedrobaeza (commented):
I miss a date/datetime reference in the historization, and the possibility to define the scope of the KPI (hourly, daily, weekly, etc), for not storing always a new entry if still in the same range (or store the last one only). This is critical as your cron will be executed daily, but a weekly KPI will have 7 historization records!

etobella:
Good point!!! I did not think about it, but it has a lot of sense :wink: WOI

etobella:
Comments attended, now we can define a interval scope :smile:

legalsylvain:
Hi @etobella. tested. works like a charm thanks ! 
I made a little PR to improve the new feature. https://github.com/tegin/reporting-engine/pull/1

etobella:
Good for me!

jarroyomorales (approved):
Functional test LGTM! :+1:

legalsylvain (approved):

etobella:
@legalsylvain I have some improvements that can be really interesting, like allowing to refresh the compute_on_fly using an interval, Do you prefer this here or on a different PR?

etobella:
https://github.com/tegin/reporting-engine/tree/12.0-imp-kpi-onfly

legalsylvain:
Hi @etobella, 

First thanks for there improvements.

I think it can be integrated in this PR, as it is improving the new compute on the fly feature.

Good idea to add the context (https://github.com/tegin/reporting-engine/commit/aff958e21b4534ca36edf1e4b5d05b1a1e1ddbe2) do you think it could be possible to add also a domain ? 
This is a current regression, regarding the ``web_dashboard_tile`` features. (https://github.com/OCA/web/blob/10.0/web_dashboard_tile/models/tile_tile.py#L165)

kind regards.

etobella:
I will check the domain, I am not sure if we can do something specific :wink: but I will try

PR #417, opened 2020-07-08
[MIG] base_comment_template: Migration to 13.0 - moved from account-invoice-reporting

feketemihai:
migrated the module and made it more general:

added model_id relation to ir.model
added domain field to filter the records, so we can have one template for out_invoice and other template for in_invoice
added partner_id field related to res.partner
added default boolean field
added method to base model to get the comment template based on record model, position, partner and domain
The method will aso process the template, so we can use variables in the comments just like mako email template variables: ${object.}, ${user.}


pedrobaeza:
I'm afraid you haven't preserved commit history.

feketemihai:
@pedrobaeza right, i will redo it

feketemihai:
@pedrobaeza Done.

pedrobaeza:
Great, thanks!

AaronHForgeFlow (approved):
:+1: LGTM!

base_comment_template/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,17 @@
+* Xavier Jimenez <xavier.jimenez@qubiq.es>
+* Nicolas Bessi <nicolas.bessi@camptocamp.com>
+* Yannick Vaucher <yannick.vaucher@camptocamp.com>
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Simone Rubino <simone.rubino@agilebg.com>
```
emagdalenaC2i, line 5: This line is repeated

base_comment_template/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,11 @@
+Go to *Settings > Technical > Reporting > Comment Templates* and start designing you comment templates.
```
emagdalenaC2i, line 1: Add a line to first activate de Developer mode

base_comment_template/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,11 @@
...
+
+The template are general, and can be attached to any Model and based on some domain defined in the template.
```
emagdalenaC2i: Why select any model?
I believe there are a lot of models that can't use this templates

base_comment_template/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,16 @@
+Add a new model to define templates of comments to print on
+documents.
+
+Two positions are available for the comments:
+* above document lines
+* below document lines
+
+The template are general, and can be attached to any Model and based on some domain defined in the template.
```
emagdalenaC2i, line 9: This is also in the CONFIGURE.rst file

emagdalenaC2i:
Base comment template

base_comment_template/readme/CONTRIBUTORS.rst
feketemihai: Done

base_comment_template/readme/CONFIGURE.rst
feketemihai: Done

base_comment_template/readme/CONFIGURE.rst
feketemihai: I added fields view get to filter models that have reports.

feketemihai:
@emagdalenaC2i Please recheck.

base_comment_template/readme/DESCRIPTION.rst
feketemihai: Done

emagdalenaC2i:
@feketemihai I don't understand why the user should set the IR Model an Filter Domain when a comment template is created.

If this is a base module it should just provide tools for other modules 
* sale_comment_template
* purchase_comment_template
* invoice_comment_template
* stock_picking_comment_template

feketemihai:
@emagdalenaC2i as you see in the l10n-romania PR you use have more comment templates depending on some domain, I want to have one comment for receptions (stock.picking) and another comment for deliveries which are in the same report, the same for invoices.

feketemihai:
@emagdalenaC2i  https://github.com/OCA/l10n-romania/pull/70

pedrobaeza (changes requested):
This module is not for the goal you have intended, and overhead too much its purpose, so please restore it and do a migration per se, and mount in top of it what you want to do.

feketemihai:
@pedrobaeza I disagree, because with this configuration it simplifies the creation of comments, previously you had separate fields in models to choose which template you want to print, that's the essence of the model_id here, you define on the template but you don't need to define them on the other models: https://github.com/OCA/stock-logistics-reporting/tree/12.0/stock_picking_comment_template/models
Leaving the domain field empty, it will work the same as before, you will have only one template "before_lines" and one "after_lines".
I see it more easy to work like this, you don't need to add any fields, code in the other modules, only to add the statement:
 `<xpath expr="//table[hasclass('table-sm')]" position="before">
            <p t-if="o.get_comment_template('before_lines', o.company_id.id, o.partner_id and o.partner_id.id or False)">
                <span t-raw="o.get_comment_template('before_lines', o.company_id.id, o.partner_id and o.partner_id.id or False)"/>
            </p>
        </xpath>`
in the report, add it kept it's usability, define separate templates per company, per partner.

feketemihai:
Screencast-2020-07-11-072954

pedrobaeza:
Automations are good, but when the mechanism can't be fully implemented, then it's better to not add difficulties: you need anyway to add the corresponding glue modules for adding the comment in reports, and also the onchanges (or you need to assume / parametrize the partner field name). On contrary, you are now mixing some concepts that was hard-coded with other that are configurable. You are also restricting now at partner level the possibility of selecting one comment template, not one per model.

I don't see the way to do it a generic solution that worth to increase the difficulty. I would tend to accept this:

- Through a mixin that is inherited in the model (not inheriting base module).
- Restricting the models to the one with that mixin.
- Doing the model field m2m for being able to specify several models that can be tied.
- Doing also m2m the field in the partner for being able to define a comment per model (and this can be problematic on multi-company).
- Defining the partner field through an internal variable in the model class and automatically populating onchange for that partner.

Do you see my point why I think is half baked?

feketemihai:
@pedrobaeza The partner field is optional, so you can have one general template for model (not specifying the partner field), but in the same time you can have separate templates for each partner if you set the partner_id field. An it's the same for company, so you have separate templates depending on the company (if the company field is set), general ones without setting the partner and for each partner if the field is set. You need the corresponding modules just for report inheritance, without any code in them. 
So:
- general templates: not setting the company, partner and domain field, just the model instead of using in the inherited module other fields
- templates by company, setting the company field in the templates
- templates by partner, setting the partner field in the templates
- template by company and partner, setting the company and partner field in the templates
- template by domain, setting the domain field, and it will be eval based on the values in the record
- the models are restricted to the one's that have at least one report defined

Simplified use:
in in the inherited modules just adapt the report like:
https://github.com/OCA/stock-logistics-reporting/blob/a62dc7efdb300685c5481f4a0787d84d107f8039/stock_picking_comment_template/views/report_picking.xml

pedrobaeza:
As said, I don't see the implementation as is. I would accept if the added difficulties are well implemented for a real gain. Meanwhile, I prefer the simpler module, but as said, you can mount on top of this simple module your idea in a separate module.

feketemihai:
@pedrobaeza @emagdalenaC2i Can you recheck, i've added the pr as dependency in https://github.com/OCA/stock-logistics-reporting/pull/109 and https://github.com/OCA/account-invoice-reporting/pull/150 so you can test it over there with real reports

cbssolutions:
@pedrobaeza 
now all is like you told here https://github.com/OCA/reporting-engine/pull/417#issuecomment-656998472
the code is refactored, has migration script, tests ...
and with this is very simple to do mako report at any model and with any conditions.

if you still do not want to modify this, what about renaming it mako_report?

marylla:
Hi there,
I started a functional review and I got the following exception after clicking on menu Invoicing > Configuration > Account Invoice Comments. I don't know if this is caused by this module or by base_comment_template.

@feketemihai Can you check this please?

<pre>
Error:
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 624, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 669, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 350, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 94, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
  File "/.repo_requirements/odoo/odoo/http.py", line 915, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1285, in search_read
    return self.do_search_read(model, fields, offset, limit, domain, sort)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1304, in do_search_read
    return Model.web_search_read(domain, fields, offset=offset, limit=limit, order=sort)
  File "/.repo_requirements/odoo/addons/web/models/models.py", line 39, in web_search_read
    records = self.search_read(domain, fields, offset=offset, limit=limit, order=order)
  File "/.repo_requirements/odoo/odoo/models.py", line 4912, in search_read
    records = self.search(domain or [], offset=offset, limit=limit, order=order)
  File "/.repo_requirements/odoo/odoo/models.py", line 1643, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/.repo_requirements/odoo/odoo/models.py", line 4457, in _search
    query = self._where_calc(args)
  File "/.repo_requirements/odoo/odoo/models.py", line 4177, in _where_calc
    e = expression.expression(domain, self)
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 677, in __init__
    self.parse()
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 858, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'model' in leaf "osv.ExtendedLeaf: ('model', '=', 'account.move') on base_comment_template (ctx: ) "
</pre>

marylla (changes requested):
Hi there,
can we please use the help texts on field definitions as a help for users and not as a help for programmers (because therefore we can add code comments)?

I fear that something like <pre>"[...] (if the model_id has a partner_id field)."</pre> requires the normal user to know about technical names...  and there are a lot of similiar help texts here.
Additionally, I am not so happy about help texts that kind of look like sentences but don't have punctuation at the end. As I add a lot of german translations for OCA modules I am a bit frustrated about it. :/

I know, this is no functional and no real code review but maybe we can take the end user more into consideration? Thanks a lot. :)

CarlosRoca13 (changes requested):
Please squash all the administrative commits. Furthermore squash the 3 final commits into 1 for avoid the noise on the history

pedrobaeza:
Superseded by #481. Thanks anyway for the effort put in the refactoring (which has been credited and preserved its autorship in the other PR).

PR #439, opened 2020-10-09
[FIX] report_qr: New QR Library does not contain the old failure

etobella:
With Odoo 13, this is no longer necessary, as the version of QR was changed:
https://github.com/odoo/odoo/blob/13.0/requirements.txt#L38
https://github.com/odoo/odoo/blob/12.0/requirements.txt#L37

This has made the old issue about background fixed.

With the current code and the default version of QR we have:

imagen
 
and we should obtain:

imagen

### DIFF
--- a/report_qr/models/ir_actions_report.py
+++ b/report_qr/models/ir_actions_report.py
@@ -17,9 +17,8 @@ class IrActionsReport(models.Model):
             "svg-fragment": svg.SvgFragmentImage,
             "svg-path": svg.SvgPathImage,
         }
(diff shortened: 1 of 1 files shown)


etobella:
@LoisRForgeFlow

jarroyomorales (approved):

LoisRForgeFlow (approved):
Tested, looks great!

image

etobella:
@chienandalu This solves an issue you fixed and is no longer present on 13.0

PR #441, opened 2020-10-11
[14.0][MIG] report_xlsx: Migration to 14.0

tupaq:


tonygalmiche (changes requested):
it does not work with runbot:
image

tupaq:
It is fixed,
I closed it by mistake.
Check #443

PR #455, opened 2020-11-12
 [FIX] bi_sql_editor : handle correctly translation for action name

legalsylvain:
Trivial patch.

without this patch, when the cron is executing the refresh of the materialized view, the name of the action is not correctly updated.


alan196 (approved):
Technicall Review

legalsylvain:
@quentinDupont : could you take a look on this one ? it is on production since November.

quentinDupont (approved):

PR #456, opened 2020-11-19
[14.0][MIG] report_xlsx_helper

Saran440:
This module change format to global file

In v14 AbstractModel class can't used self.<new attribute>

Example
`self.format_ws_title` > it will throw error.
So, I change it to `FORMATS["format_ws_title"]`

Call function v13 (old)
Selection_002

Call function v14 (new)
from .report_xlsx_format import FORMATS, XLS_HEADERS
Selection_001

WDYT?

cc : @luc-demeyer


luc-demeyer:
@Saran440 
Thanks a lot for porting this module. 
Code review and testing ok.
Replacing the self.format_xxx by FORMATS["format_xxx."] looks the best way forward (and we may consider backporting this approach also to 13.0 while keeping the self.format_ approach to make it a non-breaking change).

A few remarks:
- I think we should use the occasion of this new port to also add color "grey" to the formats. Can you do this or do you prefer me to make a PR against your PR ?
- I think we should drop the following lines of code from controllers/main.py
               if data["context"].get("lang"):
                    del data["context"]["lang"]
  I took this part of code over from the report_xlsx code a couple of years ago but never understood really why we should do so. 
 It looks logical to me that the language of the user is the default language for the excel reports. 
- In odoo 12 I also had a demo module, cf  https://github.com/OCA/reporting-engine/tree/12.0/report_xlsx_helper_demo.
  I think it will be usefull to add this demo module also to 14.0. If you want I can make a PR against your PR to add a 14.0 port of this module.

luc-demeyer (changes requested):
cf. changes suggested in the conversation.

Saran440:
@luc-demeyer Fixed. Thank you for your reviewed.
- add grey color in format
- delete lang context in controller
- Could you help migrated demo to this PR or new PR?

luc-demeyer:
@Saran440 
Thanks a lot. I'll make a PR on to this PR to add the demo module.

luc-demeyer:
@Saran440 
I have now added the report_xlsx_helper_demo module, can you merge it into this PR ?
Cf. https://github.com/ecosoft-odoo/reporting-engine/pull/1

luc-demeyer (approved):

alan196 (approved):
LGTM 👍

kittiu (approved):
Functional tested.

PR #457, opened 2020-11-19
[12.0][ADD] report_label

ivantodorovich:
This module allows you to create self-adhesive label printing actions on any model.

It provides a generic mechanism that can replace modules like: https://github.com/OCA/partner-contact/tree/12.0/partner_label (@hbrunn) 

---

configure_paperformat

---

configure_report_label

---

label_wizard


report_label/readme/ROADMAP.rst
```diff
@@ -0,0 +1,6 @@
+* `wkhtmltopdf` doesn't always respect dpi, and mm measures don't match. For 
+  this matter, it's recommended to use this module along with 
+  `report_wkhtmltopdf_param` and enable `--disable-smart-shrinking`.
+
+* In `13.0`, it could be based on `mass_operation_abstract`.
```
legalsylvain: I developped ``mass_operation_abstract`` without knowing``action.server``. My bad. 
I think that it should better be based on action server. don't you think ? mass_operation_abstract will be deprecated in next version I think.

marcelsavegnago (approved):
LGTM

legalsylvain (commented):
interesting module. thanks !

CC : @quentinDupont

report_label/readme/ROADMAP.rst
ivantodorovich: Yes I see your point.
I guess modules like this could add a new action "type" (`state` selection field).

imagine `mass_editing` like this:
 *(and other fields with conditional visibility)*
image

Is that what you have in mind?

hbrunn (approved):

report_label/readme/ROADMAP.rst
legalsylvain: > imagine mass_editing like this: ... Is that what you have in mind?

Exactly !

i began to developp ``mass_operation_abstract``, with ``mass_editing`` (discussion [here](https://github.com/OCA/server-ux/pull/94#issuecomment-538015686)) without knowing ``action.server``. When I proposed a ``mass_action`` module, @hbrunn told me that I was reinventing the wheel [here](https://github.com/OCA/server-ux/pull/97#pullrequestreview-317654463).

There are other modules with the same logic. 
- action_server_navigate : https://github.com/OCA/server-backend/pull/65 
- server_action_sort : https://github.com/OCA/server-backend/pull/70 that is a generic tools that could some module like purchase_order_line_deep_sort (review welcome BTW ;-))

So yes, I think that : 

- that ``mass_editing`` should inherit also from ``server.action``
- you will save code, if your module inherits from ``server.action``. but not a blocking point. It can be just written in the roadmap. It's a huge refactoring, and I could understand that it is not in the scope of this PR.

ivantodorovich: I've refactored so that it inherits `ir.actions.server`.
Indeed, this seems to be the way to go. Code is much simpler now.

Thanks!! :tada:

gregory-moka (approved):

gregory-moka:
Nice Job @ivantodorovich  !

report_label/readme/ROADMAP.rst
legalsylvain: Nice ! 

No time to review, but with 3 approvals, I guess  @OCA/reporting-engine-maintainers  could merge it.

thanks !

PR #459, opened 2020-12-25
[12.0][ADD] report_qweb_encrypt

kittiu:
This module based on the work from @etobella
https://github.com/tegin/reporting-engine/tree/12.0-add-report_qweb_encrypted/report_qweb_encrypt


kittiu:
@etobella thanks for the original code, it works already. I just add option that I need, which is to auto generate the password in addition to manual input.

image

Next I will add some test just to pass through the codes. If you need any other options please let me know. :)

report_qweb_encrypt/models/ir_actions_report.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class IrActionsReport(models.Model):
+
+    _inherit = 'ir.actions.report'
+
+    encrypt = fields.Boolean(
+        string="Encryption",
+    )
+    encrypt_password_option = fields.Selection(
```
etobella: With this is not required the flag. Encrypt flag is true if this value is set. WDYT?

kittiu: Agreed, I think we can remove encrypt_password_option and use encrypt as the options itself.

kittiu: WDYT?

image

etobella: Gr8

kittiu:
Test added, ready for review.
**Note:** Test case still don't have case for manual password yet, don't know how to call report_download() from test script. If you have sample, please let me know and I will add it. Never the less, current is suffice I guess.

marcelsavegnago (commented):
I am testing by the runbot and when typing a password and pressing OK the password is deleted and is waiting to be typed again.

image

marcelsavegnago (commented):
I tested the printout of the quote by pressing the PRINT button on the status bar and in this case no encryption password was requested. Is this behavior expected?

image

kittiu:
> I am testing by the runbot and when typing a password and pressing OK the password is deleted and is waiting to be typed again.

Thank you for your comment. I don't have that issue though. Can you try again locally?

Peek 2020-12-27 10-46

kittiu:
> I tested the printout of the quote by pressing the PRINT button on the status bar and in this case no encryption password was requested. Is this behavior expected?

Yes, it is expected. For manual key-in it is only applicable with **Print Action** because it need to pass the context through URL (the report action itself can't pass normal context).
By odoo standard, it is not possible to pass the context value when printing. The Print button, as  well as any other PDF creation like Send Email will not work (because it will need a dialog or the like, to key-in password).

I wrote in the field help:
* Manual Input Password: allow user to key in password on the fly.  This option available only on document print action.
* Auto Generated Password: system will auto encrypt password when PDF created, based on provided python syntax.

marcelsavegnago (approved):
LGTM

etobella (approved):

kittiu:
@OCA/reporting-engine-maintainers can this one merged, we also migrated to v14.
https://github.com/OCA/reporting-engine/pull/464

StefanRijnhart:
Nice module!

/ocabot merge nobump

PR #461, opened 2020-12-30
[FIX] Import time from safe_eval

jacob-willdooit:
In v14, Odoo now requires that modules be wrapped by safe_eval.
Attempting to evaluate a string with modules that are not imported from safe_eval will throw an error.

See below references.

https://github.com/odoo/odoo/blob/61e5389236ad767827ef2fe039aa78b20848298e/addons/web/controllers/main.py#L41
https://github.com/odoo/odoo/blob/61e5389236ad767827ef2fe039aa78b20848298e/odoo/tools/safe_eval.py#L416


jacob-willdooit:
Someone removed the time module in another PR, I've updated my PR to re-add the time module from safe_eval.

PR #463, opened 2021-01-05
Add "only-inner" class in report_qweb_element_page_visibility module

pelayog:
Add the ability to hide an item on both the first and last pages of a report to the `report_qweb_element_page_visibility` module


cvinh:
@pelayog could you please do the same to v14.0 ?

Issue #468, opened 2021-01-20
bi_sql_editor - action name for transient model


mattmutex:
When transient model is updated by a scheduled action the action name is not translated so the users cannot see the correct new action name if he uses a different langage

mattmutex:
```
    @api.multi
    def _refresh_materialized_view(self):
        for sql_view in self.filtered(lambda x: x.is_materialized):
            req = "REFRESH %s VIEW %s" % (
                sql_view.materialized_text, sql_view.view_name)
            self._log_execute(req)
            sql_view._refresh_size()
            if sql_view.action_id:
                # Alter name of the action, to display last refresh
                # datetime of the materialized view
                sql_view.action_id.name = sql_view._prepare_action_name()
                **#translate action name
                translations = self.env['ir.translation'].search([('name','=','ir.actions.act_window,name'),
                                                                 ('res_id','=',sql_view.action_id.id)])
                for translation in translations:
                    translation.write({'value': sql_view.action_id.name})**
```

PR #469, opened 2021-01-21
[12][FIX] translate action name for view cron update

mattmutex:
To fix pb of action name translation if code it is suitable for you...

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -652,6 +652,11 @@ class BiSQLView(models.Model):
                 # Alter name of the action, to display last refresh
                 # datetime of the materialized view
                 sql_view.action_id.name = sql_view._prepare_action_name()
+                #translate action name
+                translations = self.env['ir.translation'].search([('name','=','ir.actions.act_window,name'),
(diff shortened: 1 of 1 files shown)


mattmutex:
#468

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -652,6 +652,11 @@ def _refresh_materialized_view(self):
                 # Alter name of the action, to display last refresh
                 # datetime of the materialized view
                 sql_view.action_id.name = sql_view._prepare_action_name()
+                #translate action name
+                translations = self.env['ir.translation'].search([('name','=','ir.actions.act_window,name'),
+                                                                 ('res_id','=',sql_view.action_id.id)])
+                for translation in translations:
+                    translation.write({'value': sql_view.action_id.name})
```
legalsylvain, line 659: I think this implementation will disable the ``_prepare_action_name()`` feature. don't you think ?

I think you should better make something like : 

```
for translation in translations:
    sql_view.action_id.with_context(lang=translation.lang).name = sql_view.with_context(lang=translation.lang)._prepare_action_name()
```

or something like that...

Issue #470, opened 2021-01-22
13.0 report_xlsx multi-company bug


luc-demeyer:
Report_xlsx drops the 'allowed_company_ids' context field.
Instead of the allowed_company_ids of the session the default company_id of the user in present in the context.
The result is a record rule error messages when making a report when the selected company is different from the default company of the user.

This is caused by the following lines in the action_manager_report.js code:

                if (cloned_action.context.active_ids) {
                    new_url += "/" + cloned_action.context.active_ids.join(",");
                }
 In case of 'active_ids' the session context that is present in the 'cloned_action' is dropped.

sbidoul:
@luc-demeyer this isssue is resolved, right ?

luc-demeyer:
yes

PR #471, opened 2021-01-23
[13.0][FIX]report_xlsx multi-company bug

luc-demeyer:
This PR fixes https://github.com/OCA/reporting-engine/issues/470
Context dropping (hence also allowed_company_ids) causing access errors in multi-company environment.

We'll also need to forward port this fix to 14.0.

### DIFF
--- a/report_xlsx/static/src/js/report/action_manager_report.js
+++ b/report_xlsx/static/src/js/report/action_manager_report.js
@@ -39,6 +39,7 @@ odoo.define("report_xlsx.report", function(require) {
                     url: new_url,
                     data: {
                         data: JSON.stringify([new_url, type]),
+                        context: JSON.stringify(cloned_action.context),
                     },
                     success: resolve,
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):

pedrobaeza:
Is there the possibility of some context needed for the report being dropped with this? Maybe it's better to just remove the offending keys?

luc-demeyer:
Now we drop the context resulting in broken report_xlsx based reports for multi-company, I don't see which harm we can cause by passing on the session context.

I tested with a couple of commonly used reports in the OCA modules and didn't experience any problem.
Also Odoo standard reports use this logic.
If there is any report that breaks on this than we should imho adapt the logic in that report but I don't expect issues.

pedrobaeza:
OK, I see, let's merge it then:

/ocabot merge patch

I understand this is needed as well in 14.0

PR #481, opened 2021-02-09
[13.0] [MIG] base_comment_template: Migration to v13

CarlosRoca13:
cc @Tecnativa TT26763

this PR superseds https://github.com/OCA/reporting-engine/pull/417 attending @marylla and my own comment. For the superseding I kept the functionality added by @feketemihai and @alexteodor 

@joao-p-marques @chienandalu review this, please


chienandalu (changes requested):
Tests are failing

chienandalu (commented):
Still no luck

CarlosRoca13:
@chienandalu Now solved :100:

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+            (
+                "fields_id",
+                "=",
+                env.ref("base.field_res_partner_property_comment_template_id").id,
+            )
+        ]
+    )
+    if properties:
+        for template in properties.mapped("value_reference"):
+            template_id = template.value_reference.split(",")[-1]
```
joao-p-marques: I think this would probably break, as you are ultimately trying to access `value_reference` of the `value_reference` of `ir.property` (and the latest will be a string as you are using `mapped()`). It would probably be something like:
```suggestion
        for template in properties.mapped("value_reference"):
            if not template:
                continue
            template_id = template.split(",")[-1]
```

Example in Odoo console:
```python
>>> env["ir.property"].search([]).mapped("value_reference")
[False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, 'product.pricelist,1', False, 'account.payment.term,4', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,7', 'account.payment.term,7', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,4', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,4', 'account.payment.term,7', 'product.pricelist,1', 'product.pricelist,1', 'product.pricelist,1', 'account.payment.term,7', 'account.payment.term,7', 'account.payment.term,7', 'account.account,193', 'account.account,188', 'account.account,357', 'account.account,481']
>>> env["ir.property"].search([]).mapped("value_reference")[-1]
'account.account,481'  # This is a string
>>> env["ir.property"].search([]).mapped("value_reference")[-1].value_reference
AttributeError: 'str' object has no attribute 'value_reference'
>>> 
```

joao-p-marques (commented):
Code review :+1: 

Overall LGTM. Just a comment about the migration scripts.

CarlosRoca13:
@joao-p-marques changed!

chienandalu (approved):
Ok, let's see it in action with the extensions :+1:

joao-p-marques (approved):

etobella (approved):
We will need to wait until the related modules are finished, so I think we should proceed with merge too

feketemihai:
@pedrobaeza and Tecnativa guys, thanks for continuing it.

pedrobaeza:
I'm now seeing that the company dependent field in res.partner has been lost, and we are not doing multi-company compliant this feature. We need to recover such option, and change migration scripts accordingly.

cbssolutions:
from my point of view I think that is going to be more clear, less code,
simpler to explain, less complexity if we are going to let only the domain
(no company, no partner...).
another thing is that is going to be simpler if model is going to be
many2one and not many2many

base_comment_template/security/security.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model="ir.rule" id="rule_multicompany">
+        <field name="name">Base comment multi-company</field>
+        <field name="model_id" ref="model_base_comment_template" />
+        <field name="global" eval="True" />
+        <field name="domain_force">
+             ['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
```
pedrobaeza, line 8: This rule is incorrect. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0:
```suggestion
             ['|',('company_id','=',False),('company_id','in',company_ids)]</field>
```

pedrobaeza:
OK, I see now what you mean, and I think it can be OK, but I don't see clear then the mix of several templates from the same company. All of them are put together? There's no sequence that determines the order. It should be added as well and check if the mix is done.

PR #482, opened 2021-02-09
[13.0][FIX] report_xlsx_helper: force Libreoffice to recompute formulas

StefanRijnhart:
As per https://xlsxwriter.readthedocs.io/faq.html:

Q. Why do my formulas show a zero result in some, non-Excel applications?
A. [...] Or, you can set a blank result in the formula, which will also force
recalculation.

Before the fix:
image

After the fix:
image

I could not add the lines inline because of 
`report_xlsx_helper/report/report_xlsx_abstract.py:564:5: C901 'ReportXlsxAbstract._write_line' is too complex (18)`


StefanRijnhart:
14.0 version is here: https://github.com/OCA/reporting-engine/pull/483

pedrobaeza (approved):

luc-demeyer (approved):
code review plus testing with xlsx ok.

PR #496, opened 2021-04-02
[REF] bi_sql_editor : move function into abstract module

legalsylvain:
remove function, that is now defined in the abstract model to be available in other module.

See : https://github.com/OCA/server-tools/pull/2061/

this PR should be merged once other PR is merged.


len-foss (approved):
:+1:

legalsylvain:
CC :@quentinDupont

PR #497, opened 2021-04-08
[13.0][IMP] base_comment_template: Add function to get_comment_template records

victoralmau:
Add function to get_comment_template records.

It's related to https://github.com/OCA/account-invoice-reporting/pull/174#discussion_r609506163

Please @CarlosRoca13 and @ernestotejeda can you review it?

@Tecnativa TT26758

### DIFF
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -16,12 +16,9 @@ class CommentTemplate(models.AbstractModel):
         "in reports based on created comment templates"
     )
 
-    def get_comment_template(
+    def get_comment_template_records(
         self, position="before_lines", company_id=False, partner_id=False
     ):
-        """ Method that is called from report xml and is returning the
-            position template as a html if exists
-        """
         self.ensure_one()
         if not company_id:
             company_id = self.env.company.id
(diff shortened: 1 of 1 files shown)


ernestotejeda (approved):

ernestotejeda (commented):
I think the 'priority' field should be taken into account here, isn't it?

CarlosRoca13 (approved):

victoralmau:
> I think the 'priority' field should be taken into account here, isn't it?

Actually, in `get_comment_template_records()` priority appear:

```
....
templates = self.env["base.comment.template"].search(
            default_dom, order="priority"
        )
.....
```

ernestotejeda:
> Actually, in get_comment_template_records() priority appear

oh yes it's there, sorry :)

victoralmau:
I have try to apply in 14.0 (FWP) but it's not apply migration correctly, right?  (`base.comment.template` model have disappear)
What do you think @pedrobaeza  and @CarlosRoca13 ?

pedrobaeza:
We need to check what was done in v14 and maybe compatibilize it with this scheme. We have been a bit late to the party and a version jump has been made.

PR #499, opened 2021-04-12
[13.0][FIX] base_comment_template: Fix tests: test_company_general_template

victoralmau:
Fix tests: `test_company_general_template`.

Please @CarlosRoca13  and @ernestotejeda can you review it?

@Tecnativa TT29138

PR #495, opened 2021-03-30
[14.0] base_comment_template: add templating

petrus-v:
By this PR I'm proposing to improve `base_comment_template` to be able to use jinja templating on base comments template

PR #500, opened 2021-04-12
[14.0][IMP] base_comment_template: Include changes from 13.0

victoralmau:
Include changes from 13.0 to standardize both versions:

- Migration to v13: https://github.com/OCA/reporting-engine/pull/481
- Add function to get_comment_template records: https://github.com/OCA/reporting-engine/pull/497
- Fix tests: test_company_general_template: https://github.com/OCA/reporting-engine/pull/499
- Fix Change domain in rule according to company: https://github.com/OCA/reporting-engine/pull/501

Locked by:
- [x] https://github.com/OCA/reporting-engine/pull/504

Please @CarlosRoca13  and @ernestotejeda can you review it?

@Tecnativa TT29138


CarlosRoca13 (approved):

chienandalu (commented):
Is it very hard to cherry-pick the original commits that include these changes?

chienandalu (approved):

base_comment_template/README.rst
```diff
@@ -14,23 +14,33 @@ Base Comments Templates
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Freporting--engine-lightgray.png?logo=github
-    :target: https://github.com/OCA/reporting-engine/tree/14.0/base_comment_template
+    :target: https://github.com/OCA/reporting-engine/tree/13.0/base_comment_template
```
ernestotejeda: It looks like you generated the readme for branch 13.0

base_comment_template/__manifest__.py
```diff
@@ -3,18 +3,19 @@
 
 {
     "name": "Base Comments Templates",
-    "summary": "Comments templates on documents",
+    "summary": "Add conditional mako template to any report "
```
ernestotejeda:
```suggestion
    "summary": "Add conditional mako template to any report"
```

base_comment_template/tests/fake_models.py
```diff
@@ -0,0 +1,34 @@
...
+# Copyright 2018 Camptocamp
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+def setup_test_model(env, model_cls):
+    """Pass a test model class and initialize it.
+
+    Courtesy of SBidoul from https://github.com/OCA/mis-builder :)
```
ernestotejeda, line 12: this dockstring should be kept?

ernestotejeda (approved):
LGTM.
Just a few comments.

base_comment_template/README.rst
victoralmau: Although bot will regenerated it, i have changed.

base_comment_template/tests/fake_models.py
victoralmau: It's exists now in 13.0, IMO kep according commits.

base_comment_template/README.rst
pedrobaeza: This has not been changed at the end

base_comment_template/models/__init__.py
```diff
@@ -1,3 +1,3 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from . import comment
```
pedrobaeza, line 2: I don't see the removal of this file.

base_comment_template/models/res_partner.py
```diff
@@ -1,19 +1,14 @@
...
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    property_comment_template_id = fields.Many2one(
+    base_comment_template_ids = fields.Many2many(
```
pedrobaeza, line 11: You are losing here the `company_dependent` property, which is important. Migration scripts are needed for converting data from many2one to many2many (and with property) if we change to m2m, but I don't know the use case for requiring a m2m. Where that was changed?

pedrobaeza (changes requested):
@petrus-v your opinion is welcome here as well, as you were the porter.

base_comment_template/models/res_partner.py
pedrobaeza: Commented in https://github.com/OCA/reporting-engine/pull/481#issuecomment-821799830, it's OK to change m2m + multi-company security, but migration scripts should be provided from those on 14.0.1.0.0 to go to this approach, and not fail when migrating from 13.0.1.0.0 to 14.0.1.0.0.

base_comment_template/models/base_comment_template.py
```diff
@@ -0,0 +1,190 @@
...
+            return ""
+        return self.env["mail.template"]._render_template(
+            template.text, self._name, self.ids, post_process=True
+        )[self.id]
+
+
+class BaseCommentTemplate(models.Model):
+    """Comment templates printed on reports"""
+
+    _name = "base.comment.template"
```
petrus-v, line 12: I'm wondering suggested way to move from previous version to this new version?

It would be nice to provide some compatibility with deprecated method that could be remove later (in the next version ?), or at least raise an explicit message that explain to developer how to manage that change.

For instance `get_value` is in used in other modules likes this OCA module [account_comment_template](https://github.com/OCA/account-invoice-reporting/blob/14.0/account_comment_template/models/account_move.py#L32)

base_comment_template/tests/test_base_comment_template.py
```diff
@@ -1,37 +1,171 @@
...
+                "model_ids": [(6, 0, self.user_obj.ids)],
+                "priority": 2,
+            }
+        )
 
-    def test_get_value_without_partner(self):
-        self.assertEqual(self.template_id.get_value(), "<p>Text before lines</p>")
+        self.assertEqual(new_template.text, "Text before lines 1")
 
-    def test_get_value_with_partner(self):
```
petrus-v, line 27: While migration I've had that test to improve the coverage, here you get a nice coverage regarding that code but could be nice to ensure lang param is properly forward by using a different language and asserting expected result, I haven't seen such test case kept.

base_comment_template/README.rst
```diff
@@ -77,7 +134,14 @@ Contributors
 * `Druidoo <https://www.druidoo.io>`_:
 
   * Iván Todorovich <ivan.todorovich@druidoo.io>
-* Pierre Verkest <pierreverkest84@gmail.com>
```
petrus-v, line 80: Does my works has no value? :sweat_smile:

base_comment_template/tests/test_base_comment_template.py
```diff
@@ -1,37 +1,171 @@
...
-class TestResPartner(TransactionCase):
-    def setUp(self):
-        super(TestResPartner, self).setUp()
-        self.template_id = self.env["base.comment.template"].create(
+from .fake_models import ResUsers, setup_test_model, teardown_test_model
+
+
+class TestCommentTemplate(common.SavepointCase):
+    at_install = False
+    post_install = True
```
petrus-v: I'm not sure, should we use decorator instead ?

```suggestion
from odoo.tests import tagged


@tagged("post_install", "-at_install")
class TestCommentTemplate(common.SavepointCase):
```

(not sure you should apply this suggestion import wont be at the right place)

base_comment_template/models/ir_model.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def _reflect_model_params(self, model):
+        vals = super(IrModel, self)._reflect_model_params(model)
+        vals["is_comment_template"] = issubclass(
+            type(model), self.pool["comment.template"]
+        )
+        return vals
+
+    @api.model
+    def _instanciate(self, model_data):
```
petrus-v, line 24: I've seen in unit test that a res.users fake model is created, in suppose if we add a test that instantiate that model this code would be called, isn't it ?

also I'm not sure what that code is supposed to do, it's look like a bit too magic I wondering if a mixin would do the trick ?

petrus-v (changes requested):
It would be really lovely to provide a smoothy transition for those (likes me) who use that base module from different place in different other modules !

base_comment_template/models/res_partner.py
victoralmau: OK but first, it's important to apply m2m in 13.0 in other PR, right?

victoralmau:
According to https://github.com/OCA/reporting-engine/pull/504 it's ready to review now, right?
I'm going to create another PR for `account_comment_template` in 14.0 according to these changes.

base_comment_template/models/base_comment_template.py
victoralmau: Now, it's not necessary.

base_comment_template/tests/test_base_comment_template.py
victoralmau: Now, it's not necessary.

base_comment_template/models/base_comment_template.py
petrus-v: Could you explain why or how ? I'm not sure to understand what's is not necessary, I've some modules (not OCA modules) that's depends on this one.

victoralmau: IMO, these examples: https://github.com/OCA/account-invoice-reporting/pull/174 (13.0) and https://github.com/OCA/account-invoice-reporting/pull/194 (14.0) will help you understand how other addons (OCA or custom addons) `comment.template` will use and how to display them in a report.

base_comment_template/README.rst
pedrobaeza: Please regenerate this for not having this double diff

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 NextERP Romania SRL
```
pedrobaeza: Remove this, but add the corresponding script for this version, taking into account that you can come from 13.0.2.0.0.

base_comment_template/migrations/13.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
```
pedrobaeza: The same

base_comment_template/tests/test_base_comment_template.py
pedrobaeza: This shouldn't be put at all. No need to execute at the end.

pedrobaeza (changes requested):
Attend also the rest of the comments

base_comment_template/models/ir_model.py
victoralmau: Yest, exactly.
These addon need to test if some model inherit comment template and everything work fine. 
We need add some code to ir_model because only with inherit comment.template we can select model in the templates.
Maybe some documentation about fake_models https://pypi.org/project/odoo-test-helper/ help you.

victoralmau:
It's ready to review.
@petrus-v can you re-review?

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
```
pedrobaeza: These folders from old version must be removed.

base_comment_template/__manifest__.py
```diff
@@ -3,18 +3,19 @@
 
 {
     "name": "Base Comments Templates",
-    "summary": "Comments templates on documents",
+    "summary": "Add conditional mako template to any report"
+    "on models that inherits comment.template.",
     "version": "14.0.1.0.0",
```
pedrobaeza: Change this to `14.0.2.0.0`

base_comment_template/migrations/14.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,29 @@
...
+from openupgradelib import openupgrade
+
+field_renames = [
+    ("base.comment.template", "base_comment_template", "priority", "sequence"),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Not tested
```
pedrobaeza: This should be executed only if coming from 14.0.1.0.0 to 14.0.2.0.0, so change the folder name, and we need to check if we are coming from 13.0.x or from 14.0.1.0.0, so look for a data condition that is only in 14.0.1.0.0.

It should be basically the same as what we have done from 13.0.1.0.0 to 13.0.2.0.0, that is tested, so remove this not elegant comment and make sure it works.

pedrobaeza (changes requested):

pedrobaeza:
@petrus-v we don't want to break your current module status, but there were also a jump from 12 to 14 in this migration, and with a design that is in the middle of the old and the new one, so we need to unblock this transition (that we are trying to be smooth through migration script). Is there any doubt you have with the changes?

petrus-v:
@pedrobaeza 

> @petrus-v we don't want to break your current module status, but there were also a jump from 12 to 14 in this migration, and with a design that is in the middle of the old and the new one, so we need to unblock this transition (that we are trying to be smooth through migration script). Is there any doubt you have with the changes?

I haven't deep into real impact in my customer context, I've also [suggested change in an other PR](https://github.com/OCA/reporting-engine/pull/495) to be able to use jinja template that I use, but anyway I understand your point go ahead I'll adapt all my code once this is merged.

I'm pretty confident as I've tools and tests to prevent myself to deploy breaking change in production and I'll be able to freeze things in the mean time to adapt code or fix before deploy the last version once I'll be happy.

Thanks again for all your comments, was very appreciated.

victoralmau:
Thanks @pedrobaeza, changes done.

joao-p-marques (approved):

pedrobaeza (approved):

PR #504, opened 2021-04-22
[13.0][IMP] base_comment_template: Refactor code and convert to Many2Many

victoralmau:
Refactor code and convert to Many2Many.

@Tecnativa TT29364

### DIFF
--- a/base_comment_template/migrations/13.0.1.0.0/pre-migration.py
+++ b/base_comment_template/migrations/13.0.1.0.0/pre-migration.py
@@ -1,4 +1,5 @@
 # Copyright 2020 NextERP Romania SRL
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
 
@@ -6,23 +7,16 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version):
     # Not tested
-    properties = env["ir.property"].search(
-        [
-            (
-                "fields_id",
-                "=",
-                env.ref("base.field_res_partner_property_comment_template_id").id,
-            )
-        ]
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO base_comment_template_res_partner_rel
+        (res_partner_id, base_comment_template_id)
+        SELECT SPLIT_PART(ip.res_id, ',', 2)::int AS res_partner_id,
+        SPLIT_PART(ip.value_reference, ',', 2)::int AS base_comment_template_id
+        FROM ir_property ip
+        JOIN ir_model_fields imf ON ip.fields_id = imf.id
+        WHERE imf.name = 'property_comment_template_id'
+        AND imf.model = 'res.partner'
+        """,
     )
-    if properties:
-        for template in properties.mapped("value_reference"):
-            if not template:
-                continue
-            template_id = template.split(",")[-1]
-            if template_id:
-                template = env["base.comment.template"].browse(template_id)
-                part_prop = properties.filtered(lambda p: p.value_reference == template)
-                template.partner_ids = [
-                    (prop["res_id"] or "").split(",")[-1] for prop in part_prop
-                ]
--- a/base_comment_template/migrations/13.0.2.0.0/pre-migration.py
+++ b/base_comment_template/migrations/13.0.2.0.0/pre-migration.py
@@ -1,0 +1,12 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+field_renames = [
+    ("base.comment.template", "base_comment_template", "priority", "sequence"),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, field_renames)
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -1,76 +1,9 @@
 # Copyright 2014 Guewen Baconnier (Camptocamp SA)
 # Copyright 2013-2014 Nicolas Bessi (Camptocamp SA)
 # Copyright 2020 NextERP Romania SRL
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from lxml import etree
-
-from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
-from odoo.tools.safe_eval import safe_eval
-
-
-class CommentTemplate(models.AbstractModel):
-    _name = "comment.template"
-    _description = (
-        "base.comment.template to put header and footer "
-        "in reports based on created comment templates"
-    )
-
-    def get_comment_template_records(
-        self, position="before_lines", company_id=False, partner_id=False
-    ):
-        self.ensure_one()
-        if not company_id:
-            company_id = self.env.company.id
-        present_model_id = self.env["ir.model"].search([("model", "=", self._name)])
-        default_dom = [
-            ("model_ids", "in", present_model_id.id),
-            ("position", "=", position),
-        ]
-        lang = False
-        if partner_id and "partner_id" in self._fields:
-            default_dom += [
-                "|",
-                ("partner_ids", "=", False),
-                ("partner_ids", "in", partner_id),
-            ]
-            lang = self.env["res.partner"].browse(partner_id).lang
-        if company_id and "company_id" in self._fields:
-            if partner_id and "partner_id" in self._fields:
-                default_dom.insert(-3, "&")
-            default_dom += [
-                "|",
-                ("company_id", "=", company_id),
-                ("company_id", "=", False),
-            ]
-        templates = self.env["base.comment.template"].search(
-            default_dom, order="priority"
-        )
-        if lang:
-            templates = templates.with_context({"lang": lang})
-        return templates
-
-    def get_comment_template(
-        self, position="before_lines", company_id=False, partner_id=False
-    ):
-        """ Method that is called from report xml and is returning the
-            position template as a html if exists
-        """
-        self.ensure_one()
-        templates = self.get_comment_template_records(
-            position=position, company_id=company_id, partner_id=partner_id
-        )
-        template = False
-        if templates:
-            for templ in templates:
-                if self in self.search(safe_eval(templ.domain or "[]")):
-                    template = templ
-                    break
-        if not template:
-            return ""
-        return self.env["mail.template"]._render_template(
-            template.text, self._name, self.id, post_process=True
-        )
+from odoo import fields, models
 
 
 class BaseCommentTemplate(models.Model):
@@ -78,11 +11,12 @@ class BaseCommentTemplate(models.Model):
 
     _name = "base.comment.template"
     _description = "Comments Template"
+    _order = "sequence,id"
 
     active = fields.Boolean(default=True)
     position = fields.Selection(
         string="Position on document",
-        selection=[("before_lines", "Before lines"), ("after_lines", "After lines")],
+        selection=[("before_lines", "Top"), ("after_lines", "Bottom")],
         required=True,
         default="before_lines",
         help="This field allows to select the position of the comment on reports.",
@@ -101,7 +35,7 @@ class BaseCommentTemplate(models.Model):
         help="This is the text template that will be inserted into reports.",
     )
     company_id = fields.Many2one(
-        "res.company",
+        comodel_name="res.company",
         string="Company",
         ondelete="cascade",
         index=True,
@@ -110,79 +44,46 @@ class BaseCommentTemplate(models.Model):
     )
     partner_ids = fields.Many2many(
         comodel_name="res.partner",
+        relation="base_comment_template_res_partner_rel",
+        column1="res_partner_id",
+        column2="base_comment_template_id",
         string="Partner",
-        ondelete="cascade",
+        readonly=True,
         help="If set, the comment template will be available only for the selected "
         "partner.",
     )
-
     model_ids = fields.Many2many(
         comodel_name="ir.model",
         string="IR Model",
         ondelete="cascade",
+        domain=[
+            ("is_comment_template", "=", True),
+            ("model", "!=", "comment.template"),
+        ],
         required=True,
         help="This comment template will be available on this models. "
         "You can see here only models allowed to set the coment template.",
     )
-
     domain = fields.Char(
-        "Filter Domain",
+        string="Filter Domain",
         required=True,
         default="[]",
         help="This comment template will be available only for objects "
         "that satisfy the condition",
     )
-
-    priority = fields.Integer(
-        default=10, copy=False, help="the highest priority = the smallest number",
+    sequence = fields.Integer(
+        required=True, default=10, help="The smaller number = The higher priority"
     )
 
-    @api.constrains("domain", "priority", "model_ids", "position")
-    def _check_partners_in_company_id(self):
-        templates = self.search([])
-        for record in self:
-            other_template_same_models_and_priority = templates.filtered(
-                lambda t: t.priority == record.priority
-                and set(record.model_ids).intersection(record.model_ids)
-                and t.domain == record.domain
-                and t.position == record.position
-                and t.id != record.id
+    def name_get(self):
+        """Redefine the name_get method to show the template name with the position.
+        """
+        res = []
+        for item in self:
+            name = "{} ({})".format(
+                item.name, dict(self._fields["position"].selection).get(item.position)
             )
-            if other_template_same_models_and_priority:
-                raise ValidationError(
-                    _(
-                        "There are other records with same models, priority, "
-                        "domain and position."
-                    )
-                )
-
-    @api.model
-    def fields_view_get(
-        self, view_id=None, view_type="form", toolbar=False, submenu=False
-    ):
-        # modify the form view of base_commnent_template
-        # Add domain on model_id to get only models that have a report set
-        # and those whom have inherited this model
-        res = super().fields_view_get(
-            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
-        )
-        if view_type == "form":
-            doc = etree.XML(res["arch"])
-            for node in doc.xpath("//field[@name='model_ids']"):
-                report_models = self.env["ir.actions.report"].search([]).mapped("model")
-                model_ids = (
-                    self.env["ir.model"]
-                    .search(
-                        [
-                            ("model", "in", report_models),
-                            ("is_comment_template", "=", True),
-                            "!",
-                            ("name", "=like", "ir.%"),
-                        ]
-                    )
-                    .ids
-                )
-                model_filter = "[('id','in'," + str(model_ids) + ")]"
-                node.set("domain", model_filter)
-            res["arch"] = etree.tostring(doc, encoding="unicode")
+            if self.env.context.get("comment_template_model_display"):
+                name += " (%s)" % ", ".join(item.model_ids.mapped("name"))
+            res.append((item.id, name))
         return res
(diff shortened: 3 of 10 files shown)


base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,47 @@ class CommentTemplate(models.AbstractModel):
...
+            partner = record[self._comment_template_partner_field_name]
+            record.comment_template_ids = [(5,)]
+            templates = self.env["base.comment.template"].search(
+                [
+                    ("id", "in", partner.base_comment_template_ids.ids),
+                    ("model_ids.model", "=", self._name),
+                ]
+            )
+            for template in templates:
+                if self in self.search(safe_eval(template.domain or "[]")):
```
pedrobaeza:
```suggestion
                if not template.domain or self in self.search(safe_eval(template.domain)):
```

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,47 @@ class CommentTemplate(models.AbstractModel):
...
+            result.append((record.id, name))
+        return result
 
 
 class BaseCommentTemplate(models.Model):
     """Comment templates printed on reports"""
 
     _name = "base.comment.template"
     _description = "Comments Template"
+    _order = "priority"
```
pedrobaeza: This is usually called `sequence`. The existing field is already at the view with the handle widget?

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,47 @@ class CommentTemplate(models.AbstractModel):
         "base.comment.template to put header and footer "
         "in reports based on created comment templates"
     )
+    _comment_template_partner_field_name = "partner_id"
```
joao-p-marques: I don't really understand the need for this variable... Could you explain a bit, maybe through a comment?

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,47 @@ class CommentTemplate(models.AbstractModel):
...
+            record.comment_template_ids = [(5,)]
+            templates = self.env["base.comment.template"].search(
+                [
+                    ("id", "in", partner.base_comment_template_ids.ids),
+                    ("model_ids.model", "=", self._name),
+                ]
+            )
+            for template in templates:
+                if self in self.search(safe_eval(template.domain or "[]")):
+                    record.comment_template_ids = [(4, template.id)]
```
joao-p-marques: You can store this in a variable along with the one you do in line 34 above, and instead of writing twice in the compute, do a single one at the end:
```python
domain = [(5,)]

# ...

for ... :
    if ... :
        domain = [(6, 0, template.ids)]

record.comment_template_ids = domain
```

base_comment_template/models/base_comment_template.py
pedrobaeza: This should have a comment indeed for explaining the purpose, but I tell you quickly: it's for allowing to have another field that determines the source partner for the comment templates. Imagine other models where the partner field is called `customer_id` for example...

base_comment_template/models/base_comment_template.py
pedrobaeza: Since v13, it doesn't matter the number of assignations you do inside a compute. Everything goes to cache, so it's not too much overhead to do it in several steps.

base_comment_template/models/base_comment_template.py
joao-p-marques: Ok, I see... :+1: Thanks

base_comment_template/models/base_comment_template.py
joao-p-marques: Oh ok :+1: MB 
Thanks for the pointer

base_comment_template/models/base_comment_template.py
victoralmau: It's a good moment to rename field. I have changed name field to `sequence` and have added migration script.

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,46 @@ class CommentTemplate(models.AbstractModel):
         "base.comment.template to put header and footer "
         "in reports based on created comment templates"
     )
+    """
+    It's for allowing to have another field that determines the source partner for
+    the comment templates.
+    Imagine other models where the partner field is called customer_id for example.
+    """
+    _comment_template_partner_field_name = "partner_id"
```
joao-p-marques: Thanks for the comment @victoralmau 
However, I think this would need a comment, not really a docstring for the class:
```suggestion
    # This field allows to set any given field that determines the source partner for
    # the comment templates downstream.
    # E.g.: other models where the partner field is called customer_id.
    _comment_template_partner_field_name = "partner_id"
```

base_comment_template/models/base_comment_template.py
```diff
@@ -124,38 +104,44 @@ class BaseCommentTemplate(models.Model):
...
         required=True,
         default="[]",
         help="This comment template will be available only for objects "
         "that satisfy the condition",
     )
-
-    priority = fields.Integer(
-        default=10, copy=False, help="the highest priority = the smallest number",
+    sequence = fields.Integer(
+        required=True, default=1, help="the highest sequence = the smallest number"
```
joao-p-marques: This help message is a bit confusing... I'd leave it as before, with isn't wrong:
```suggestion
        required=True, default=1, help="the highest priority = the smallest number"
```

base_comment_template/views/base_comment_template_view.xml
```diff
@@ -26,7 +26,6 @@
                 <field name="partner_ids" />
                 <field name="model_ids" />
                 <field name="domain" />
-                <field name="priority" />
```
joao-p-marques, line 29: Shouldn't this be added somewhere with the new name?

base_comment_template/views/base_comment_template_view.xml
```diff
@@ -59,7 +58,6 @@
                             />
                             <field name="model_ids" widget="many2many_tags" />
                             <field name="partner_ids" widget="many2many_tags" />
-                            <field name="priority" />
```
joao-p-marques, line 62: Same comment as above

base_comment_template/models/base_comment_template.py
pedrobaeza: And the proper expression is:
```suggestion
        required=True, default=1, help="The smaller number = The higher priority"
```

base_comment_template/views/base_comment_template_view.xml
victoralmau: `sequence` field it not show in search view or form view, only in tree view with `widget="handle"` right?

pedrobaeza: You can show it also as number in form. Some objects do that for a direct assignation.

joao-p-marques (approved):

base_comment_template/migrations/13.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+field_renames = [
+    ("base.comment.template", "base.comment.template", "priority", "sequence"),
```
pedrobaeza: Take care with the proper arguments, that these things are very complicated to be detected, and they are not tested in CI:
```suggestion
    ("base.comment.template", "base_comment_template", "priority", "sequence"),
```

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +16,44 @@ class CommentTemplate(models.AbstractModel):
...
+                ):
+                    record.comment_template_ids = [(4, template.id)]
 
 
 class BaseCommentTemplate(models.Model):
     """Comment templates printed on reports"""
 
     _name = "base.comment.template"
     _description = "Comments Template"
+    _order = "sequence"
```
pedrobaeza:
```suggestion
    _order = "sequence,id"
```

base_comment_template/models/base_comment_template.py
```diff
@@ -110,12 +86,14 @@ class BaseCommentTemplate(models.Model):
     )
     partner_ids = fields.Many2many(
         comodel_name="res.partner",
+        relation="res_partner_base_comment_template_rel",
+        column1="pid",
```
pedrobaeza: Why using this ugly column names. Use `partner_id`. Anyway, you must reuse the same existing relation and columns if you don't want to add migration scripts for moving data previous to this update. Anyways, you must take care also about the migration from v12 to v13.

base_comment_template/models/base_comment_template.py
```diff
@@ -124,38 +102,44 @@ class BaseCommentTemplate(models.Model):
...
     )
-
-    priority = fields.Integer(
-        default=10, copy=False, help="the highest priority = the smallest number",
+    sequence = fields.Integer(
+        required=True, default=1, help="The smaller number = The higher priority"
     )
 
-    @api.constrains("domain", "priority", "model_ids", "position")
+    @api.constrains("domain", "sequence", "model_ids", "position")
```
pedrobaeza: Why this constraint? With this new approach, you can add as much templates as you want. The `sequence` only determines the order in which they are going to be applied.

base_comment_template/models/base_comment_template.py
```diff
@@ -124,38 +102,44 @@ class BaseCommentTemplate(models.Model):
...
                         "domain and position."
                     )
                 )
 
+    def name_get(self):
+        """Redefine the name_get method to show the template name with the position.
+        """
+        res = []
+        for item in self:
+            res.append((item.id, "{} ({})".format(item.name, item.position)))
```
pedrobaeza: I'm thinking this should include the model as well (model name, not model technical name), for being selected in partners knowing to which document they applies, but only do this through a context that is injected in the partner form, not in the other views.

base_comment_template/models/res_partner.py
```diff
@@ -9,6 +9,11 @@ class ResPartner(models.Model):
 
     base_comment_template_ids = fields.Many2many(
         comodel_name="base.comment.template",
+        relation="res_partner_base_comment_template_rel",
```
pedrobaeza: The same comment than previous m2m.

base_comment_template/models/res_partner.py
```diff
@@ -9,6 +9,11 @@ class ResPartner(models.Model):
...
+        column1="tid",
+        column2="pid",
         string="Comment Templates",
         help="Specific partner comments that can be included in reports",
+        readonly=False,
```
pedrobaeza: No need of these 2 attributes, as they are the default.

base_comment_template/models/res_partner.py
```diff
@@ -9,6 +9,11 @@ class ResPartner(models.Model):
 
     base_comment_template_ids = fields.Many2many(
```
pedrobaeza, line 11: I think you should include the field in commercial fields for parent/child propagation.

base_comment_template/tests/test_base_comment_template.py
```diff
@@ -37,99 +37,84 @@ def setUpClass(cls):
...
+        ]
 
     @classmethod
     def tearDownClass(cls):
         teardown_test_model(cls.env, ResUsers)
         super(TestCommentTemplate, cls).tearDownClass()
 
+    def test_template_name_get(self):
+        self.assertEqual(
+            self.before_template_id.display_name, "before_lines (before_lines)",
```
pedrobaeza: The `name_get` should return a proper human readable text, not `before_lines`. Use the natural value (with `_fields[field_name]._get_selection` or similar method name).

pedrobaeza (changes requested):

base_comment_template/models/base_comment_template.py
victoralmau: Constraint have removed.

base_comment_template/models/base_comment_template.py
```diff
@@ -15,69 +15,44 @@ class CommentTemplate(models.AbstractModel):
...
+                [
+                    ("id", "in", partner.base_comment_template_ids.ids),
+                    ("model_ids.model", "=", self._name),
+                ]
+            )
+            for template in templates:
+                if (
+                    not template.domain
+                    or template.domain
+                    and self in self.search(safe_eval(template.domain))
```
pedrobaeza: No need of repeating conditionals. The flow is getting to the second part if there's a domain

```suggestion
                    or self in self.search(safe_eval(template.domain))
```

base_comment_template/models/base_comment_template.py
```diff
@@ -124,37 +101,29 @@ class BaseCommentTemplate(models.Model):
...
         required=True,
         default="[]",
         help="This comment template will be available only for objects "
         "that satisfy the condition",
     )
-
-    priority = fields.Integer(
-        default=10, copy=False, help="the highest priority = the smallest number",
+    sequence = fields.Integer(
+        required=True, default=1, help="The smaller number = The higher priority"
```
pedrobaeza: Ten is a good default and follows previous value.
```suggestion
        required=True, default=10, help="The smaller number = The higher priority"
```

base_comment_template/models/base_comment_template.py
```diff
@@ -124,37 +101,29 @@ class BaseCommentTemplate(models.Model):
...
+                item.name, dict(self._fields["position"].selection).get(item.position)
             )
-            if other_template_same_models_and_priority:
-                raise ValidationError(
-                    _(
-                        "There are other records with same models, priority, "
-                        "domain and position."
-                    )
-                )
+            if self.env.context.get("add_models_in_name"):
```
pedrobaeza: Use a more specific context value:
```suggestion
            if self.env.context.get("comment_template_model_display"):
```

base_comment_template/models/base_comment_template.py
```diff
@@ -124,37 +101,29 @@ class BaseCommentTemplate(models.Model):
...
+                name = "{} Apply on: {}".format(name, item.model_ids.mapped("name"))
```
pedrobaeza: This is not translatable, and not correctly spelled (it would be `Applied on:`), but let's keep it minimal:
```suggestion
                name += " (%s)" % ", ".join(item.model_ids.mapped("name"))
```

base_comment_template/models/ir_model.py
```diff
@@ -15,9 +16,10 @@ class IrModel(models.Model):
 
     def _reflect_model_params(self, model):
         vals = super(IrModel, self)._reflect_model_params(model)
-        vals["is_comment_template"] = issubclass(
-            type(model), self.pool["comment.template"]
-        )
+        vals["is_comment_template"] = (
+            (issubclass(type(model), self.pool["comment.template"]))
+            or model._name == "res.partner"
```
pedrobaeza: Why this is harcoded?

base_comment_template/models/ir_model.py
```diff
@@ -15,9 +16,10 @@ class IrModel(models.Model):
...
+        ) and model._name not in ("base.comment.template", "comment.template")
```
pedrobaeza: And why this if the operator is `issubclass`?

base_comment_template/models/res_partner.py
```diff
@@ -1,14 +1,24 @@
...
+        column1="template_id",
+        column2="partner_id",
         string="Comment Templates",
         help="Specific partner comments that can be included in reports",
     )
+
+    @api.model
+    def _commercial_fields(self):
+        """Add this field to commercial fields, as it should be propagated
+        to children.
```
pedrobaeza:
```suggestion
        """Add comment templates to commercial fields"""
```

base_comment_template/views/base_comment_template_view.xml
```diff
@@ -76,6 +76,10 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">base.comment.template</field>
         <field name="view_mode">tree,form</field>
+        <field
+            name="context"
+            eval="{'default_model_ids': [(4, ref('base.model_res_partner'))]}"
```
pedrobaeza: Why applying by default this model? This one is not going to get templates. The target are sales orders, invoices, etc.

base_comment_template/views/res_partner_view.xml
```diff
@@ -6,14 +6,21 @@
...
                     <field
-                        colspan="4"
-                        mode="tree"
                         name="base_comment_template_ids"
                         nolabel="1"
-                        widget="one2many"
-                        height="300"
-                        context="{'default_partner_id': active_id}"
-                    />
+                        context="{'add_models_in_name': True}"
```
pedrobaeza: Remember to change the context name here. Is this working putting the context at this level?

base_comment_template/models/base_comment_template.py
victoralmau: Ok thanks, i have changed it according to previous relation.

pedrobaeza:
Have you checked migration from 12.0?

victoralmau:
> Have you checked migration from 12.0?

IMO these changes not modify nothing according to `base.comment.template` model from 13.0 and in script migration https://github.com/OCA/reporting-engine/blob/13.0/base_comment_template/migrations/13.0.1.0.0/pre-migration.py everything seem correct.

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -8,11 +8,8 @@ def migrate(env, version):
     # Not tested
     properties = env["ir.property"].search(
         [
-            (
-                "fields_id",
-                "=",
-                env.ref("base.field_res_partner_property_comment_template_id").id,
-            )
+            ("fields_id.name", "=", "property_comment_template_id"),
+            ("fields_id.model_id", "=", env.ref("base.model_res_partner")),
```
pedrobaeza:
```suggestion
            ("fields_id.model_id", "=", env.ref("base.model_res_partner").id),
```

pedrobaeza: Anyways, this is very inefficient and should be done in one `INSERT INTO` SQL in the m2m table.

base_comment_template/models/comment_template.py
```diff
@@ -0,0 +1,44 @@
...
+    # the comment templates downstream.
+    # E.g.: other models where the partner field is called customer_id.
+    _comment_template_partner_field_name = "partner_id"
+
+    comment_template_ids = fields.Many2many(
+        compute="_compute_comment_template_ids",
+        comodel_name="base.comment.template",
+        string="Comment Template",
+        domain=lambda self: [("model_ids.model", "=", self._name)],
+        store=True,
```
pedrobaeza, line 26: This is missing `readonly=False` for being a computed writable.

pedrobaeza (changes requested):

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -1,28 +1,22 @@
...
-                "=",
-                env.ref("base.field_res_partner_property_comment_template_id").id,
-            )
-        ]
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO base_comment_template_res_partner_rel
+        (res_partner_id, base_comment_template_id)
+        SELECT SPLIT_PART(ip.res_id, ',', 2) AS res_partner_id,
```
pedrobaeza: Isn't this directly the ID without doing split?

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -1,28 +1,22 @@
...
+        SPLIT_PART(ip.value_reference, ',', 2) AS base_comment_template_id
```
pedrobaeza: I think you have to do a cast of this to integer:
```suggestion
        SPLIT_PART(ip.value_reference, ',', 2)::int AS base_comment_template_id
```

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -1,28 +1,22 @@
...
+        FROM ir_property ip
+        LEFT JOIN ir_model_fields imf ON ip.fields_id = imf.id
```
pedrobaeza:
```suggestion
        JOIN ir_model_fields imf ON ip.fields_id = imf.id
```

base_comment_template/migrations/13.0.1.0.0/pre-migration.py
victoralmau: Is not possible access to any ID because `ir.property` table have relations with char fields.

pedrobaeza (approved):

PR #507, opened 2021-04-30
[IMP] - use display for element visibility

sbejaoui:
the visibility attribute with the hidden option hides the element but
don't free the space.

A common use case is to have a specific header for the first page and
another for the other pages. With visibility: hidden the second header
cannot take the first one place and will keep a huge margin.
One solution is to use the display:none instead.

### DIFF
--- a/report_qweb_element_page_visibility/views/layouts.xml
+++ b/report_qweb_element_page_visibility/views/layouts.xml
@@ -32,16 +32,16 @@
                         }
                         var operations = {
                             'not-first-page': function (elt) {
-                                elt.style.visibility = (vars.sitepage === vars.frompage) ? "hidden" : "visible";
+                                elt.style.display = (vars.sitepage === vars.frompage) ? "none" : "inherit";
                             },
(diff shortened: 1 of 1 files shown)


cvinh:
@sbejaoui Thank you ! This commit is IMP for v13 but for v14, it should be a FIX because without it, the pdf reports are not acceptable (some text gets hidden) Tested with wkhtmltopdf v0.12.5-1. With old wkhtmltopdf-0.12.1-1 the text is not hidden...
@LuceEasyPME that solves your issue with v14
@OCA/reporting-engine-maintainers please merge this PR and also against v14.0

cvinh:
Hello, is there a psc member who can apply 'no stale' please ?

cvinh:
Cc @LuceEasyPme

cvinh (approved):

cvinh (approved):

HviorForgeFlow (approved):

cvinh (approved):

PR #509, opened 2021-05-25
[14.0][FIX] base_comment_template: Filter domain correctly

joao-p-marques:
1. Safe eval domain before checking if it is defined. This makes `[]` be `False` on the first check, and provides the proper use case for when the domain is empty
2. Use `filtered_domain` in **record** instead of a `search` in **self**. When on a compute, we are on a `NewId` context, so the search would always fail

@Tecnativa
TT29309

@victoralmau @pedrobaeza please review
I think it would also be needed in v13, can you check @victoralmau?

### DIFF
--- a/base_comment_template/models/comment_template.py
+++ b/base_comment_template/models/comment_template.py
@@ -39,7 +39,6 @@ class CommentTemplate(models.AbstractModel):
                 ]
             )
             for template in templates:
-                if not template.domain or self in self.search(
-                    safe_eval(template.domain)
-                ):
+                domain = safe_eval(template.domain)
+                if not domain or record.filtered_domain(domain):
                     record.comment_template_ids = [(4, template.id)]


pedrobaeza (changes requested):
Please don't safe eval. That should be the last resort. Do instead `if template.domain is False` or similar

joao-p-marques:
> Please don't safe eval. That should be the last resort. Do instead if template.domain is False or similar

The thing is that, in this case, some sort of "conversion" is needed. As I mentioned in the comment above, the check `if template.domain is False` will evaluate to `"[]" is False`, which would not be true, so it would never enter that situation. OTOH, apart from this check, we will always need to use `template.domain` to filter the recordset and check if the current element applies, so we need a way of evaluating the string domain to a python object. How could we do that?

pedrobaeza (approved):
OK, let's go that way

joao-p-marques:
> OK, let's go that way

Looking through other modules that also implement the feature of a user defined domain (as a string), seems like an `eval()` is the only option here :man_shrugging:

victoralmau (approved):

PR #514, opened 2021-05-26
[MIG] report_xml: Migration to 14.0

etobella:
Superseding due to inactivity:

#472

Just a rebase was made in order to solve the issues with the requirements.

PR #489, opened 2021-03-03
[FIX] report_xml: add context to /report/download

gaikaz:
Odoo did this change in [this commit](https://github.com/odoo/odoo/commit/4c9624db9d4656457f4100ee2e848f0f81d264a9)

PR #517, opened 2021-06-04
[14.0][FIX] report_xml

Du-ma:
Passing `context`, and taking `safe_eval` from where it's supposed to be taken from.
Also passing `time` from `safe_eval` to report name, because why not?

[](http://www.youtube.com/watch?v=Ciw-almypOU "Video Title")

### DIFF
--- a/report_xml/controllers/main.py
+++ b/report_xml/controllers/main.py
@@ -6,7 +6,8 @@ import json
 from werkzeug.urls import url_decode
 
 from odoo.http import content_disposition, request, route, serialize_exception
-from odoo.tools import html_escape, safe_eval
+from odoo.tools import html_escape
+from odoo.tools.safe_eval import safe_eval, time
 
 from odoo.addons.web.controllers import main as report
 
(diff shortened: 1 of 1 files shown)


Du-ma:
Also a thing to mention, the `context` thing was actually fixed on 13.0 #489 but, during the migration to 14.0 #514 it, together with `time` param, vanished :smile:

Du-ma:
@etobella @ozono 
Check this out lads

ozono (approved):
LGTM. Well done.

Du-ma:
@OCA/reporting-engine-maintainers 
Could this be looked at? This module isn't in a working state without this fix :smile:

pedrobaeza:
Is it safe eval strictly needed?

etobella:
It is used for attachment name, the same logic applied on odoo:
https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_actions_report.py#L202

etobella (approved):

PR #519, opened 2021-06-07
 [IMP] bi_sql_editor : conserve cron settings, when setting materialized view to draft ; Set correct default values for cron and enable it

legalsylvain:
Use case : 
- create materialized views, create models and configure the cron
- reset to draft to change request. (adding / removing fields)
- recreate models 

-> the cron settings are lost and has to be set again.

With that patch : 
- cron are kept if materialized views are set to draft
- default run values are now daily at midnight, beginning tomorrow (before was monthly with the datetime of the validation of the view that was making the cron to be refreshed immediately)
- cron is now active by default. (and disabled when view is reset to draft)

CC @quentinDupont 

apps/deck/#/board/144/card/1611

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -233,6 +233,8 @@ class BiSQLView(models.Model):
             raise UserError(
                 _("You can only unlink draft views."
                   "If you want to delete them, first set them to draft."))
+        if self.mapped("cron_id"):
+            self.mapped("cron_id").unlink()
         return super(BiSQLView, self).unlink()
 
     @api.multi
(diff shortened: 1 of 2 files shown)


etobella (approved):
Makes sense

tfossoul:
@legalsylvain Hi, I would like to have this fix in v13. What is the right approach to do it ? Can I just cherry-pick your commit (+ fix eventual migration bug) into a new v13 branch and do a pull request ?

legalsylvain:
/ocabot merge patch

> @legalsylvain Hi, I would like to have this fix in v13. What is the right approach to do it ? Can I just cherry-pick your commit (+ fix eventual migration bug) into a new v13 branch and do a pull request ?

yes it should work.

legalsylvain:
damn'it, not defined as maintainer of this module.

legalsylvain:
now it is.

/ocabot merge patch

PR #522, opened 2021-06-30
[13.0][MIG] report_qweb_signer

Laurent-Corron:


bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -237,8 +237,8 @@ def _prepare_search_filter_field(self):
         self.ensure_one()
         res = ""
         if self.field_description and self.is_group_by:
-            res = """<filter name="%s" string="%s"
-                        context="{'group_by':'%s'}"/>""" % (
+            res = """<filter name="{}" string="{}"
+                        context="{{'group_by':'{}'}}"/>""".format(
```
sbidoul, line 241: This change should not be in this PR ?

report_qweb_signer/data/defaults.xml
```diff
@@ -2,6 +2,6 @@
 <odoo>
     <record model="ir.config_parameter" id="report_qweb_signer_java_param">
         <field name="key">report_qweb_signer.java_parameters</field>
-        <field name="value">-Xms4M -Xmx4M</field>
+        <field name="value">-Xms4M -Xmx1g -XX:CompressedClassSpaceSize=256m</field>
```
sbidoul: Is this change related to the migration ?

sbidoul (commented):
Looks good.

bi_sql_editor/models/bi_sql_view_field.py
Laurent-Corron: This was changed automatically by pre-commit, I can remove it if you would like ?

report_qweb_signer/data/defaults.xml
Laurent-Corron: If I didn't add -XX:CompressedClassSpaceSize=256m to the file, I couldn't run tests (same thing for a gitlab runner)
-Xmx1g was just a test I didn't remove, changed it back to -Xmx4M

bi_sql_editor/models/bi_sql_view_field.py
sbidoul: Ah ok, it must be pyupgrade that does it. Strange that it changes in this PR and while the main branch is green ?

sbidoul:
@pedrobaeza if you are interested. There is also https://github.com/OCA/reporting-engine/pull/528 which adds optional endesive support (no java \o/).

pedrobaeza:
Thanks for the pointer.

If the Java-free solution is working, shouldn't we go directly to that one and don't make it optional?

sbidoul:
We could make endesive the default in 13.0+, and keep the java code path for a little while until people confirm they are happy with endesive.

pedrobaeza:
If there's no drawback, I think the goal of the module is very clear: to sign the PDF. Letting 2 options for doing exactly the same is maintenance-wise? If there are some differences, then let's put them on the table for deciding.

sbidoul:
I agree we'll need to keep only one at some point. But for now we can't promise that endesive has no drawbacks. It is simpler to deploy and has better performance, but there might be some exotic signature feature missing (such as the LTV option).

pedrobaeza (approved):
OK then, let's go with such approach.

Code review

sbidoul (approved):

PR #523, opened 2021-06-30
[13.0][ADD] report_py3o_signer

Laurent-Corron:
- [x] :warning:  Requires #522


sbidoul:
@Laurent-Corron can you rebase ?

setup/report_py3o_signer/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
+
+This directory caches those eggs to prevent repeated downloads.
+
+However, it is safe to delete this directory.
+
```
sbidoul: This directory must not be in git

report_py3o_signer/models/ir_actions_report.py
```diff
@@ -0,0 +1,42 @@
...
+        res = super()._check_valid_action_report()
+        res += self.report_type != "py3o"
+        return res
+
+    def render_py3o(self, res_ids, data):
+        certificate = self._certificate_get(res_ids)
+        content, filetype = super().render_py3o(res_ids, data)
+        if certificate:
+            # Creating temporary origin PDF
+            pdf_fd, pdf = tempfile.mkstemp(suffix=".pdf", prefix="report.tmp.")
```
sbidoul: It seems the deletion of the temp file is missing ?
Also have you considered using the higher level `tempfile.NamedTemporaryFile` (with delete=False) ?

report_py3o_signer/models/ir_actions_report.py
```diff
@@ -0,0 +1,42 @@
...
+                pf.write(content)
+            _logger.debug(
+                "Signing PDF document '%s' for IDs %s with certificate '%s'",
+                self.report_name,
+                res_ids,
+                certificate.name,
+            )
+            signed = self.pdf_sign(pdf, certificate)
+            # Read signed PDF
+            if os.path.exists(signed):
```
sbidoul: Shouldn't we raise an exception if signed does not exist ? 
Actually, is it possible for `pdf_sign` to return a nonexistent file ?

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -30,9 +30,13 @@ def _normalize_filepath(path):
 class IrActionsReport(models.Model):
     _inherit = "ir.actions.report"
 
+    def _check_valid_action_report(self):
+        return self.report_type != "qweb-pdf"
```
sbidoul: Returning True if valid would be easier to understand ?

report_py3o_signer/models/ir_actions_report.py
```diff
@@ -0,0 +1,42 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class IrActionReport(models.Model):
+
+    _inherit = "ir.actions.report"
+
+    def _check_valid_action_report(self):
+        res = super()._check_valid_action_report()
+        res += self.report_type != "py3o"
```
sbidoul: += on booleans is a bit weird. 

I'd say `return res or (self.report_type == "py3o")`.

sbidoul:
Can you also add a DESCRIPTION.rst and minimal USAGE.rst ?

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -30,9 +30,15 @@ def _normalize_filepath(path):
...
+    def _check_valid_action_report(self):
+        if self.report_type != "qweb-pdf":
+            return True
+        return False
+
     def _certificate_get(self, res_ids):
         """Obtain the proper certificate for the report and the conditions."""
-        if self.report_type != "qweb-pdf":
+        valid = self._check_valid_action_report()
+        if not valid:
```
sbidoul: So 
```suggestion
        if not self._is_report_type_signable():
            return False
```

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -30,9 +30,15 @@ def _normalize_filepath(path):
 class IrActionsReport(models.Model):
     _inherit = "ir.actions.report"
 
+    def _check_valid_action_report(self):
+        if self.report_type != "qweb-pdf":
+            return True
```
sbidoul: With this the behaviour is not the same as before ?

Also, I'd rename `_check_valid_action_report` to something more specific such as `_is_report_type_signable` ?

So 
```suggestion
    def _is_report_type_signable(self):
        return self.report_type == "qweb-pdf"
```

?

report_py3o_signer/models/ir_actions_report.py
```diff
@@ -0,0 +1,47 @@
...
+            # Read signed PDF
+            if os.path.exists(signed):
+                with open(signed, "rb") as pf:
+                    content = pf.read()
+            # Manual cleanup of the temporary files
+            for fname in (pdf, signed):
+                try:
+                    os.unlink(fname)
+                except (OSError, IOError):
+                    _logger.error("Error when trying to remove file %s", fname)
```
sbidoul: This is fragile, because if there is an exception before the temp file will not be removed.
This would be more robust (pseudocode)

```
tmpfile = mkstemp()
try:
    do things
finally:
    remote tempfile
```

PR #524, opened 2021-07-02
[IMP] report_qweb_signer use absolute path for certificate files

Laurent-Corron:
- [ ] :warning:  Requires #522


sbidoul:
@Laurent-Corron can you explain why this change is necessary ?

Laurent-Corron:
@sbidoul we preferred to be able to have the p12 and password file outside odoo and be able to link it with the path

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -17,16 +17,6 @@
 _logger = logging.getLogger(__name__)
 
 
-def _normalize_filepath(path):
-    path = path or ""
-    path = path.strip()
-    if not os.path.isabs(path):
```
sbidoul: but it did already support absolute paths before ? So why removing the possibility to use relative path ?

PR #528, opened 2021-07-19
[13.0] [ADD] endesive to report_qweb_signer

Laurent-Corron:
This adds endesive as an option to digitally sign documents in report_qweb_signer

PR #529, opened 2021-07-27
Update dotfiles

sbidoul:
Apply dotfiles template, and add `swig` in apt dependencies because it is required to build pykcs11, which is a dependency of endesive which we need in #528


sbidoul:
Also made a minor fix in `report_csv`'s manifest.

Fast tracking as trivial.

/ocabot merge patch

PR #532, opened 2021-08-05
[13.0] Remove swig dependency + Copier update

joao-p-marques:
For `13.0` branches, the `requirements.txt` file is not currently generated from the external dependencies, so we can remove it here.

Also includes a copier update to ensure the template is up to date.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.1.5
+_commit: v1.2.0
 _src_path: git+https://github.com/OCA/oca-addons-repo-template
 dependency_installation_mode: PIP
 generate_requirements_txt: true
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why is this green if the requirement is not there? No test maybe for that part?

joao-p-marques:
Indeed, it is weird... It seems like the dependency is still being installed here: https://app.travis-ci.com/github/OCA/reporting-engine/jobs/529325049#L886

It looks like we are hitting the same issue as a couple of weeks ago with https://github.com/OCA/account-payment/pull/427, where we removed/changed a dependency but the Travis build didn't reflect it... Maybe this is a bug (or a feature :sweat_smile:) we don't know about?

pedrobaeza:
Let's fast-track it for now for fixing the rest of the repos, and let's see how it goes:

/ocabot merge nobump

joao-p-marques:
Anyway, the dependency should be moved to `test_requirements.txt`, I guess... So I just added it

joao-p-marques:
The last commit didn't get picked. Let's wait and see if we see any impact without it, otherwise adding the dependency to `test-requirements` would make it be installed as before

PR #533, opened 2021-08-13
[13.0] [IMP] report_qweb_signer: allowed reports on certificate

ThomasBinsfeld:
Add a new 'Allowed reports' field on the certificate. Only reports listed in this field can be signed. No report means all reports are allowed.

### DIFF
--- a/report_qweb_signer/README.rst
+++ b/report_qweb_signer/README.rst
@@ -23,7 +23,7 @@ Qweb PDF reports signer
     :target: https://runbot.odoo-community.org/runbot/143/13.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module extends the functionality of report module to sign
 PDFs using a PKCS#12 certificate.
--- a/report_qweb_signer/models/ir_actions_report.py
+++ b/report_qweb_signer/models/ir_actions_report.py
@@ -45,7 +45,13 @@ class IrActionsReport(models.Model):
             if "company_id" in obj:
                 company_id = obj.company_id.id or company_id
         certificates = self.env["report.certificate"].search(
-            [("company_id", "=", company_id), ("model_id", "=", self.model)]
+            [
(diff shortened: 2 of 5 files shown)


pedrobaeza:
I suppose you need this because not all the reports of the selected model should be signed. Why not taking the occasion to add the whole configuration at report level (if to be signed or not, the domain, etc...)?

ThomasBinsfeld:
> I suppose you need this because not the reports of the selected model should be signed. 

Yes

> Why not taking the occasion to add the whole configuration at report level (if to be signed or not, the domain, etc...)?

Not sure there is a real added value to do that. Wyt?

pedrobaeza:
Well, we decentralize the configuration per document, and avoid side effects if a new report is added in the model and it gets automatically signed because we haven't excluded it.

chienandalu (approved):
Aside from @pedrobaeza suggestion I think is ok :+1:

chienandalu:
Maybe a comment can be added to the ROADMAP so we can move on with this change.

chienandalu:
@ThomasBinsfeld I think this is a good addition to the module. Only a ROADMAP map with @pedrobaeza was left, I think. Will you finish it?

ThomasBinsfeld:
> @ThomasBinsfeld I think this is a good addition to the module. Only a ROADMAP map with @pedrobaeza was left, I think. Will you finish it?

Thanks for the review @chienandalu 

I rebased the branch and solved the conflicts. This PR is in production for more than a year and is ready to be merged IMO.

I added @pedrobaeza 's refactoring idea in the ROADMAP since i have neither the time nor the intention to implement it.

pedrobaeza:
Hi, Thomas, you should add it in readme/ROADMAP.rst, not in README.rst

ThomasBinsfeld:
> Hi, Thomas, you should add it in readme/ROADMAP.rst, not in README.rst

Fixed, thanks

chienandalu:
Will you forward it to latter versions?

ThomasBinsfeld:
> Will you forward it to latter versions?

Here it is https://github.com/OCA/reporting-engine/pull/668

PR #534, opened 2021-08-30
[14.0][MIG] report_qweb_signer

omar7r:


chienandalu (commented):
Hi @omar7r please include #542

pedrobaeza:
There has been also some moves in v13 version to add (and planning to switch to) a new Python library.

chienandalu:
> There has been also some moves in v13 version to add (and planning to switch to) a new Python library.

At the time I didn't find a library mature enough to stick to, although the current implementation isn't very neat anyway...

pedrobaeza:
Please see #528. It seems this one is promising. @sbidoul how were your tests? Should we remove java library in favor of this one?

sbidoul:
@pedrobaeza @chienandalu We have positive results so far with endesive but no large scale production feedback yet. So its probably premature to drop the java code yet.

chienandalu:
Oh, yeah. That the one I had spotted but two years ago it was fairly new

pedrobaeza:
Ok, then at least this PR should be updated with latest 13.0 commits

sbidoul:
And there is also #533 that is pending.

pedrobaeza:
@omar7r did you check about the new sign method? Travis is red by the way.

omar7r:
Yes, I tested two options, both works fine. I think that Travis fails because needs to have installed the package swig in the system, where can I add it?

pedrobaeza:
You can put it on `requirements.txt` file in the root folder.

pedrobaeza:
Should we remove Java library then and leave only the Pythonic one?

chienandalu:
Out of curiosity, how do both methods compare in performance when printing a big number of signed documents?

omar7r:
Swig it's a dependency in the machine not at python level.

In our case, we are using java's library in production, because the project was a migration, from 10.0 to 14.0, I don't test the python library printing several invoices at time. I don't know if someone use the pythonic lib in production and could give feedback, if not, it will be a good test to do, before the decision.

pedrobaeza:
Then you need to change .travis.yml to install such dependency:

https://github.com/OCA/reporting-engine/blob/13.0/.travis.yml#L15

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -39,9 +38,14 @@ def _certificate_get(self, res_ids):
         """Obtain the proper certificate for the report and the conditions."""
         if self.report_type != "qweb-pdf":
             return False
+        company_id = self.env.user.company_id.id
```
ernestotejeda: Shouldn't be this self.env.company ?

ernestotejeda (commented):
On runbot:
Peek 07-02-2022 17-09

ernestotejeda:
Hello @omar7r , are you working on this PR? If not, I can supersede it to speed up the migration, thanks.

omar7r:
Hi @ernestotejeda 

Yes, but I didn't have much time since your comment, on friday I think I can review it.

omar7r:
@ernestotejeda I tested it again locally and it works, maybe Runbot doesn't have enough memory for running Java.

.travis.yml
```diff
@@ -13,6 +13,7 @@ addons:
     packages:
       - expect-dev # provides unbuffer utility
       - graphviz
+      - "swig"
```
chienandalu, line 16: Mm, why quoted?

omar7r: It is like this in 13.0 version

chienandalu (approved):
Ok, perfect :) Can you squash the migration commits?

ernestotejeda (approved):
tested locally 👍🏼

PR #536, opened 2021-09-17
[13.0][IMP][FIX] report_xlsx: handle duplicate name exceptions

chienandalu:
We want to avoid duplicated sheet names exceptions the same following the same
philosophy that Odoo implements overriding the main library to avoid the 31
characters limit triming the strings before sending them to the library.

In some cases, there's not much control over this as the reports send automated
data and the potential exception is hidden underneath making it hard to debug
the original issue. Even so, different names can become the same one as their
strings are trimmed to those 31 character limit.

This way, once we come across with a duplicated, we set that final 3 characters
with a sequence that we evaluate on the fly. So for instance:

<pre>
- 'Sheet name' will be 'Sheet name~01'
- The next 'Sheet name' will try to rename to 'Sheet name~01' as well and then
  that will give us 'Sheet name~02'.
- And the next 'Sheet name' will try to rename to 'Sheet name~01' and then to
  'Sheet name~02'` and finally it will be able to 'Sheet name~03'.
- An so on as many times as duplicated sheet names come to the workbook up to
  100 for each sheet name. We set such limit as we don't want to truncate the
  strings too much and keeping in mind that this issue don't usually ocurrs.
</pre>

cc @Tecnativa TT31938

please take a look @carlosdauden @pedrobaeza

### DIFF
--- a/report_xlsx/report/report_abstract_xlsx.py
+++ b/report_xlsx/report/report_abstract_xlsx.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import logging
+import re
 from io import BytesIO
 
 from odoo import models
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

carlosdauden (approved):

report_xlsx/report/report_abstract_xlsx.py
```diff
@@ -10,6 +11,56 @@
...
+
+            In some cases, there's not much control over this as the reports send
+            automated data and the potential exception is hidden underneath making it
+            hard to debug the original issue. Even so, different names can become the
+            same one as their strings are trimmed to those 31 character limit.
+
+            This way, once we come across with a duplicated, we set that final 3
+            characters with a sequence that we evaluate on the fly. So for instance:
+
+            - 'Sheet name' will be 'Sheet name~01'
```
etobella, line 30: Just wondering, with this change, the sheet will be named always this way or only when you choose it?

In the first case, that might be complicated, as you are changing the current output of the document.

etobella (approved):

pedrobaeza:
Please fw-port it to v14

PR #541, opened 2021-09-27
More precisions regarding KPI module

### LINKED ISSUE #520: Precision regarding difference between KPI and KPI Dashboard
Hi everyone,

sorry for this probably weird question.
I would like to know the difference between Kpi and Kpi dashboard.
I mean what are the recommandation and usages differences.
Both seems to be able to set up min and max value through code or config, both deal with periodical updates and store the values etc..
Any advice and clarifications would be appreciated.
Regards

flotho:
close #520

### DIFF
--- a/kpi_dashboard/README.rst
+++ b/kpi_dashboard/README.rst
@@ -79,12 +79,6 @@ If you spotted it first, help us smashing it by providing a detailed and welcome
 
 Do not contact contributors directly about support or help with technical issues.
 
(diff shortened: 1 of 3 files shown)


legalsylvain (changes requested):
Please change text in one of the files présent in the readme folder. (Like DEVELOP.rst or ROADMAP.rst)
Thé current file is autogenerated.

legalsylvain (approved):

PR #542, opened 2021-10-01
[FIX] report_qweb_signer: noupdate defaults

chienandalu:
We don't want our custom parameters be reset every time the module
updates.

cc @Tecnativa TT32256

### DIFF
--- a/report_qweb_signer/__manifest__.py
+++ b/report_qweb_signer/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Qweb PDF reports signer",
     "summary": "Sign Qweb PDFs usign a PKCS#12 certificate",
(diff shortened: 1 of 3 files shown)


sergio-teruel (approved):

pedrobaeza:
True, but you need a migration script for existing databases changing such flag.

carlosdauden (approved):
Create migration script

chienandalu:
Added migration script

sbidoul:
@ThomasBinsfeld

ThomasBinsfeld (approved):

pedrobaeza:
I suppose this should be fw-ported to v13.

chienandalu:
Forwarded to #543

PR #544, opened 2021-10-03
[FIX-OU] report_py3o: Missing migration script

etobella:
Binary fields on 13 are by default attachments:

https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1900

It is different on 12.0:

https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L1853

So, we need to overwrite it in order to fix it.

### DIFF
--- a/report_py3o/migrations/13.0.1.0.0/post-migration.py
+++ b/report_py3o/migrations/13.0.1.0.0/post-migration.py
@@ -1,0 +1,14 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr, """SELECT id, py3o_template_data FROM py3o_template"""
+    )
+    templates = env.cr.fetchall()
+    for template_id, data in templates:
+        template = env["py3o.template"].browse(template_id)
+        template.write({"py3o_template_data": bytes(data)})


JordiBForgeFlow (approved):

etobella:
@lmignon As original migrator of the module to 13.0, may I ask for your review?

report_py3o/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr, """SELECT id, py3o_template_data FROM py3o_template"""
+    )
+    templates = env.cr.fetchall()
+    for template_id, data in templates:
+        template = env["py3o.template"].browse(template_id)
+        template.write({"py3o_template_data": bytes(data)})
```
lmignon, line 14: Should we drop the `py3o_template_data` column from the `py3o_template` table? Nevertheless, it seems that this kind of task into the migration is covered by the Odoo's migration scripts.

etobella: I prefer to not drop columns as they could be used later on other scripts.
I think that with clean database we could find this lost fields

pedrobaeza: OpenUpgrade scripts never drop columns as one of the principles of its design. As said by @etobella, it can be later used by other scripts or module. And if not, you can pass `database_cleanup` when you want for removing them.

lmignon (approved):

PR #546, opened 2021-10-20
[11.0][ADD] report_py3o

onurugur:


onurugur:
I downgraded for odoo 11 report_py3o from odoo12 .

PR #550, opened 2021-11-26
[15.0][MIG] Migrate report_xml to 15.0

aisopuro:
Last 2 commits have all the changes.

The most notable change is the unfortunate copying of JavaScript code from Odoo core into this module in order to properly handle report actions of type `qweb-xml`. I couldn't figure out a better way, since the core action service uses a closure to hide the private methods I need to call.

My suspicion is that the only way to do this cleanly is to have the core service expose the necessary methods publically. I don't unfortunately have the time to make such a proposal.

I am very open to anyone with better JavaScript skills to propose an alternate solution if they can think of one.

Ping #545


report_xml/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+    "installable": True,
+    "application": False,
+    "summary": "Allow to generate XML reports",
+    "depends": ["web"],
+    "data": [
+        "views/ir_actions_report_view.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "report_xml/static/src/js/report/action_manager_report.js",
```
lytruongduan2020: action_manager_report.esm.js ?

aisopuro: Whoops :sweat: , fixed

gaikaz (approved):
Apart from the JS overrides, LGTM
Also checked the code if I could come up with a better way to go around the private methods and couldn't :(
I suppose, this is ok then.

Mantux11 (approved):

report_xml/static/src/js/report/action_manager_report.esm.js
```diff
@@ -0,0 +1,66 @@
...
+            // Build a query string with `action.data` (it's the place where reports
+            // using a wizard to customize the output traditionally put their options)
+            const options_ = encodeURIComponent(JSON.stringify(action.data));
+            const context_ = encodeURIComponent(JSON.stringify(actionContext));
+            url_ += `?options=${options_}&context=${context_}`;
+        } else {
+            if (actionContext.active_ids) {
+                url_ += `/${actionContext.active_ids.join(",")}`;
+            }
+            if (type === "html") {
```
ChrisOForgeFlow: Please change to xml, to ensure that you give context keys to report, like tz, lang, etc

aisopuro:
The runbot build seems to be failing, though I don't know why. I tried to open the detailed logs but got a 404. Do any of the reviewers (@lytruongduan2020 @gaikaz @Mantux11 ) know if there's anything I can do to fix it?

gaikaz:
> The runbot build seems to be failing, though I don't know why. I tried to open the detailed logs but got a 404. Do any of the reviewers (@lytruongduan2020 @gaikaz @Mantux11 ) know if there's anything I can do to fix it?

Yeah... I'm not sure about runbot failing. Tho, tests on Github do pass, which I personally trust more these days 😄 

Let's wait for some maintainer opinion or someone in the know.

report_xml/static/src/js/report/action_manager_report.esm.js
aisopuro: Changed.

gaikaz:
@aisopuro 
Could you please rebase again, so that the conflicts go away?
This could also potentially solve the runbot failing.

hegenator:
I rebased the branch since @aisopuro is currently on vacation.

It seems like the pre-commit job is now failing. Am I correct in assuming it is because my rebase already contained the change to the `requirements.txt` that it attempts to apply? Should I drop the `black, isort, prettier` commit and push again?

Looks like the runbot is also still failing.

hegenator:
Fixed the pre-commit. The branch is now rebased and otherwise in the same state as it was previously @gaikaz

gaikaz:
@OCA/tools-maintainers 
Could we please get a maintainer to review this? 🙂

Also, the ci/runbot wouldn't allow to see the logs to tell why it's failing.
Tho, this (new to me) runbo**a**t does pass as green. Is that an equivalent runner? An alternative to runbot?

sebalix (approved):
@gaikaz yes, `runboat` is an alternative to `runbot`, I think we can fairly ignore the `ci/runbot` check here.

I don't know Owl a lot but the PR LG.

gaikaz:
I just looked up the runboat. Love it already. @sbidoul is a CI rockstar!

gaikaz:
@sbidoul @pedrobaeza 
Is the runbot now removed?
Could we try merging this again?

PR #555, opened 2021-12-10
[14.0][FW] report_xlsx: multiple ports from 12.0, 13.0

nguyenhk99:
Port of the following PRs from 13.0 to 14.0:
* #335 
* #471

PR #560, opened 2021-12-14
[15.0][FW] [14.0][FW] report_xlsx: multiple ports from 12.0, 13.0

sebalix:
Port of #555 from 14.0 to 15.0.
The commit https://github.com/OCA/reporting-engine/pull/555/commits/cf040010296feede0432df96942bae18e7884e48 wasn't ported as the JS implementation has changed starting from v15, I suppose it is not needed anymore.

### DIFF
--- a/report_xlsx/report/report_partner_xlsx.py
+++ b/report_xlsx/report/report_partner_xlsx.py
@@ -11,8 +11,6 @@ class PartnerXlsx(models.AbstractModel):
 
     def generate_xlsx_report(self, workbook, data, partners):
         sheet = workbook.add_worksheet("Report")
-        i = 0
-        for obj in partners:
+        for i, obj in enumerate(partners):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
And what about the context propagation one? I know that the code is now different with OWL, but is the problem still there?

sebalix:
@pedrobaeza honestly I don't know, I checked the code and saw that the new implementation is playing with the context already, so I supposed the issue is not there...

EDIT: The new code is like this, this `download` function replaces the `session.get_file` from v14 I guess, and the context is already sent in the `data` parameter:
```js
                await download({
                    url: "/report/download",
                    data: {
                        data: JSON.stringify([url, action.report_type]),
                        context: JSON.stringify(env.services.user.context),
                    },
                });

```

PR #557, opened 2021-12-13
[15.0][MIG] bi_sql_editor

mariadforgeflow:
Standard migration to v15
@ForgeFlow

PR #567, opened 2021-12-17
[IMP][12.0] bi_sql_editor : use ace widget

legalsylvain:
## Before

- display : 
image
- "tab" doesn't work.

## After

- display : 
image
- "tab" key works !

CC : 
- other contributors : @tfossoul, @vnahaulogy, @hbrunn
- @mariadforgeflow could you include this trivial change in your V15 PR ? https://github.com/OCA/reporting-engine/pull/557
- @quentinDupont 

Note : unfortunately, In V12 (at least) it is not possible to use syntaxic color with something like ``options="{'mode': 'sql'}"`` (unlike for xml, because ace javascript lib doesn't seems to be complete in ``web`` module)

### DIFF
--- a/bi_sql_editor/views/view_bi_sql_view.xml
+++ b/bi_sql_editor/views/view_bi_sql_view.xml
@@ -69,7 +69,7 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
                     </group>
                     <notebook>
                         <page string="SQL Query">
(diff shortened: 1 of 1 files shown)


mariadforgeflow:
@legalsylvain done!

houssine78 (approved):

hparfr (approved):

legalsylvain:
thanks for your review !

florian-dacosta (approved):

rvalyi (approved):

PR #573, opened 2022-01-07
[14.0][FIX] report_qweb_encrypt: conflict with bank statement

ps-tubtim:
This pull request is used to fix an error when **Validate** a bank statement.

### Step to error:

1. Install account and report_qweb_encrypt
2. Create and reconcile a bank statement
3. **Validate** a bank statement

Selection_082

### DIFF
--- a/report_qweb_encrypt/models/ir_actions_report.py
+++ b/report_qweb_encrypt/models/ir_actions_report.py
@@ -34,6 +34,8 @@ class IrActionsReport(models.Model):
         document, ttype = super(IrActionsReport, self)._render_qweb_pdf(
             res_ids=res_ids, data=data
         )
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
         password = self._get_pdf_password(res_ids[:1])
         document = self._encrypt_pdf(document, password)
         return document, ttype


pedrobaeza:
The PR is changing the signature, so it's not valid for a stable version, and moreover, there's a chance of singleton error in case more than one record is called for render.

ps-tubtim:
@pedrobaeza I change the solution to the problem. Could you review it again?

report_qweb_encrypt/models/ir_actions_report.py
```diff
@@ -34,6 +34,8 @@ def _render_qweb_pdf(self, res_ids=None, data=None):
         document, ttype = super(IrActionsReport, self)._render_qweb_pdf(
             res_ids=res_ids, data=data
         )
+        if not isinstance(res_ids, list):
```
pedrobaeza:
```suggestion
        if isinstance(res_ids, int):
```

pedrobaeza (commented):
I'm still not sure the problem should be fixed enough. Shouldn't the other module always pass a list?

report_qweb_encrypt/models/ir_actions_report.py
ps-tubtim: I have already updated this code.

pedrobaeza:
But what about my question?

ps-tubtim:
Most of the other modules passed values in a list, but I just found out that the bank statement is passed values in the integer.

pedrobaeza (approved):

PR #577, opened 2022-01-18
[FIX] report_xlsx: fix generation of pdf reports

MihranThalhath:
the report handler for xlsx added in report_xlsx module is always returning `true` even if the report being
generated is not xlsx. Due to this, the report handler doesn't check for other report types
at [action_service.js#L1019](https://github.com/odoo/odoo/blob/b309d3a99f8ff4a43a21c4772c344ed3250e319f/addons/web/static/src/webclient/actions/action_service.js#L1019) and thus doesn't generate any other reports. So if the converter is not xlsx,
return false instead of true.

Also, updating the docids and context in the report_routes controller only needs to be done if
the converter is xlsx and not in any other case.

### DIFF
--- a/report_xlsx/__manifest__.py
+++ b/report_xlsx/__manifest__.py
@@ -6,7 +6,7 @@
     "author": "ACSONE SA/NV," "Creu Blanca," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/reporting-engine",
     "category": "Reporting",
-    "version": "15.0.1.0.1",
+    "version": "15.0.1.0.2",
     "development_status": "Production/Stable",
     "license": "AGPL-3",
     "external_dependencies": {"python": ["xlsxwriter", "xlrd"]},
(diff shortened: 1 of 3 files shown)


MihranThalhath:
@pedrobaeza

pedrobaeza (approved):
Code review. Seems reasonable. Please ping original authors of the migration.

MihranThalhath:
@JoanMForgeFlow @JordiBForgeFlow @AaronHForgeFlow

JoanMForgeFlow (approved):
Code review LG! Thanks for the fixing.
@ChrisOForgeFlow what do you think?

ChrisOForgeFlow:
> Code review LG! Thanks for the fixing. @ChrisOForgeFlow what do you think?
LGTM, Thanks for fix

ChrisOForgeFlow (approved):

AaronHForgeFlow (approved):

AaronHForgeFlow (approved):

Issue #586, opened 2022-02-16
report_xlsx: report run in su mode


sbidoul:
## Module
report_xlsx

## Describe the bug
The method `generate_xlsx_report` ends being called with su=True. It should not as the reports ends up running without record rules.

This because it is invoked via `_render_xlsx` on the `ir.action.report` object, which itself comes from `_get_report_from_name` which looks up the action with sudo.

pedrobaeza:
Patch merged, so closing.

PR #587, opened 2022-02-16
[FIX] report_xlsx running report in su mode

xavier-bouquiaux:
#586 
problem explained in the issue

### DIFF
--- a/report_xlsx/models/ir_report.py
+++ b/report_xlsx/models/ir_report.py
@@ -18,10 +18,10 @@ class ReportAction(models.Model):
         report_model = self.env.get(report_model_name)
         if report_model is None:
             raise UserError(_("%s model was not found") % report_model_name)
-        return report_model.with_context(
-            active_model=self.model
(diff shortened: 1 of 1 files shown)


sbidoul (approved):
This imitates what Odoo [does](https://github.com/odoo/odoo/blob/4c403cb87a0c6e88795231e4ba77935771474aaa/odoo/addons/base/models/ir_actions_report.py#L876).

sbidoul:
@luc-demeyer you may be interested in this fix.

Cedric-Pigeon (approved):

sbidoul:
Note this may have backward compatibility implication if people have reports that relied accidentally on running as sudo. But this has to be fixed.

This bug appeared in v14, where Odoo added a sudo() in `_get_report_from_name()`. So v13 is ok. This will have to be ported to v15.

pedrobaeza:
Please fw-port it to v15:

/ocabot merge patch

PR #589, opened 2022-02-21
Update dotfiles, test with GitHub actions

sbidoul:


sbidoul:
By adding `deb` external dependencies in the manifest, tests pass and everything gets installed on ruboat.

pedrobaeza:
Is such key something customized for runboat?

sbidoul:
> Is such key something customized for runboat?

And running the tests. It's part of the [oca-ci image installation script](https://github.com/OCA/oca-ci/blob/9eff1d7235a52bde738b6ed1503e766a38f6dd70/bin/oca_install_addons#L40).

pedrobaeza:
OK, maybe we should have a how to page about CI. Merging this.

PR #594, opened 2022-03-09
10.0 backport report_async

ntsirintanis:


ntsirintanis:
@thomaspaulb here is the backport in a super draft version. Currently this is red because:

```
UserError: (u'Unable to install module "report_qweb_signer" because an external dependency is not met: Unable to find \'/usr/bin/java\' in path', '')
2022-03-09 14:13:40,840 133 CRITICAL b82b9c883-70dd-45cc-9be2-cff56fc9768d odoo.service.server: Failed to initialize database `b82b9c883-70dd-45cc-9be2-cff56fc9768d`.
  File "/opt/odoo/odoo/service/server.py", line 974, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/opt/odoo/odoo/modules/registry.py", line 83, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/odoo/modules/loading.py", line 342, in load_modules
    modules.button_install()
  File "/opt/odoo/odoo/addons/base/module/module.py", line 369, in button_install
    modules.state_update('to install', ['uninstalled'])
  File "/opt/odoo/odoo/addons/base/module/module.py", line 347, in state_update
    self.check_external_dependencies(module.name, newstate)
  File "/opt/odoo/odoo/addons/base/module/module.py", line 320, in check_external_dependencies
    raise UserError(msg % (module_name, e.args[0]))
UserError: (u'Unable to install module "report_qweb_signer" because an external dependency is not met: Unable to find \'/usr/bin/java\' in path', '')

```

Also, because of backporting, this now supports only qweb-pdf and qweb-html types. Tests run successfully on local, and also when playing on the ui. @gfcapalbo,  @Kiplangatdan, @lfreeke take a look too please if you feel like it.

report_async/models/report_async.py
```diff
@@ -0,0 +1,162 @@
...
+        return result
+
+    @api.model
+    @job
+    def run_report(self, docids, data, report_id, user_id):
+        report = self.env["ir.actions.report.xml"].browse(report_id)
+        # Render report
+        out_file = self.env["report"].render(report.report_name)
+        out_file = base64.b64encode(out_file)
+        file_ext = report.report_type.replace("qweb-", "")
```
ntsirintanis, line 133: this is of course a vile thing to do

report_async/models/report_async.py
```diff
@@ -0,0 +1,162 @@
...
+        self.ensure_one()
+        action = self.env.ref("queue_job.action_queue_job")
+        result = action.read()[0]
+        result["domain"] = [("id", "in", self.job_ids.ids)]
+        result["context"] = {}
+        return result
+
+    @api.model
+    @job
+    def run_report(self, docids, data, report_id, user_id):
```
thomaspaulb, line 128: So I think... that module `base_report_to_printer` is using [Javascript](https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/static/src/js/qweb_action_manager.js) to hijack the action of a report being printed on a certain selection of records, and changing printing behaviour based on the result of a call to `print_action_for_report_name`.

We may copy that JS code into this module and make it do a similar query for some flag that we set on the report record to specify that we want it to run asynchroneously. If that flag is True, then we notify the user with a popup that a printing job has been created and when finished, they'll get a download link in the mail. We then call some custom backend method which in turn calls this function through a cron job, like this:

```
report_async.with_delay().run_report(
     res["context"].get("active_ids", []), data, self.id, self._uid
)
```

We can then fully ignore the whole Report Center and its shenanigans and just use it in this way.

thomaspaulb (changes requested):

thomaspaulb:
@ntsirintanis Please close, we'll reopen later after using it for some time.

PR #597, opened 2022-03-16
[IMP] bi_sql_editor: Remove useless monkeypatch

lmignon:
On class creation by the ORM of 'manual' model, odoo now check if the table is a real table or a view to automatically set the '_auto' property on the class definition. We can therefore remove the monkey patch on 'ir.model'.

see https://github.com/odoo/odoo/commit/17c4f47b0ae7a0eec78b5555d52eeb0bb5e3cc5a

ping @MiquelRForgeFlow , @HviorForgeFlow @hbrunn

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -9,37 +9,10 @@ from psycopg2 import ProgrammingError
 
 from odoo import SUPERUSER_ID, _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import pycompat, sql, table_columns
+from odoo.tools import sql, table_columns
 from odoo.tools.safe_eval import safe_eval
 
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

legalsylvain (approved):
Nice clean !

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -12,35 +12,10 @@
 from odoo.tools import pycompat, sql, table_columns
```
MiquelRForgeFlow:
```suggestion
from odoo.tools import sql, table_columns
```

MiquelRForgeFlow (approved):

lmignon:
@hbrunn @MiquelRForgeFlow @legalsylvain Thank your for your fast review!

PR #579, opened 2022-01-26
[15.0][MIG] report_wkhtmltopdf_param: Migration to 15.0

yankinmax:

PR #603, opened 2022-03-30
[ADD] report_page_orientation

KKamaa:
So having issue with pre-commit due to click version as described [here ](https://github.com/dask/distributed/issues/6013) an proper version is required.


thomaspaulb:
@NL66278 This is the current PR including tests

report_page_orientation/views/views.xml
```diff
@@ -0,0 +1,62 @@
...
+    <template id="minimal_layout" inherit_id="web.external_layout_standard">
+        <xpath expr="//span[@class='page']" position="after">
+            <span><t t-esc="env.context.get('num_page')" /></span>
+         </xpath>
+        <xpath expr="//span[@class='topage']" position="after">
+            <span><t t-esc="env.context.get('tot_page')"/></span>
+         </xpath>
+     </template>
+     -->
+</odoo>
```
thomaspaulb: Can we remove Gio's abandoned great ideas?

report_wkhtmltopdf_param/tests/test_report_paperformat.py
```diff
@@ -23,6 +36,8 @@ def test_wkhtmltopdf_valid_parameter(self):
                 report_paperformat.update(
                     {"custom_params": [(0, 0, {"name": "--disable-smart-shrinking"})]}
                 )
+            except ParserError:
+                _logger.error("Document is empty")
```
thomaspaulb: Hey this is a change to another module, did this come from you in order to make it green? Or from elsewhere

thomaspaulb (changes requested):

report_wkhtmltopdf_param/tests/test_report_paperformat.py
KKamaa: Yeah did this to make it green.

thomaspaulb: Ok but then in which line does it throw ParserError when you run the tests?

https://github.com/OCA/reporting-engine/blob/15.0/report_wkhtmltopdf_param/models/report_paperformat.py#L41

KKamaa: so I was due to a _**"Document is Empty"**_ error this can be seen at [line 306](https://github.com/OCA/reporting-engine/runs/5760772752?check_suite_focus=true#step:8:306) from the previous build, that is  why I added the exception part. Mostly on a new demo db.

thomaspaulb: Actually, something else was going on. In the [15.0 migration](https://github.com/OCA/reporting-engine/pull/579/files) in [this commit](https://github.com/OCA/reporting-engine/commit/3638d7d0b5510da400334bfa9b0ab18a3f615197) the `content_html` array was replaced by `sample_html`, and no-one complained about it during review, maybe it was overlooked.

As a result, `_run_wkhtmltopdf` was run directly with a HTML string as the second parameter, instead of a `bodies` array of strings like expected.

So `for i, body in enumerate(bodies)` splits up the string into its 293 characters which miraculously does not lead to problems in the normal test, but when we added our code, we are overriding `_run_wkhtmltopdf` and actually trying to decode each of the bodies from HTML to a structure, and then it gives an error.

The correct solution would thus be not to change the test to deal with ParserError, but to prevent ParserError in the first place by reverting the change that was made in the module.

report_wkhtmltopdf_param/models/report_paperformat.py
```diff
@@ -36,7 +36,8 @@ def _check_recursion(self):
             report = self.env["ir.actions.report"].new(
                 {"paperformat_id": paperformat.id}
             )
-            content = report._run_wkhtmltopdf(sample_html)
+            contenthtml = [bytes(sample_html, "utf-8")]
```
thomaspaulb: Test now fails on the `bytes` object.... maybe it should be just `[sample_html]` ?

KKamaa: @thomaspaulb think we good now :smiling_face_with_tear:

thomaspaulb: Yayyy green

thomaspaulb (approved):

report_page_orientation/demo/report_page_orientation_demo.xml
```diff
@@ -0,0 +1,43 @@
...
+            <t t-call="web.external_layout">
+                <div class="page">
+                    List of Current Users in The system
+                    <ul>
+                        <li t-foreach="docs" t-as="doc">
+                            <t t-esc="doc.name" />
+                        </li>
+                    </ul>
+                    <h2>Hello this is a page</h2>
+                    <t t-foreach="range(100)" t-as="index">
```
gfcapalbo, line 29: @tom this does one portrait page (1xp), 100 rotated pages (100 x r) and closes with a portrait page (1x p)  -  the 100 number seems unnecessary.
wouldn't it be better to have test data that checks different patterns? ( different alternations, with different first pages and last pages, to uncover corner cases) 
1x p - 3 x r - 1x r
1x r - 3x p - 2 x r 


the test infrastructure is there should be easy to adapt.

report_page_orientation/models/ir_actions_report.py
```diff
@@ -0,0 +1,235 @@
+# Copyright 2022 Sunflower IT <https://www.sunflowerweb.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+import io
+import logging
+import tempfile
+from copy import deepcopy
+
```
gfcapalbo, line 8: why space?

report_page_orientation/models/ir_actions_report.py
```diff
@@ -0,0 +1,235 @@
...
+                if num_pages > 1:
+                    for _i in range(num_pages - 1):
+                        page_break.addprevious(deepcopy(morepage))
+            if sections:
+                body = lxml.html.tostring(root, encoding="unicode")
+            new_bodies.append(body)
+
+        # If no change, just return super.
+        if not bodies_with_rotation:
+            return super(IrActionsReport, self)._run_wkhtmltopdf(
```
gfcapalbo, line 123: odoo 15/ python 3X  we can write super()._run_wkhtmltopdf()  
"old" nomenclature is necessary if you need a modified "self" e.g. self.with_context()  or self.sudo()
or whatever. but in this case the more concise version is Okay.

report_page_orientation/README.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Authors
+~~~~~~~
+
+* Sunflower IT
+
+Contributors
+~~~~~~~~~~~~
+
+* Giovanni Capalbo <giovanni@therp.nl>
```
gfcapalbo, line 73: @thomaspaulb i should not be first author. 
made core idea , but the "placeholder" substitution, the wkhtml2pdf improvements, AND most importantly the the JS subst solution  are yours. keep alphabetical IMO.

report_page_orientation/models/ir_actions_report.py
```diff
@@ -0,0 +1,235 @@
...
+            if (section_counter < len(sections)) and sections[section_counter].get(
+                "page", 0
+            ) == input_page_counter + 1:
+                footer_element = lxml.html.fromstring(footer)
+                subst_elements = footer_element.xpath(
+                    "//script[contains(text(), 'subst')]"
+                )
+                multi_footer = footer
+                if subst_elements:
+                    subst2 = lxml.html.fragment_fromstring(
```
gfcapalbo, line 172: killer trick.

gfcapalbo (approved):
LGTM, some minor observations. :+1:

report_page_orientation/views/views.xml
gfcapalbo: @thomaspaulb     a  "great idea" can be removed when a working idea exists.
removing this myself.

gfcapalbo: i see it was cleaned up. ignore.

report_page_orientation/README.rst
thomaspaulb: @KKamaa Could you fix this and add the both of us too, and then squash into a single commit? After that it is ready for merge.

KKamaa: @thomaspaulb added the contributors though kept it alphabetical and squashed to one commit as you mentioned

report_page_orientation/models/ir_actions_report.py
```diff
@@ -0,0 +1,235 @@
...
+                    )
+                    subst_elements[-1].addnext(subst2)
+                    multi_footer = lxml.html.tostring(
+                        footer_element, encoding="unicode"
+                    )
+
+                section = sections[section_counter]
+                num_pages = section["num_pages"]
+                section_pdf_content = super(IrActionsReport, self)._run_wkhtmltopdf(
+                    [section["content"]],
```
thomaspaulb, line 212: @gfcapalbo and me are currently investigating a case whereby relevant CSS was not included in the landscaped version, which would then be because the HTML included here does not contain any wrapping. Pending this investigation, this should not be merged.

thomaspaulb (changes requested):

PR #607, opened 2022-04-06
[15.0][FW][IMP] report_wkhtmltopdf_param: de-duplicate custom params

hailangvn:
Avoid duplicated parameters when adding custom ones
In case custom paramter (ie. --zoom with value 1.0) is already in
command args list (ie. --zoom with value 0.25), its value should be
updated with custom value (ie. 1.0) instead of having two parameters
with same name.

### DIFF
--- a/report_wkhtmltopdf_param/models/report.py
+++ b/report_wkhtmltopdf_param/models/report.py
@@ -22,8 +22,17 @@ class IrActionsReport(models.Model):
         )
 
         for param in paperformat_id.custom_params:
-            command_args.extend([param.name])
+            try:
+                value_index = command_args.index(param.name)
+            except ValueError:
+                value_index = len(command_args)
+                command_args.extend([param.name])
             if param.value:
-                command_args.extend([param.value])
+                try:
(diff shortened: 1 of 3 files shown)


flotho (approved):
code review, LGTM

report_wkhtmltopdf_param/models/report.py
```diff
@@ -22,6 +22,17 @@ def _build_wkhtmltopdf_args(
...
         for param in paperformat_id.custom_params:
+            try:
+                value_index = command_args.index(param.name)
+                if param.value:
+                    command_args[value_index + 1] = param.value
+                else:
+                    pass  # param.name is already in command_args
+                is_existing = True
+            except ValueError:
+                is_existing = False
```
etobella: Wouldn't be cleaner to remove the is_existing and do the command_args.exentd and so on inside the except?

hailangvn: Thanks @etobella for the suggestion. It has been updated.

PR #615, opened 2022-04-19
Updated from template

etobella:
From Travis to Github Actions


etobella:
I needed to upgrade report_qweb_signer (cherry-picked all commits from 14.0) and added external dependencies for some modules. It seems to work fine now

report_qweb_signer/models/ir_actions_report.py
```diff
@@ -39,9 +39,14 @@ def _certificate_get(self, res_ids):
         """Obtain the proper certificate for the report and the conditions."""
         if self.report_type != "qweb-pdf":
             return False
+        obj = self.env[self.model].browse(res_ids[0])
+        if "company_id" in obj:
+            company_id = obj.company_id.id
+        else:
+            company_id = self.env.user.company_id.id
```
pedrobaeza: `self.env.company`?

etobella: Yes, it is changed later on another commit (I just kept the original commit history from 14.0)

PR #620, opened 2022-04-28
[14.0] IMP: bi_sql_editor lang handling

gurneyalex:
Allow to call refresh materialized view with an explicit lang passed in
the context and have this get priority over the user's lang


nbessi (approved):
Nice

leemannd (approved):

legalsylvain (approved):
thanks.

PR #623, opened 2022-06-01
[REF] *: remove packages from external dependencies

fernandahf:
Because those dependencies are already part of Odoo's base dependencies.

https://github.com/odoo/odoo/blob/0aff8bb9484a23c0d875d3b12259dd4e0d51e716/requirements.txt#L38

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,5 @@ cryptography
 endesive
 py3o.formats
 py3o.template
-PyPDF2
 xlrd
 xlsxwriter
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,3 @@ cryptography
 endesive
 py3o.formats
 py3o.template
-xlrd
-xlsxwriter
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 # generated from manifests external_dependencies
 cryptography
 endesive
-lxml
 py3o.formats
 py3o.template
 PyPDF2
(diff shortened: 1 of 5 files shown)


fernandahf:
@luisg123v and @moylop260

Could you review this, please?

requirements.txt
```diff
@@ -4,6 +4,5 @@ endesive
 lxml
```
luisg123v: This and other dependencies are also Odoo base ones, e.g. `xlrd`. I'd remove all non-external dependencies in the same commit

luisg123v (changes requested):
One commit per module, please. Remember modules are migrated separately, which mean the same commit will be duplicated when migrated.

requirements.txt
fernandahf: Done

moylop260:
Similar to unpinned and unneeded dependencies:
 - https://github.com/OCA/server-tools/pull/2349

It could raise the following error:

```bash
File "odoo/__init__.py", line 84, in <module>
PyPDF2.filters.decompress = _decompress
AttributeError: module 'PyPDF2' has no attribute 'filters'
```

From the following line of code:
 - https://github.com/odoo/odoo/blob/0aff8bb9484a23c0d875d3b12259dd4e0d51e716/odoo/__init__.py#L84

Since that Odoo is compatible with the pinned version but not for latest

luisg123v:
@fernandahf  Typo in report_py3o:

PR #631, opened 2022-06-24
[FIX] report_xlsx: Fix dynamic report download

JasminSForgeFlow:
Fix dynamic report download issue.

Download failed because of the dynamic report doesn't have report action and that's why the filename set false and the report can't be downloaded

@ForgeFlow


MiquelRForgeFlow (approved):

mariadforgeflow (approved):

andreampiovesana (approved):
go

PR #636, opened 2022-07-14
[14.0][IMP] report_xlsx: add option for PDF creation

TheMule71:
Use libreoffice to create PDF reports.

### DIFF
--- a/report_xlsx/__manifest__.py
+++ b/report_xlsx/__manifest__.py
@@ -10,7 +10,11 @@
     "development_status": "Mature",
     "license": "AGPL-3",
     "depends": ["base", "web"],
-    "data": ["views/webclient_templates.xml"],
+    "external_dependencies": {
+        "python": ["xlsxwriter"],
+        "deb": ["libreoffice"],
+    },
+    "data": ["views/webclient_templates.xml", "views/ir_actions_report.xml"],
     "demo": ["demo/report.xml"],
     "installable": True,
 }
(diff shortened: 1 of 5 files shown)


andreampiovesana (approved):
good

etobella:
I might prefer to have this on separate module. It seems interesting, but it breaks the idea of the original module.

alepit (approved):

MarcoCalcagni (approved):
LGTM

etobella (changes requested):
This PR is adding some extra dependencies on a module and it is not necessary. You need to add a new module that adds this behavior

andreampiovesana:
> This PR is adding some extra dependencies on a module and it is not necessary. You need to add a new module that adds this behavior

not really, this feature is requested if you use xlsxwriter for all your reports.
It is much much faster than qweb+wkhtmtopdf (not maintained) , specially for report > 100 pages
and you can debug. Libreoffice headless is used also in report_py3o for conversion (where I copied the functionality)
If you not enable xlsx_to_pdf the flow is exactly the same before.

marcelofrare (approved):

pedrobaeza (changes requested):
Indeed you can't change dependencies on the middle of a stable version, and more with such heavy ones like full `libreoffice`. Do it in an extra module that adds such dependencies.

primes2h:
I agree that this nice feature should be in a separate module. 
It works on Linux system only, so it would restrict the scope of use of the original module.

ruben-tonetto (approved):
Ok

dcorio:
Please, put this in a separate module. Those deps cannot be installed everywhere (eg: odoo.sh?)

andreampiovesana:
close for new version

PR #639, opened 2022-07-26
[FIX] kpi_dashboard: Meter widget can show actual values

### LINKED ISSUE #428: [12.0] kpi_dashboard: Meter widget max value is 100 and does not take mix/max into account
The meter widget does not take the "min" and "max" keys in the result dict into account, nor displays numbers for the "value" key, if they are bigger than 100 - then it just displays 100.

Example:

Having the following result dict:

`result = {"min": 0, "max": 600, "value": 250}`

will result in displaying just 100 in the meter widget.

This happens also in the demo data on runbot.

lijoantony:
Set min, total, used options to configure GaugeMeter to show actual values
Fixes #428

### DIFF
--- a/kpi_dashboard/static/src/js/widget/meter_widget.js
+++ b/kpi_dashboard/static/src/js/widget/meter_widget.js
@@ -17,7 +17,6 @@ odoo.define("kpi_dashboard.MeterWidget", function (require) {
         _getMeterOptions: function (values) {
             var size = Math.min(this.widget_size_x, this.widget_size_y - 40) - 10;
             return {
-                percent: values.value.value,
                 style: "Arch",
                 width: 10,
                 size: size,
@@ -25,6 +24,8 @@ odoo.define("kpi_dashboard.MeterWidget", function (require) {
                 append: values.suffix === undefined ? "" : values.suffix,
                 color: values.font_color,
                 animate_text_colors: true,
+                ...(values.value.value && {percent: values.value.value}),
+                ...values.value,
             };
         },
     });
--- a/kpi_dashboard/demo/demo_dashboard.xml
+++ b/kpi_dashboard/demo/demo_dashboard.xml
@@ -29,17 +29,19 @@
     />
     <record id="widget_meter_01" model="kpi.kpi">
         <field name="name">Meter 01</field>
-        <field name="suffix">€</field>
+        <field name="suffix">%</field>
         <field name="computation_method">code</field>
         <field name="widget">meter</field>
-        <field name="code">result = {"min": 0, "max": 100, "value": 90}</field>
+        <field name="code">result = {"value": 90}</field>
     </record>
     <record id="widget_meter_02" model="kpi.kpi">
         <field name="name">Meter 02</field>
         <field name="prefix">$</field>
         <field name="computation_method">code</field>
         <field name="widget">meter</field>
-        <field name="code">result = {"min": 0, "max": 100, "value": 40}</field>
+        <field
+            name="code"
+        >result = {"min": 0, "total": 200, "used": 90, showvalue: True}</field>
     </record>
     <function
         model="kpi.kpi"
--- a/kpi_dashboard/readme/CONFIGURE.rst
+++ b/kpi_dashboard/readme/CONFIGURE.rst
@@ -5,7 +5,7 @@ Configure KPIs
 #. Create a new KPI specifying the computation method and the kpi type
 
    #. Number: result must contain a `value` and, if needed, a `previous`
-   #. Meter: result must contain `value`, `min` and `max`
+   #. Meter: To show a percentage value, result must contain `value`. `value` must be a positive integer between 0 to 100. Meter is implemented with GaugeMeter (https://github.com/Mictronics/GaugeMeter). It can be customized by setting parameters for GaugeMeter in result. See https://github.com/Mictronics/GaugeMeter#parameter-definitions for details.
    #. Graph: result must contain a list on `graphs` containing `values`, `title` and `key`
 
 #. In order to compute the KPI you can use a predefined function from a model or


kpi_dashboard/README.rst
```diff
@@ -42,7 +42,7 @@ Configure KPIs
 #. Create a new KPI specifying the computation method and the kpi type
 
    #. Number: result must contain a `value` and, if needed, a `previous`
-   #. Meter: result must contain `value`, `min` and `max`
+   #. Meter: To display percentage values, result must contain `value`, `min` and `max`. To display actual values, result must contain `min`, `total` and `used`.
```
etobella: Don't edit here. You should edit the file on README/CONFIGURE.rst

etobella (commented):
I will try to execute runboat again tomorrow, to see how it works :wink:

kpi_dashboard/README.rst
lijoantony: Fixed. Please check.

etobella:
runboat is failing, Let me ask @sbidoul If he knows why it is happening :cry: 

Error raised:

```
+ apt-get install -qq --no-install-recommends default-jre-headless graphviz libreoffice
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/p/python3.8/libpython3.8_3.8.10-0ubuntu1~20.04.4_amd64.deb  404  Not Found [IP: 185.125.190.39 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

sbidoul:
@etobella you are hitting https://github.com/OCA/oca-ci/issues/26

etobella:
Ups! I see.... How may I help?

sbidoul:
> Ups! I see.... How may I help?

Well someone as to translate this pseudocode into proper bash. I can try to do it in the coming days, but I won't complain if someone beats me to it :)

```
deps = $(oca_list_external_dependencies deb)
if $deps:
    apt-get update
    apt-get install $deps
```

etobella:
Wow!!! :green_apple: 

Thanks @sbidoul :smile:

etobella:
@lijoantony  I am unable to start the runboat :cry: May you try to amend the commit and push force please :pray:

lijoantony:
done.

etobella:
I am unable to make it work following your comments from the readme (maybe I am doing something wrong). Can you send an image of what do you expect and the used configuration?

Also, can you modify one of the meters on the demo to show it?

kpi_dashboard/static/src/js/widget/meter_widget.js
```diff
@@ -17,7 +17,11 @@ odoo.define("kpi_dashboard.MeterWidget", function (require) {
         _getMeterOptions: function (values) {
             var size = Math.min(this.widget_size_x, this.widget_size_y - 40) - 10;
             return {
-                percent: values.value.value,
+                ...(values.value.value && {percent: values.value.value}),
+                ...(values.value.min && {min: values.value.min}),
+                ...(values.value.total && {total: values.value.total}),
```
etobella: I don't understand why do you need so many parameters.

total is for max? used for value? In that case you can simplify all, isn't it?

etobella:
I think I found what did you expect. You want the bar to show the percentage using min, total and used, isn't it?

In that case, why not using the following:

```
                ...(values.value.value && {percent: values.value.value}),
                ...(values.value.used && {showvalue: true, min: values.value.min, used: values.value.value, total: values.value.max}),
```

if used is set as True we set all the parameters

lijoantony:
Thank you for your suggestion. That is better.
I would like to make a minor modification to avoid confusion with respect to the `used` parameter, as it has it's own meaning in the context of GaugeMeter as below,

```
...(values.value.value && {percent: values.value.value}),
...(values.value.showvalue && {showvalue: true, min: values.value.min, used: values.value.value, total: values.value.max}),
```

If you agree, I will update the PR. I will also update one of the meters in the demo as you suggested.
Thank You for your review!

lijoantony:
In existing code even though `min`, `max` & `value` were documented as supported, only `value` was being used and `min` & `max` params were unused in the meter widget. So, I ended up using the existing but unused `max` parameter to switch between percentage and actual values, to avoid introducing a new parameter `showvalue`.  Demo has been updated to demonstrate both of these cases as well.

lijoantony:
@etobella Coming to think about it, there are many more parameters supported by GaugeMeter which would be of interest to the user. It would be very useful to just expose all parameters supported by the GaugeMeter to the odoo user as well (just for the meter widget),

```
...values.value,
...(values.value.value && { percent: values.value.value }),
style: 'Arch',
width: 10,
 size: size,
prepend: values.prefix === undefined ? '' : values.prefix,
append: values.suffix === undefined ? '' : values.suffix,
color: values.font_color,
animate_text_colors: true,
```
What do you think? If you think this is a good idea, I will update the PR accordingly.

etobella:
Yes, you are right, it might be interesting

lijoantony:
> Yes, you are right, it might be interesting

@etobella PR has been updated with this change

etobella:
Can you check pre-commit? @lijoantony

lijoantony:
@etobella Sorry I missed your comment. Could you please trigger the rebase again?

lijoantony:
@etobella pre-commit issues have been fixed. Drop in the coverage is not because of this PR, but from indirect changes in the file kpi_dashboard_bokeh/models/kpi_kpi.py which I am not knowledgeable to handle. Kindly merge as per your convenience. Thanks you very much for your time.

etobella (approved):

thomaspaulb:
@etobella LGTM; Since you are the maintainer, will you merge?

etobella:
There is only my review right now.

I was waiting for another one

thomaspaulb (approved):

PR #647, opened 2022-09-02
[13.0][IMP] report_xlsx: Add function to get format to be used in cells.

victoralmau:
Add function to get format to be used in cells.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT38722

### DIFF
--- a/report_xlsx/report/report_abstract_xlsx.py
+++ b/report_xlsx/report/report_abstract_xlsx.py
@@ -91,6 +91,13 @@ class ReportXlsxAbstract(models.AbstractModel):
             ids = self.env.context.get("active_ids", [])
         return self.env[self.env.context.get("active_model")].browse(ids)
 
+    def _report_xlsx_currency_format(self, currency):
+        """Get the format to be used in cells (symbol included).
+        Used in account_financial_report addon"""
+        s_before = currency.symbol if currency.position == "before" else ""
+        s_after = " %s" % currency.symbol if currency.position == "after" else ""
+        return f"{f'{s_before}'}#,##0.{'0' * currency.decimal_places}{f'{s_after}'}"
+
     def create_xlsx_report(self, docids, data):
         objs = self._get_objs_for_report(docids, data)
         file_data = BytesIO()
--- a/report_xlsx/tests/test_report.py
+++ b/report_xlsx/tests/test_report.py
@@ -15,7 +15,7 @@ except ImportError:
 
 class TestReport(common.TransactionCase):
     def setUp(self):
-        super(TestReport, self).setUp()
+        super().setUp()
         report_object = self.env["ir.actions.report"]
         self.xlsx_report = self.env["report.report_xlsx.abstract"].with_context(
             active_model="res.partner"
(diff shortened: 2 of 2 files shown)


report_xlsx/models/res_currency.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    def _get_format_amount(self):
```
chienandalu: I'd add a flag to show the symbol or not
```suggestion
    def _get_format_amount(self, symbol=True):
```

report_xlsx/models/res_currency.py
```diff
@@ -0,0 +1,17 @@
...
+        """Get the format to be used in cells (symbol included).
+        Used in account_financial_report addon"""
+        return "%s#,##%s%s" % (
+            self.symbol if self.position == "before" else "",
```
chienandalu: It isn't correct. This will render a string like
```python
'$ #,##0,00'
```
The symbol string should go between apostrophes:
```python
'"$" #,##0,00'
```

Easy way:
```python
f'"{self.symbol}"'
```

report_xlsx/tests/test_report.py
```diff
@@ -55,3 +55,7 @@ def test_id_retrieval(self):
         # Typical call from render
         objs = self.xlsx_report._get_objs_for_report(self.docs.ids, {})
         self.assertEquals(objs, self.docs)
+
+    def test_currency_format(self):
+        usd = self.env.ref("base.USD")
+        self.assertTrue(usd.symbol in usd._get_format_amount())
```
chienandalu: Better to test the expected string
```suggestion
        self.assertEqual(usd._get_format_amount(), '"$" #,##0,00')
```

report_xlsx/models/res_currency.py
chienandalu: Too generic. Maybe something like:
```suggestion
    def _report_xlsx_format_amount(self):
```

Anyway shouldn't it be in the abstract?

chienandalu (changes requested):

pedrobaeza:
I was thinking more in a method inside `report.report_xlsx.abstract` that takes the currency as argument and return the format.

victoralmau:
Changes done.

pedrobaeza:
Commit message should be more explicit: `[IMP] report_xlsx: Method for getting cell format for a currency`

report_xlsx/report/report_abstract_xlsx.py
```diff
@@ -91,6 +91,13 @@ def _get_objs_for_report(self, docids, data):
             ids = self.env.context.get("active_ids", [])
         return self.env[self.env.context.get("active_model")].browse(ids)
 
+    def _report_xlsx_format_amount(self, currency):
```
pedrobaeza:
```suggestion
    def _report_xlsx_currency_format(self, currency):
```

chienandalu (approved):

pedrobaeza (approved):

PR #658, opened 2022-10-14
[MIG] report_batch: Migration to 14.0

amh-mw:


amh-mw:
I don't run any 14.0 environments, so I am putting up this PR as a courtesy on my way to 15.0 (#659), which I will test.

PR #660, opened 2022-10-14
[MIG] report_layout_config: Migration to 15.0

amh-mw:


bosd (approved):
LGTM :+1: 
Usage instructions could be updated.
That you cannot use the header and footer images with the standard layouts.
One has to check:
`Layout images standard`
or
`Layout images clean`

bosd:
@oca/reporting-engine-maintainers Can you add the needs review label?

PR #663, opened 2022-10-18
[16.0] [MIG] report_xml

Du-ma:
Migration to `16.0`.

Since odoo 16.0 is on a theme of making the thing run faster, thought some optimization changes could be made.

Slightly reworked the controller, renamed the file to more match the current odoo file structure.
Controller got reworked in a more logical manner, looked through illogical lines, and merged some if statements.

Cleaned up the `.js` file to make it more readable (offered a similar change in [oca/server-tools/pull/2329](https://github.com/OCA/server-tools/pull/2329))

Hard-coded in some variables that are related to `xml` report (if we print a xml report it won't become a pdf out of thin air).

Of course did all the necessary changes to make the thing run.

Also what the hell is that lang "mechanism"?
Like, I get it, but why?


Du-ma:
@etobella @aisopuro @ozono 
Mind having a gander lads?

etobella (approved):
Functional review. 

Worked like a charm :smile: Thanks for the refactoring of the JS

:+1:

pedrobaeza (approved):
Tested on runboat

Du-ma:
@pedrobaeza 
Hey, I think the command didn't work for some reason :grin:

PR #669, opened 2022-10-25
[16.0][MIG] sql_request_abstract

legalsylvain:
- Add new main entry menu for SQL configuration
- Add abstract tree view
- Add abstract form view
- ADD ace/mode-pgsql.js file to have syntaxic coloration in sql field
- move has_grouped_changed from bi_sql_editor to sql_request_abstract

**Note**

indroduce ace for postgresql looks great ! See

image

### Description

This module provides an abstract model to manage SQL Select requests on database.
It is not usefull for itself. You can see an exemple of implementation in the
'sql_export' module. (same repository).

**Implemented features**

* Add some restrictions in the sql request:
    * you can only read datas. No update, deletion or creation are possible.
    * some tables are not allowed, because they could contains clear password
      or keys. For the time being ('ir_config_parameter').

* The request can be in a 'draft' or a 'SQL Valid' status. To be valid,
  the request has to be cleaned, checked and tested. All of this operations
  can be disabled in the inherited modules.

* This module two new groups:
    * SQL Request / User : Can see all the sql requests by default and execute
      them, if they are valid.
    * SQL Request / Manager : has full access on sql requests.

### DEVELOP

Inherit the model:
```
    from odoo import models

    class MyModel(models.model)
        _name = 'my.model'
        _inherit = ['sql.request.mixin']

        _sql_request_groups_relation = 'my_model_groups_rel'

        _sql_request_users_relation = 'my_model_users_rel'
```

See implementations in the modules ``bi_sql_editor`` and ``sql_export``. (same OCA/reporting-engine repository)

PR #670, opened 2022-10-25
[16.0][MIG] bi_sql_editor (from 15.0)

legalsylvain:
- move menu and action in according view file
- use new sql_request_abstract menu entry for SQL views configuration
- Use sql_request_abstract primary tree view
- Use sql_request_abstract primary form view
- move has_grouped_changed in sql_request_abstract module
- improve demo data
- add default count value to display in pivot, if no pivot_measures are defined ;
- set widget handle to sequence field
- reorder fields in tree view and add optional hide for some fields
- simplify : make field_description field required
- improve : add optional='hide' and 'show' on tree visibility
- Cherri-pick : https://github.com/OCA/reporting-engine/pull/566

Depends on https://github.com/OCA/reporting-engine/pull/669

### Description

This module extends the functionality of reporting, to support creation
of extra custom reports.
It allows user to write a custom SQL request. (Generally, admin users)

Once written, a new model is generated, and user can map the selected field
with odoo fields.
Then user ends the process, creating new menu, action and graph view.

Technically, the module create SQL View (or materialized view, if option is
checked). Materialized view duplicates datas, but request are fastest. If
materialized view is enabled, this module will create a cron task to refresh
the data).

By default, users member of 'SQL Request / User' can see all the views.
You can specify extra groups that have the right to access to a specific view.

Warning
-------

This module is intended for technician people in a company and for Odoo integrators.

It requires the user to know SQL syntax and Odoo models.

If you don't have such skills, do not try to use this module specially on a production
environment.

Use Cases
---------

this module is interesting for the following use cases

* You want to realize technical SQL requests, that Odoo framework doesn't allow
  (For exemple, UNION with many SELECT) A typical use case is if you want to have
  Sale Orders and PoS Orders datas in a same table

* You want to customize an Odoo report, removing some useless fields and adding
  some custom ones. In that case, you can simply select the fields of the original
  report (sale.report model for exemple), and add your custom fields

* You have a lot of data, and classical SQL Views have very bad performance.
  In that case, MATERIALIZED VIEW will be a good solution to reduce display duration

### configure

* Go to Dashboard / Configuration / SQL Views

* tip your SQL request

01_sql_request

* Select the group(s) that could have access to the view

02_security_access

Optionnaly, you can add a domain.

A tipical domain in a multi company context is to write
``['|', ('company_id', '=', False), ('company_id', 'in', company_ids)]``
to make reporting depending on the current companies of the user.

* Click on the button 'Validate SQL Expression'

* Once the sql request checked, the module analyses the column of the view,
  and propose field mapping. For each field, you can decide to create an index
  and set if it will be displayed on the pivot graph as a column, a row or a
  measure.

03_field_mapping

* Click on the button 'Create SQL elements'. (this step could
  take a while, if view is materialized)

* If it's a MATERIALIZED view:

    * a cron task is created to refresh
      the view. You can so define the frequency of the refresh.
    * the size of view (and the indexes is displayed)

04_materialized_view_setting

* Finally, click on 'Create UI', to create new menu, action, graph view and
  search view.


bealdav:
Probably needs a rebase because of sql_request_abstract merged cc @legalsylvain

bealdav:
Oops @legalsylvain you probably needs to remove test-requirement.txt

legalsylvain:
> Oops @legalsylvain you probably needs to remove test-requirement.txt

Done. CI is right. Thanks for updating your review.

bealdav (approved):
Only review in 3 last commits
Tested on runboat

Thanks a lot

bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -142,41 +145,42 @@ def _compute_index_name(self):
...
+                )
+
+            vals.update(
+                {
+                    "ttype": ttype,
+                    "field_description": field_description,
+                    "many2one_model_id": many2one_model_id,
+                }
+            )
+        return super().create(vals)
```
florian-dacosta: It seems quite wrong to me, should be `vals_list` instead of `vals` right ?

florian-dacosta (changes requested):
Code review only
Just one remark @legalsylvain

bi_sql_editor/models/bi_sql_view_field.py
bealdav: Good catch

legalsylvain: thanks for your review ! 
fixed.

florian-dacosta (approved):

legalsylvain:
~I have to cherry pick https://github.com/OCA/reporting-engine/pull/566#event-8108362157~

Edit : 01/03/2022 : done.

Ready to be merged.

legalsylvain:
Hum... https://github.com/OCA/oca-github-bot/issues/226 is definitively annoying...

legalsylvain:
Hi. CI is broken for unrelated problem. Waiting for the fix of the 16.0 branch.

nilshamerlinck:
Hi @legalsylvain I believe 16.0 branch has just been fixed

PR #671, opened 2022-10-26
[16.0][MIG] sql_export (from 15.0)

legalsylvain:
- split view in two files, according OCA guidelines;
- add legalsylvain as maintainers
- use abstract tree and form views
- refactor : split demo data into two files, according OCA guidelines
- replace obsolete base.menu_reporting_dashboard by spreadsheet_dashboard entries
- update translation
- prevent usage of export with parameters, that requires extra work

**Note**  export with parameters doesn't work for the time being, due to the refactoring of ``fields_view_get`` in V16.
this should be reimplemented in another PR.


bealdav:
@legalsylvain test-requirement.txt should alseo be removed here. Thanks a lot

legalsylvain:
Done @bealdav
Could you update your review ?

Thanks !

bealdav:
Thanks a lot @legalsylvain for this huge improvement and cleaning.

2 bugs found testing it.

In draft state all is nice
image

Missing `select` string and other left string
image
Probably due to js psql syntax lib. A workaround could be to not strip the query. I think it's sufficient

bealdav:
Syntax is not correct

image

But the module considers it's correct at the validation step.

image

Finally it fails at the execution step

image

legalsylvain:
Hi @bealdav. 

1) regarding the first bug, i faced to it. I think it's due to a problem in odoo js framework, when the field become readonly. Workaround found : reload the page, and all is OK. Is it OK for you ?

2) regarding the second bug : yes, it seems that it was explicitely disabled. the last commit fixes the problem. Could you test again ?

bealdav (approved):
LGTM tested again on runboat.
OK now
Thanks a lot

florian-dacosta (approved):
LGTM and thanks for this port

PR #676, opened 2022-10-27
[IMP] report_py3o: add format datetime

hoangtrann:
This feature is available in both `12.0` and `14.0`, but not yet in `13.0`.


hoangtrann:
I'm not sure who is the maintainer for this module, so could you have a look at this @dreispt ?

Also, Github seems to have an issue today, I pushed a change to my branch to format the code but it isn't included in the PR.

hoangtrann:
@lmignon could you have a look on this? I really don't know who are the maintainers or how we would govern this repo :)

PR #677, opened 2022-11-02
[15.0][FIX] base_comment_template: Add models text field to prevent error from ir.model model

victoralmau:
Add models text field to prevent error from `ir.model` model

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40209

### DIFF
--- a/base_comment_template/migrations/15.0.1.1.0/pre-migration.py
+++ b/base_comment_template/migrations/15.0.1.1.0/pre-migration.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr, "ALTER TABLE base_comment_template ADD COLUMN IF NOT EXISTS models text"
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE base_comment_template template
+        SET models = (
+            SELECT string_agg(model.model, ',')
+            FROM base_comment_template_ir_model_rel AS rel
+            JOIN ir_model AS model ON rel.ir_model_id = model.id
+            WHERE rel.base_comment_template_id = template.id
+        )
+        """,
+    )
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -1,9 +1,10 @@
 # Copyright 2014 Guewen Baconnier (Camptocamp SA)
 # Copyright 2013-2014 Nicolas Bessi (Camptocamp SA)
 # Copyright 2020 NextERP Romania SRL
-# Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021-2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class BaseCommentTemplate(models.Model):
@@ -51,17 +52,15 @@ class BaseCommentTemplate(models.Model):
         help="If set, the comment template will be available only for the selected "
         "partner.",
     )
+    models = fields.Text(required=True)
     model_ids = fields.Many2many(
         comodel_name="ir.model",
+        compute="_compute_model_ids",
+        compute_sudo=True,
         string="IR Model",
-        ondelete="cascade",
-        domain=[
-            ("is_comment_template", "=", True),
-            ("model", "!=", "comment.template"),
-        ],
-        required=True,
         help="This comment template will be available on this models. "
         "You can see here only models allowed to set the coment template.",
+        search="_search_model_ids",
     )
     domain = fields.Char(
         string="Filter Domain",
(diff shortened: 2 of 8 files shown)


CarlosRoca13 (approved):

pedrobaeza:
This is not the idea, but the reverse: `model_ids` should be computed with sudo, and the user input is the `Char` field where models are introduced with commas.

victoralmau:
> This is not the idea, but the reverse: `model_ids` should be computed with sudo, and the user input is the `Char` field where models are introduced with commas.

Although I understand the approach you raise, I think it is important to clarify that currently the access error exists in: https://github.com/OCA/reporting-engine/blob/15.0/base_comment_template/models/comment_template.py#L25 and https://github.com/OCA/reporting-engine/blob/15.0/base_comment_template/models/comment_template.py#L38 (although this could be solved with compute_sudo=True).

pedrobaeza:
OK, understood, but the problem exists also when creating templates for a specific model, isn't it?

And for the other part, it's not OK, as you are using like, so `sale.order` will match for a template done over `sale.order.line`. Please give this another think to accommodate both needs.

victoralmau:
Changes done according to new approach.
Pre-commit error will be solved with https://github.com/OCA/reporting-engine/pull/678

victoralmau:
Ping @pedrobaeza

base_comment_template/models/base_comment_template.py
```diff
@@ -74,6 +75,33 @@ class BaseCommentTemplate(models.Model):
         required=True, default=10, help="The smaller number = The higher priority"
     )
 
+    def _get_ir_model_items(self, models):
+        return self.env["ir.model"].search(
```
pedrobaeza: No sudo here?

base_comment_template/models/comment_template.py
```diff
@@ -20,9 +20,14 @@ class CommentTemplate(models.AbstractModel):
...
         compute="_compute_comment_template_ids",
+        compute_sudo=True,
         comodel_name="base.comment.template",
         string="Comment Template",
-        domain=lambda self: [("model_ids.model", "=", self._name)],
+        # model_ids.model is not allowed because the user probably does
+        # not have read access to ir.model
+        # Using like operator is not a problem because only models with
+        # is_comment_template=True flag can exist.
+        domain=lambda self: [("models", "like", self._name)],
```
pedrobaeza: This is still incorrect, as it will match any model with a longer name, but including the model name.

pedrobaeza (changes requested):

base_comment_template/models/comment_template.py
victoralmau: What do you think about this?
```

[
    ("models", "=", self._name),
    ("models", "=like", ",%s" % self._name),
    ("models", "=like", "%s," % self._name)
]
```

`sale.order` model (for example) could appear in the following ways:
- sale.order
- other.model,sale.order
- sale.order,other.model

pedrobaeza: you have to put the `search` attribute on the computed field, and search splitting by comma.

victoralmau: I'm sorry but I don't understand how to apply what you indicate in a **domain**.

victoralmau: Ok, changes done.

base_comment_template/migrations/15.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    for item in env["base.comment.template"].sudo().search([]):
+        item.models = ",".join(item.mapped("model_ids.model"))
```
pedrobaeza: Doing this, it will trigger again the recomputation of the `model_ids` computed field, so better to write it through SQL.

base_comment_template/models/base_comment_template.py
```diff
@@ -85,3 +117,19 @@ def name_get(self):
                 name += " (%s)" % ", ".join(item.model_ids.mapped("name"))
             res.append((item.id, name))
         return res
+
+    def search(self, args, offset=0, limit=None, order=None, count=False):
```
pedrobaeza: This is not the idea, but to add the `search` attribute in the field `model_ids`, and put the domain in the other model `[("model_ids", "=", self._name)]`.

pedrobaeza (changes requested):

base_comment_template/models/base_comment_template.py
victoralmau: Ok, I think I understand you (and I understood) but that is not possible because a "basic" user does not have access to  `ir.model` model and get an ACL error causing it not to reach the `search` method.

pedrobaeza: But it doesn't reach the model, it reaches the `search` method of your field, and then you do there the `sudo` operation.

victoralmau:
Changes done.

pedrobaeza:
You are still not using the `search` attribute of the computed field. Please see for example:

https://github.com/OCA/partner-contact/blob/0b26df30c0f646674411c477ee233f019f3c545d/partner_pricelist_search/models/res_partner.py#L11

base_comment_template/migrations/15.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    for item in env["base.comment.template"].sudo().search([]):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE base_comment_template
+            SET models = %s
```
pedrobaeza: At this stage the column doesn't exist, so you have to create it.

base_comment_template/migrations/15.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    for item in env["base.comment.template"].sudo().search([]):
```
pedrobaeza: You can't use the model as well, as it's not loaded in the registry. It should be pure SQL.

victoralmau:
Now, it's ready to review again.

base_comment_template/migrations/15.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr, "ALTER TABLE base_comment_template ADD COLUMN IF NOT EXISTS models text"
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE base_comment_template template
```
pedrobaeza, line 15: Have you tried if this query is working?

victoralmau: Yes, I tested it locally.

pedrobaeza (approved):

francesco-ooops:
@victoralmau do you think this needs backporting to v14?

pedrobaeza:
No, the problem is only on v15.

PR #680, opened 2022-11-03
[15.0][MIG] report_qweb_pdf_watermark

bosd:
- Standard migration
- Updated instructions
- Added Global Company Watermark


bosd:
@kittiu Can you please review?

bosd:
@grzana12 Can you please review?

report_qweb_pdf_watermark/models/report.py
```diff
@@ -0,0 +1,135 @@
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from base64 import b64decode
+from io import BytesIO
+from logging import getLogger
+
+from PIL import Image
```
grzana12, line 7: Please add `PIL` and `PyPDF2` to external_dependencies, then it not must to be bellow try/except block

grzana12 (changes requested):

report_qweb_pdf_watermark/models/report.py
pedrobaeza: Afaik, both are requirements of Odoo itself, so no need of external dependencies nor try.

grzana12: Ok, so imo it is unnecessary to check if they are available

bosd: > Ok, so imo it is unnecessary to check if they are available

Indeed, it was an dependency on the past. But since it is a core dependency it had been removed from submodules. 
Do you want to leave an approving review?

grzana12: sure

grzana12 (approved):

report_qweb_pdf_watermark/models/report.py
bosd: To complete my answer, it was changed in this pr: https://github.com/OCA/reporting-engine/pull/623

bosd:
@CasVissers-360ERP Can you review?

CasVissers-360ERP (approved):
Functional review!

report_qweb_pdf_watermark/models/report.py
```diff
@@ -27,7 +27,13 @@
 class Report(models.Model):
     _inherit = "ir.actions.report"
 
-    pdf_watermark = fields.Binary("Watermark")
+    use_company_watermark = fields.Boolean(
```
etobella, line 30: A migration script is needed for this case

etobella (changes requested):

report_qweb_pdf_watermark/models/report.py
bosd: @etobella Could you explain why an migration script would be needed?
The field pdf_watermark still exsists. See line 43
https://github.com/OCA/reporting-engine/compare/14.0...bosd:reporting-engine:15.0-mig-report_qweb_pdf_watermark#diff-66dceb063022cd58404dbbbb4649ffec677cbc89be10c385ad5714231929a220R34

The other field are a new addition, and should not cause a problem when migrating, right?

etobella: If you migrate you will need to access all reports with watermarks in order to set the boolean to true. That is the migration needed...

bosd: Oh, now I think I understand. 
By toggling the boolean you want to migrate the exsisting watermarks on the reports to the new `use_company_watermark` function.
I don’t think we want a migration script for that 😄 

The use_company_watermark is an additional new function. 🎉 
The current functionality of the pdf_watermark which can be defined on a per report basis is kept as is.

In the current situation after migrating to V15 the watermark on these reports stay unchanged.
So the result on V15 will look exactly the same as on V14. This is what we want. 👍 


For example if we would like to automatically migrate exsisting reports to the new use_company_watermark function things will get complicated and error prone.

Example:
If this module is installed on a V14 database one could have the situation that there is a different watermark specified for a SO report then a Warehouse out report.

E.g.
An SO report could have set an pdf_watermark.
`mywatermark1.pdf`

The Warehouse delivery report could have `mywatermark2.pdf`

Migrating existing pdf watermarks to the new use_global_watermark function is becoming complicated very quickly and could lead to unwanted results when done by a migration script.
In above situation the script must decide to either set mywatermark1.pdf or mywatermark2.pdf as the global company watermark.
If there are even more then 3 watermark pdf’sor custom regexes used things get unnecessary complicated.


In that scenario, it is much easier for an human to change to the global company watermark function trough the UI for those who want to use it.

bosd: @etobella Do you agree with with above explanation of the way of working?
Can we merge this one?

bosd: ping @etobella

etobella: Ouch, I see, sorry for the misunderstanding

etobella (approved):

report_qweb_pdf_watermark/models/report.py
bosd: NP, Thanks for the review/merge!!

PR #691, opened 2022-11-30
[15.0][FIX] bi_sql_editor: Set tomorrow date correctly.

victoralmau:
Set tomorrow date correctly.

It will also be necessary to fix it in 14.0

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from psycopg2 import ProgrammingError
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

legalsylvain (approved):
Don't see what it is fixing. What is the error ?
However, syntax is lighter.  👍

victoralmau:
> Don't see what it is fixing. What is the error ? However, syntax is lighter. +1

The date was impossible: https://github.com/OCA/reporting-engine/actions/runs/3583173092/jobs/6028290242#step:8:238

pedrobaeza:
The problem happens when running around midnight, joined with the timezone nos being adjusted, leads to the test not working. This time, is better to always add a full day (24 hours) for avoiding the problem.

/ocabot merge patch

pedrobaeza:
>The date was impossible

Ah, also that one, last day of each month, the tests fail.

legalsylvain:
Thanks for the explanation.

PR #693, opened 2022-11-30
[16.0][FIX] base_comment_template: Add models text field to prevent error from ir.model model

victoralmau:
FWP from 15.0: https://github.com/OCA/reporting-engine/pull/677

Add models text field to prevent error from `ir.model` model

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40209

### DIFF
--- a/base_comment_template/migrations/16.0.1.1.0/pre-migration.py
+++ b/base_comment_template/migrations/16.0.1.1.0/pre-migration.py
@@ -1,0 +1,27 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+from odoo.tools.sql import column_exists
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not column_exists(env.cr, "base_comment_template", "models"):
+        openupgrade.logged_query(
+            env.cr,
+            "ALTER TABLE base_comment_template ADD COLUMN IF NOT EXISTS models text",
+        )
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE base_comment_template template
+            SET models = (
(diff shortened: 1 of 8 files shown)


base_comment_template/migrations/16.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
```
pedrobaeza: In this case, you have to check if the column already exists, as you may come from v15 with this patch already applied.

base_comment_template/models/base_comment_template.py
```diff
@@ -85,3 +115,10 @@ def name_get(self):
                 name += " (%s)" % ", ".join(item.model_ids.mapped("name"))
             res.append((item.id, name))
         return res
+
+    def _search_model_ids(self, operator, value):
+        # We cannot use model_ids.model in search() method to avoid recursion.
```
pedrobaeza:
```suggestion
        # We cannot use model_ids.model in search() method to avoid access errors
```

base_comment_template/views/base_comment_template_view.xml
```diff
@@ -10,7 +10,8 @@
                 <field name="position" />
                 <field name="company_id" groups="base.group_multi_company" />
                 <field name="partner_ids" />
-                <field name="model_ids" />
+                <field name="models" groups="!base.group_erp_manager" />
```
pedrobaeza: `models` should be always be visible.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #700, opened 2022-12-14
[MIG] report_substitute: Migration to 16.0

matiasperalta1:


amkarthik (approved):

atchuthan (approved):
Functional QA :+1:

HviorForgeFlow:
I would migrate it directly from v13 to v16 if it your requirement. Other contributors that may need it in previous versions will backport it.

report_substitute/demo/action_report.xml
```diff
@@ -3,7 +3,7 @@
     <template id="substitution_report">
         <div class="page">Substitution Report</div>
     </template>
-    <report
+    <record
```
HviorForgeFlow: That's not right. Here you have an example of how to transform report tags to record: I would migrate it directly from v13 to v16 if it your requirement. Other contributes that may need it in previous versions will backport it.

I also recommend you to read those wikis when migrating: 
https://github.com/OCA/maintainer-tools/wiki#migration

HviorForgeFlow (changes requested):

HviorForgeFlow (approved):

PR #701, opened 2022-12-14
[14.0][report_qweb_element_page_visibility][IMP] - use display for element visibility

cvinh:
the visibility attribute with the hidden option hides the element but don't free the space.

A common use case is to have a specific header for the first page and another for the other pages. With visibility: hidden the second header cannot take the first one place and will keep a huge margin. One solution is to use the display:none instead.

Cherry picked from https://github.com/OCA/reporting-engine/pull/507

### DIFF
--- a/report_qweb_element_page_visibility/views/layouts.xml
+++ b/report_qweb_element_page_visibility/views/layouts.xml
@@ -32,16 +32,16 @@
                         }
                         var operations = {
                             'not-first-page': function (elt) {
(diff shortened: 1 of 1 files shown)


cvinh:
@OCA/reporting-engine-maintainers can you apply the no stale label please ?

HviorForgeFlow (approved):

thomaspaulb (approved):

PR #702, opened 2022-12-21
[15.0][report_qweb_element_page_visibility][IMP] - use display for element visibility

cvinh:
the visibility attribute with the hidden option hides the element but don't free the space.

A common use case is to have a specific header for the first page and another for the other pages. With visibility: hidden the second header cannot take the first one place and will keep a huge margin. One solution is to use the display:none instead.

### DIFF
--- a/report_qweb_element_page_visibility/views/layouts.xml
+++ b/report_qweb_element_page_visibility/views/layouts.xml
@@ -33,16 +33,16 @@
 
                     var operations = {
                         'not-first-page': function (elt) {
-                            elt.style.visibility = (vars.sitepage == vars.frompage) ? "hidden" : "visible";
+                            elt.style.display = (vars.sitepage == vars.frompage) ? "hidden" : "visible";
                         },
                         'not-last-page': function (elt) {
-                            elt.style.visibility = (vars.sitepage == vars.sitepages) ? "hidden" : "visible";
+                            elt.style.display = (vars.sitepage == vars.sitepages) ? "hidden" : "visible";
                         },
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):

thomaspaulb (approved):

report_qweb_element_page_visibility/views/layouts.xml
```diff
@@ -33,16 +33,16 @@
 
                     var operations = {
                         'not-first-page': function (elt) {
-                            elt.style.visibility = (vars.sitepage == vars.frompage) ? "hidden" : "visible";
+                            elt.style.display = (vars.sitepage == vars.frompage) ? "hidden" : "visible";
```
gaikaz, line 36: `display` cannot have values `hidden` or `visible`.
It should be `none`, `inherit`, or similar.

gaikaz (commented):
How was this tested/approved?
AFAIK, this just broke the module.

report_qweb_element_page_visibility/views/layouts.xml
cvinh: @gaikaz thank you, I don't remember what happenned but the cherry pick is wrong.

cvinh: I did a new PR https://github.com/OCA/reporting-engine/pull/781

cvinh:
I did a new PR here to fix https://github.com/OCA/reporting-engine/pull/781

PR #703, opened 2022-12-28
[16.0][ADD] report_company_details_translatable

AungKoKoLin1997:
This module provides translatable company details in report header.

Odoo default function is whether the customer's language is English or other langauge, the content set in company_details is simply displayed as is. So, this may cause localization issues.
@qrtl


report_company_details_translatable/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    company_details = fields.Html(
+        help="Header text displayed at the top of all reports.",
```
yostashiro:
```suggestion
```

yostashiro:
```suggestion
```

report_company_details_translatable/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Report Comapny Details Translatable",
```
yostashiro:
```suggestion
    "name": "Report Company Details Translatable",
```

report_company_details_translatable/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
yostashiro:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

yostashiro:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

report_company_details_translatable/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to translate the company details.
```
yostashiro:
```suggestion
This module allows to translate the company details to show in PDF reports.

In some countries where the dominant language is not based on Roman alphabet, it is often necessary to be able to print the company details according to the partner's language.  In case of Japan, typically Japanese for domestic transactions, and English for international ones, for example.  This module intends to cater to this requirement.
```

yostashiro (changes requested):

yostashiro (approved):
Code review.

kakurai8 (approved):
Function is OK. (After translation, I can print in English.)

Up to information sharing.
When I change translation at Settings>General Settings>Companies->Document Layout->[Configure Document Layout], there is appear Warning like this "Warning You need to save this new record before editing the translation. Do you want to proceed?", and updated automatically. But this is the other issue, so I think no need to solve in this PR. Because of this problem, I'll add company_details field to base.view_company_form.

PR #706, opened 2022-12-30
[14.0][ADD]Report Label Documentation precisions

flotho:
FIX #705

PR #708, opened 2023-01-03
[MIG][16.0] report label

legalsylvain:
Migration of ``report_label`` module from 14.0 to 16.0

- [IMP] allow to create report.paperformat.label without creating each time report.paperformat item
- [FIX] Remove useless data that are specific Label: Agipa 114016
- [REF] simplify UI, removing custom entry for ir.actions.server
- [ADD] migration script to migrate name field
- [IMP] Add label_background_color to allow to easily define background-color for label. (That can be usefull to debug label positions)
- [DOC] update documentation and screenshot
- [IMP] Set body and html margin to 0 to to able to be predictive when designing a label sheet
- [IMP] replace style by class in the label template, reducing the size of the html code generated.
- [IMP] replace label_template by label_template_view_id on the ir.actions.server model, removing useless xml and python code. (provide migration scripts)
- [REF] Split wizard file into wizard and wizard line file, following OCA guidelines

**report paperformat label configuration**

report_paperformat_label_form

**actions server configuration**

ir_actions_server_form

**Print wizard**

report_label_wizard_form


legalsylvain:
@ivantodorovich : Question about the implementation : 

you added a field label_template (char). and so implement a function ``report_label_associated_view``. 
I was thinking to replace by a simple qweb_template_id, so it reduces the code size, and avoid to write wrong values in the field "label_template". 

Are you ok with this change? 
Is there a reason for implementing it in this way?

ivantodorovich:
To be honest I don't remember why I did it like that 😓 

Looking at it now, maybe I was inspired by core's `ir.actions.report` `report_name` field.
But I think your suggestion makes sense

---

Since you're working on it, there's something else I wanted to try.
In 16 Odoo added a `disable_shrinking` field on the `report.paperformat` model: https://github.com/odoo/odoo/commit/9e4c8d61eaa68574ebe9e69f1f13eb9216111f40

If you see in this module `ROADMAP.rst` / known issues, there's a mention about that, and a suggestion to use the OCA module `report_wkhtmltopdf_param` to work around it. With this new improvement in core, we don't need that module at all. We could just set that field ourselves, or simply expose it in the paperformat form view.

What do you thikn? Has shrinking been an issue for you, or not at all?

---

Anotther thing worth looking into after every miration is the fact that we depend on `base_automation` (also see the roadmap).

It may just not be required anymore

report_label/models/report_paperformat_label.py
```diff
@@ -3,9 +3,10 @@
 
 class ReportPaperformatLabel(models.Model):
     _name = "report.paperformat.label"
-    _inherits = {"report.paperformat": "paperformat_id"}
```
ivantodorovich: I wonder what's the motivation behind this change?

IMO this inheritance made perfect sense, no?

In fact this would be the place to set a default value for the `disable_shrinking` field as mentioned above, if the inheritance is kept

report_label/README.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Configuration
+=============
+
+Go to **Settings > Technical > Analysis > Label Paper Format** and create
+your self-adhesive label paper formats.
+
+.. image:: https://raw.githubusercontent.com/OCA/reporting-engine/14.0/report_label/static/description/configure_paperformat.png
+
+Go to **Settings > Technical > Analysis > Label Report** and create your label
```
flotho, line 43: could you please integrate this #706

flotho (changes requested):

report_label/models/report_paperformat_label.py
flotho: agree with @ivantodorovich , 
maybe we're missing the benefits of the non inheritance but keeping seems to be a better option IMHO.
Could you explain what was your idea?
Regards

report_label/README.rst
legalsylvain: hi @flotho. I did it ! see readme/CONFIGURE.rst file.

legalsylvain: (the main readme file is obsolete, it is generated automatically by the oca bot, when merging).

legalsylvain:
> Anotther thing worth looking into after every miration is the fact that we depend on base_automation (also see the roadmap).
> It may just not be required anymore

I tested again, and it seems ``base_automation`` is required. I added a text in the ROADMAP.rst : 

````
  To do when migrating in version > 16 :
  1. remove  ``base_automation`` dependency
  2. install ``report_label``
  3. install then ``base_automation``.
  If the installation of ``base_automation`` works, the dependency can be
  replaced by ``base``.

```

report_label/models/report_paperformat_label.py
legalsylvain: Basically, to have the possibility to use the same report.paperformat for many report.paperformat.label.
In the current design, when you create a report.paperformat.label, it creates automatically a report.paperformat, even if one exists and is OK.
configuring report.paperformat is one thing, configuring report.paperformat.label is another one. (even if a report.paperformat is required when we have to create report.paperformat.label).

Ex : the following paperformat label should share the same paperformat. (without margin) : 
- [1 x 1](https://www.bureau-vallee.fr/100-etiquettes-multi-usages-210x297-agipa-51257.html)
- [1 x 2](https://www.bureau-vallee.fr/200-etiquettes-multi-usages-210x148-agipa-51268.html)
- [1 x 3](https://www.bureau-vallee.fr/agipa-etiquettes-multi-emploi-300-etiquette-s-210-x-99-mm.html)
- [2 x 4](https://www.bureau-vallee.fr/800-etiquettes-multiusages-105x74-174338.html)

I tested and it doesnt seems to break anything.

legalsylvain:
> In 16 Odoo added a disable_shrinking field on the report.paperformat model: https://github.com/odoo/odoo/commit/9e4c8d61eaa68574ebe9e69f1f13eb9216111f40

Thanks ! I didn't know. It solves a lot of problem. I enabled it in the demo data. it works perfectly well. See : 
In an A4 (210mm x 297mm) I set the demo data with a label_width = 70 and label_height = 49.5. (3 x 70 = 210 and 6 x 49.5 = 297). (No margin on label / paperformat).
The result, with disable_shrinking enabled, is pixel perfect  ! [test_label_report_v16.pdf](https://github.com/OCA/reporting-engine/files/10347923/test_label_report_v16.pdf)

Note : I added a 0 padding / 0 margin on the body element, otherwise, the computation is not predictive when we design 
a label sheet.

> We could just set that field ourselves, or simply expose it in the paperformat form view.

Ow. Odoo added a field, but it's not available on the view. I made a trivial PR against V16 here : https://github.com/odoo/odoo/pull/109137 could you review it ?


> To be honest I don't remember why I did it like that

> Looking at it now, maybe I was inspired by core's ir.actions.report report_name field.
> But I think your suggestion makes sense

I replaced label_template by label_template_view_id. It allows to remove xml code (button) and the according python code for the button. (migration script provided)

flotho (approved):
thnaks

report_label/models/report_paperformat_label.py
ivantodorovich: Well, in my mind, each label kind **is** a paperformat. Only it's not a regular paperformat, it's a "label" paperformat, that contains extra information.. thus the inheritance.

So when you say the label paperformat  should be able to re-use existing paperformats, I find it a bit weird, but I see what you mean.

In any case, I'm not against it

ivantodorovich (approved):

PR #711, opened 2023-01-18
[13.0][IMP] report_csv: add encoding option

AungKoKoLin1997:
@qrtl
This PR make an enhancement for generating CSV files with encoding.

### DIFF
--- a/report_csv/models/ir_report.py
+++ b/report_csv/models/ir_report.py
@@ -9,6 +9,14 @@ class ReportAction(models.Model):
     _inherit = "ir.actions.report"
 
     report_type = fields.Selection(selection_add=[("csv", "csv")])
+    encoding = fields.Char(
+        help="Encoding to be applied to the generated CSV file. e.g. cp932"
+    )
+    encode_error_handling = fields.Selection(
+        selection=[("ignore", "Ignore"), ("replace", "Replace")],
+        help="If nothing is selected, CSV export will fail with an error message when "
+        "there is a character that fail to be encoded.",
+    )
 
     @api.model
     def render_csv(self, docids, data):
@@ -17,7 +25,11 @@ class ReportAction(models.Model):
         if report_model is None:
             raise UserError(_("%s model was not found" % report_model_name))
         return report_model.with_context(
-            {"active_model": self.model}
+            {
+                "active_model": self.model,
+                "encoding": self.encoding,
+                "encode_error_handling": self.encode_error_handling,
+            }
         ).create_csv_report(docids, data)
 
     @api.model
--- a/report_csv/readme/CONFIGURE.rst
+++ b/report_csv/readme/CONFIGURE.rst
@@ -1,0 +1,9 @@
+In case the exported CSV report should be encoded in another system than UTF-8, following
+fields of the report record (*Settings > Technical > Reports*) should be populated accordingly.
+
+* Encoding: set an encoding system (such as cp932)
+* Encode Error Handling: select 'Ignore' or 'Replace' as necessary.
+
+  * 'Ignore': in case of an encoding error, the problematic character will be removed from the exported file.
+  * 'Replace': in case of an encoding error, the problematic character will be replaced with '?' symbol.
+  * Leaving the field blank: in case of an encoding error, the report generation fails with an error message.
--- a/report_csv/readme/USAGE.rst
+++ b/report_csv/readme/USAGE.rst
@@ -36,3 +36,5 @@ A report XML record ::
         file="res_partner"
         attachment_use="False"
     />
+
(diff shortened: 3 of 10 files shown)


report_csv/models/ir_report.py
```diff
@@ -9,6 +9,9 @@ class ReportAction(models.Model):
     _inherit = "ir.actions.report"
 
     report_type = fields.Selection(selection_add=[("csv", "csv")])
+    encoding = fields.Char(
+        help="Encoding to be applied to the generated CSV file. " "e.g. cp932"
```
yostashiro:
```suggestion
        help="Encoding to be applied to the generated CSV file. e.g. cp932"
```

yostashiro (changes requested):
@AungKoKoLin1997 Can you please add simple tests with encodings with 'cp932' and something random like 'xyz'?

report_csv/report/report_csv.py
```diff
@@ -46,6 +46,9 @@ def create_csv_report(self, docids, data):
         file = csv.DictWriter(file_data, **self.csv_report_options())
         self.generate_csv_report(file, data, objs)
         file_data.seek(0)
+        encoding = self._context.get("encoding")
+        if encoding:
+            return file_data.read().encode(encoding), "csv"
```
yostashiro, line 57:
```suggestion
            try:
                return file_data.read().encode(encoding), "csv"
            except Exception as e:
                raise UserError(_("Failed to encode the data with the encoding set in the report."))
```

yostashiro (changes requested):
Missed to include this in the previous review.

yostashiro:
Please also update USAGE.rst to include a comment about encoding.

report_csv/readme/USAGE.rst
```diff
@@ -36,3 +36,25 @@ A report XML record ::
...
+    <report
+        id="partner_csv"
+        model="res.partner"
+        string="Print to CSV"
+        report_type="csv"
+        encoding="cp932"
+        name="module_name.report_name"
+        file="res_partner"
+        attachment_use="False"
+    />
```
yostashiro: I suppose the report tag doesn't take the encoding parameter.  Did you try this yourself?

We may simply add the following comment instead of a lengthy explanation:

Update `encoding` with an appropriate value (e.g. `cp932`) as necessary.

yostashiro (approved):
LGTM. 👍

report_csv/readme/USAGE.rst
AungKoKoLin1997: Sorry. You are right. We can only pass value to encoding when we create record with "ir.actions.report".

report_csv/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+In order to add encode error handing in csv report you need to choose 
+encode error handling in reports.
+
+* Go to Settings > Technical > Reports > Your csv report
+* Add encoding first
+* Choose encode error handling
+
+Note: You can't see encode error handling if you don't add encoding first.
```
yostashiro:
```suggestion
In case the exported CSV report should be encoded in another system than UTF-8, following fields of the report record (*Settings > Technical > Reports*) should be populated accordingly.

* Encoding: set an encoding system (such as cp932)
* Encode Error Handling: select 'Ignore' or 'Replace' as necessary.

  * 'Ignore': in case of an encoding error, the problematic character will be removed from the exported file.
  * 'Replace': in case of an encoding error, the problematic character will be replaced with '?' symbol.
  * Leaving the field blank: in case of an encoding error, the report generation fails with an error message.
```

yostashiro:
@AungKoKoLin1997 Can you go ahead and create PRs for the upstream versions?

report_csv/report/report_csv.py
```diff
@@ -46,7 +47,19 @@ def create_csv_report(self, docids, data):
...
         file_data.seek(0)
-        return file_data.read(), "csv"
+        encoding = self._context.get("encoding")
+        if not encoding:
+            return file_data.read(), "csv"
+        error_handling = self._context.get("encode_error_handling")
+        if error_handling:
+            return file_data.read().encode(encoding, errors=error_handling), "csv"
+        try:
+            logging.info("Encdonig__________")
```
yostashiro:
```suggestion
```

fkantelberg (approved):
Works well. Can you update the forward ports with the tests too?

thomaspaulb:
@AungKoKoLin1997 Can you squash commits? Then I'll merge.

AungKoKoLin1997:
@thomaspaulb I squashed the commits. Please merge it. Thank you.

PR #716, opened 2023-01-27
[16.0][IMP] report_xml: Added `xml_extension`

Mantux11:
Adds a way to choose the file extension for your XML file, by adding the flexibility of appending more choices. For example, instead of getting `.xml`, this allows to have the report download as `.svg` or even good old `.html` if needed. This is also very useful for some localization purposes, where some scarce software uses uncommon file extensions for their XML files, like `.ffdata` for specific accounting reports here in Lithuania.

The change is not breaking, as we set the default to be `.xml` like it was by default anyway.

### DIFF
--- a/report_xml/models/ir_actions_report.py
+++ b/report_xml/models/ir_actions_report.py
@@ -29,6 +29,10 @@ class IrActionsReport(models.Model):
         help='Add `<?xml encoding="..." version="..."?>` at the start of final report '
         "file.",
     )
+    xml_extension = fields.Char(
+        default="xml",
+        help="Extension for XML Reports, by default is `xml`",
+    )
 
     @api.model
     def _render_qweb_xml(self, report_ref, res_ids, data=None):
--- a/report_xml/controllers/report.py
+++ b/report_xml/controllers/report.py
@@ -70,7 +70,7 @@ class ReportController(report.ReportController):
                     report_name = safe_eval(
                         report.print_report_name, {"object": obj, "time": time}
                     )
-                    filename = f"{report_name}.xml"
+                    filename = f"{report_name}.{report.xml_extension}"
             else:
                 data = url_parse(url).decode_query(cls=dict)
                 if "context" in data:
@@ -80,7 +80,7 @@ class ReportController(report.ReportController):
                 response = self.report_routes(
                     reportname, converter="xml", context=context, **data
                 )
-            filename = filename or f"{report.name}.xml"
+            filename = filename or f"{report.name}.{report.xml_extension}"
             response.headers.add("Content-Disposition", content_disposition(filename))
             return response
         except Exception as e:
--- a/report_xml/tests/test_report_xml.py
+++ b/report_xml/tests/test_report_xml.py
@@ -1,12 +1,15 @@
 # Copyright 2017 Creu Blanca
 # License AGPL-3.0 or later (https://www.gnuorg/licenses/agpl).
 
+import json
+
 from lxml import etree
 
+from odoo import http
 from odoo.tests import common
 
 
-class TestXmlReport(common.TransactionCase):
+class TestXmlReport(common.HttpCase):
     def test_xml(self):
         report_object = self.env["ir.actions.report"]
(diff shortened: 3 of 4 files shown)


gaikaz (approved):

gaikaz:
@OCA/reporting-engine-maintainers 
This is coming from Via laurea, where we have an extension module exactly for this.
We now decided to suggest it to the main module here on OCA, as we see it as a useful feature for the base module.

Could we get opinions and/or more reviews?

report_xml/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnuorg/licenses/agpl.html).
 {
     "name": "XML Reports",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
```
pedrobaeza: Please don't change manually the version. It will be done by bot on merge.

report_xml/models/ir_actions_report.py
```diff
@@ -29,6 +29,12 @@ class IrActionsReport(models.Model):
         help='Add `<?xml encoding="..." version="..."?>` at the start of final report '
         "file.",
     )
+    xml_extension = fields.Selection(
```
pedrobaeza: Do it a Char field for not having to program an extension module for further ones.

pedrobaeza (changes requested):

lmignon (changes requested):
Thank you @Mantux11 for your contribution. Your proposal makes sense. Can you take into account the Pedro's comments PLZ?

gaikaz:
I didn't expect to get that quick of a response. Thank you 🙏🏻 
Changing the field from Selection to Char is a good point. It also makes it a lot more testable. Therefore, we'll also add a test for this new feature. 
@Mantux11 is currently having a time-off, but should come back to this on Monday.

HviorForgeFlow:
@Mantux11 pre-commit will go green after merging this and rebasing on your side:
https://github.com/OCA/reporting-engine/pull/724

CC @pedrobaeza

pedrobaeza:
You can rebase now.

Mantux11:
@HviorForgeFlow It's not because of `pre-commit`. I'm trying to write a test function for this implentation. and locally I couldn't truly test everything. So i'm pushing here, to see if response from requests is getting correct, and I could equal if extension is correct.

Mantux11:
@HviorForgeFlow I will do that later. Now I'm trying to write a test for my added functionality, locally I couldn't test everything.

Mantux11:
@pedrobaeza @lmignon I tried to request a re-review from both, but idk why if I request from one, then from other person request is removed :D I included test for this functionality. And also rebase is done.

report_xml/models/ir_actions_report.py
```diff
@@ -29,6 +29,10 @@ class IrActionsReport(models.Model):
         help='Add `<?xml encoding="..." version="..."?>` at the start of final report '
         "file.",
     )
+    xml_extension = fields.Char(
+        default="xml",
+        help="Extension for XML Reports, by default is `.xml`",
```
pedrobaeza:
```suggestion
        help="Extension for XML Reports, by default is `xml`",
```

pedrobaeza (approved):

pedrobaeza:
Please avoid hammering the OCA CI doing a lot of pushes, and instead, test on local, and also squash your commits into one.

Mantux11:
@pedrobaeza Sorry, that's the last push for all test code refactoring, all tests is successful as I see. Squash would be the last commit push here.

gaikaz (approved):

Mantux11:
@lmignon Could you review all changes?

DonatasV (approved):

Mantux11:
@OCA/reporting-engine-maintainers Could anyone approve and merge it? Thanks:)

pedrobaeza:
As my comments were attended, which were the @lmignon's changes requested, I'm merging:

/ocabot merge minor

Thanks for the feature.

PR #718, opened 2023-02-03
[16.0][IMP] Add back the variable feature in sql_export

florian-dacosta:
This feature was removed during migration in order to migrate the module fast.
I add it back now, with a refactore, the module now uses the nice new "properties" field and we don't need to create manual fields anymore.
The use of the sql queries using variable should be easier

@bguillot @legalsylvain


sql_export/models/sql_export.py
```diff
@@ -58,11 +37,30 @@ class SqlExport(models.Model):
         default="utf-8",
     )
 
+    def _compute_use_properties(self):
+        for rec in self:
+            if rec.query_properties_definition:
+                rec.use_properties = True
+            else:
+                rec.use_properties = False
```
legalsylvain:
```suggestion
           rec.use_properties = bool(rec.query_properties_definition)

**Review by legalsylvain — CHANGES_REQUESTED** [review]
Really nice job ! 

I didn't know `PropertiesDefinition`. Looks great and more simple. (+114 / - 223 )

Partial code review & functional tests.

**Question :** 
What is the objective of setting values in this view ? 

image

I was expected it will be the default values, in the wizard, but it's not the case. 
(But I just discover PropertiesDefinition, maybe my question is not relevant).

**2 things :** 

- could you make required field in the wizard ? if not, user can click on export, but then it raise an SQL error if one of the variable is empty. 
`psycopg2.errors.SyntaxError: syntax error at or near "false"
LINE 7: AND rel.category_id in false) TO STDOUT WITH CSV HEADER DELI...`


- could you put a variable per line ? (no blocking point, but I see it more usable).

image

**Comment by florian-dacosta** [comment]
> I didn't know `PropertiesDefinition`. Looks great and more simple. (+114 / - 223 )

It is more simple (at least once you get how properties field work, but as it is a standard odoo field now, seems ok)

> **Question :** What is the objective of setting values in this view ?

This is how properties field work. On one object, you can create the property fields and set values (we want the values on the wizard to use it in the query). And when adding the properties, Odoo will store it in the related PropertiesDefinition field. 
So on sql.export, we have the field that store the properties definition (name, type, default value...) and on the wizard sql.export.file, we have the properties field, which allow us to set different value on each export.

So the configuration of the properties have to be on the sql.file.wizard model.

> I was expected it will be the default values, in the wizard, but it's not the case. (But I just discover PropertiesDefinition, maybe my question is not relevant).

Not sure I understood what you meant

>     * could you make required field in the wizard ? if not, user can click on export, but then it raise an SQL error if one of the variable is empty.

Well, this is the major draw back with the use of this properties field. I don't think it is possible to put it as required.
I believe it is kind of acceptable because if you set a variable on your query, you always want to fill a value. I don't really see use case where you would have an optional parameter for your query...
Maybe we could raise a clean UserError on validation if one of the property is empty ? I think it would be ok this way.
Anyway, in v16, required field are not even really visibible with default theme !

>     * could you put a variable per line ? (no blocking point, but I see it more usable).

Done

**Comment by florian-dacosta** [comment]
For the required field question, IMO, we can put it in the known issues/ROADMAP that we don't manage optional fields, only required (even if not visible in the view) we raise is a property is empty, and if Odoo improve the properties field with the required property, then we'll improve it.

File: sql_export/wizard/wizard_file.py [hunk 2]
```diff
@@ -75,14 +29,12 @@ def export_sql(self):
         variable_dict = {}
         now_tz = fields.Datetime.context_timestamp(sql_export, datetime.now())
         date = now_tz.strftime(DEFAULT_SERVER_DATETIME_FORMAT)
-        if sql_export.field_ids:
-            for field in sql_export.field_ids:
-                if field.ttype == "many2one":
-                    variable_dict[field.name] = self[field.name].id
-                elif field.ttype == "many2many":
-                    variable_dict[field.name] = tuple(self[field.name].ids)
+        if self.query_properties:
```
legalsylvain: useless line, as the following line is a for.

legalsylvain:
> > I was expected it will be the default values, in the wizard, but it's not the case. (But I just discover PropertiesDefinition, maybe my question is not relevant).

> Not sure I understood what you meant

Well : step to reproduce.

- go to a sql.export item.
- click on "Configure Properties"
- set default values

image

- Save
- Go back
- click again on "Configure Properties"
-> values are lost.

image


- Run the export.
-> the values are not recovered.

florian-dacosta:
> Well : step to reproduce.
> 
>     * go to a sql.export item.
> 
>     * click on "Configure Properties"
> 
>     * set default values
> 
> 
> image


You did not set default values for the properties here, you just set a value, for the current wizard.
To set a default value on the property, you have to edit the property (the definition of the property) with the pen like in the following image : 
Capture d’écran du 2023-02-03 14-29-01

Once you have set the default value on the property, it will be filled by defaut with the execute query wizard or even if you go back on the configuration (which is exactly the same thing actually).

florian-dacosta:
Hi @legalsylvain Would you mind updating your review ?

@bguillot could you do a review on this one ?

legalsylvain:
Hi @florian-dacosta. Just tested the module. all is great, except the error when user doesn't enter all the values : 


I made a PR against akretion branch. could you take a look ?

legalsylvain (changes requested):
https://github.com/akretion/reporting-engine/pull/3

florian-dacosta:
> [akretion#3](https://github.com/akretion/reporting-engine/pull/3)

Thanks for the improvement, it is merged!

legalsylvain (approved):
Thanks !
Could you ask another review from akretion team ?

bealdav (approved):
LGTM


 ... on friday with accounting made, yeoh

enjoy

legalsylvain (commented):
It requires migration scripts. Dont you think ?

florian-dacosta:
> It requires migration scripts. Dont you think ?

More or less..
Sure, it is always nice to have a migration script, but here : 
The module is already merged for some month without this feature (so the problem of this variable feature not working has been going on for months...)
The script seems quite complicated to me
Without the migration script, installation won't break, the user will only have to create the variable himself => He needs to get familiar with the new system anyway.
The user will have Manual action on migration anyway, because we can't migrate the content of the query. Since the user has to re-write the query (or at least check it), it seems legit to me that he also has a look at the variable.
Finally, the way the variable were named before was like "x_my_field" as it was using manual fields in Odoo, and with the new system, you name it as you want to, so it is the opportunity to make a clean naming.

But mainly for the fact that one have to re-write/re-check the query anyway and the complication to do the script make me think we should not bother on this one.
What do you think @legalsylvain ?

legalsylvain:
Let's go ahead.

/ocabot merge major

PR #728, opened 2023-02-22
[16.0][FIX] email template on sql_export_mail

florian-dacosta:
@legalsylvain

### DIFF
--- a/sql_export_mail/data/mail_template.xml
+++ b/sql_export_mail/data/mail_template.xml
@@ -5,19 +5,21 @@
     <record id="sql_export_mailer" model="mail.template">
         <field name="name">SQL Export</field>
         <field name="email_from">admin@example.com</field>
-        <field name="email_to">${object.get_email_address_for_template()}</field>
-        <field name="subject">${object.name or ''}</field>
+        <field name="email_to">{{object.get_email_address_for_template()}}</field>
+        <field name="subject">{{object.name or ''}}</field>
         <field name="model_id" ref="sql_export.model_sql_export" />
         <field name="auto_delete" eval="True" />
-        <field
-            name="body_html"
-        ><![CDATA[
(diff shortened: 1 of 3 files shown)


legalsylvain (commented):
Nice catch ! 
could you provide also the migration script ?

florian-dacosta:
> Nice catch ! could you provide also the migration script ?

It  seems quite complicated since the mail template can be customized by every customer, I can't juste replace the old template content by the new one.
I'd have to keep old content but replace the variables with a regex or something ?

legalsylvain:
> It seems quite complicated since the mail template can be customized by every customer, I can't juste replace the old template content by the new one.

Hi. For that topic, the postulate of openupgrade is the following : 
A) if we don't do anything, all the templates will fail. (because the syntax changed) and user should set again all the values 
B) if we erase the ``email_to``, ``subject`` and ``email_to``, it will work, but the customized values (if any) will be lost and should be set again. (or default value will be used).

The B situation is better than the A situation.
So in Openugprade, we assume loosing "now bad" values for records set as noupdate="1".

See exemple here : 
- the noupdate.xml file for a mail.template : https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/scripts/purchase/15.0.1.2/noupdate_changes.xml#L3
- the call to the load of the noupdate.xml file in a post-migration.py file : https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/scripts/purchase/15.0.1.2/post-migration.py#L13

florian-dacosta:
@legalsylvain Thanks for the hint, I've added a migration script.
I've put it for 16.0.1.0.0 version assuming we don't want to play this migration script on some instance with v16 module already installed, as we can assume they already fixed the template themselve and don't want to erase it in this case.

legalsylvain:
Thanks ! 

/ocabot merge patch

PR #729, opened 2023-02-24
[ADD] shared custom view

gurneyalex:
This module makes the user_id field of ir.ui.view.custom optional, and
therefore allows to share the definition of a dashboard view among different
users.


ivantodorovich (approved):

yankinmax (approved):

pedrobaeza:
Maybe a more "commercial" technical name can be `dashboard_shared`.

dreispt:
Sorry, I know this is annoying, but I agree with Pedro.
And to honor the convention of starting the module name with its main dependency, I propose `board_share`.
Do you agree with the renaming @gurneyalex ?

PR #753, opened 2023-04-14
[FIX] report_xlsx : fix the reportname assignment

baptisteghesquiere:
var report_name in except statement does not exist when the except is called causing this whole ting to grind Odoo to a halt.


pedrobaeza (commented):
Please don't change the module version manually. And now that we are here, let's reorganize code in a more appropriated way:

- The try should be inside `if report_type == "xlsx":`, not outside covering also the super call.
- The `reportname` variable initial assignment also inside the if.

Thanks.

pedrobaeza:
Please do all the changes in the same commit and rebase your branch to fix conflicts.

PR #757, opened 2023-04-20
[14.0][FIX] report_qweb_signer: remove `cryptography` from depends

luisg123v:
Such dependency is already included in Odoo requirements using a pinned version. Adding here could cause to upgrade the library to an incompatible version.

### DIFF
--- a/report_qweb_signer/__manifest__.py
+++ b/report_qweb_signer/__manifest__.py
@@ -14,7 +14,7 @@
     "installable": True,
     "depends": ["web_editor"],
     "external_dependencies": {
-        "python": ["endesive", "cryptography"],
+        "python": ["endesive"],
         "deb": ["default-jre-headless"],
     },
     "data": [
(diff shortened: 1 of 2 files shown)


luisg123v:
Hi @moylop260, @omar7r, @pedrobaeza,

Could you review, please?

CC @randall-vx @desdelinux

moylop260:
Could you share the URL#LINE of the requirement already defined in the commit message and description, please?

pedrobaeza:
I don't see such dependency in https://github.com/odoo/odoo/blob/14.0/requirements.txt

luisg123v:
@pedrobaeza,

You're right, it was pinned in 15.0 on https://github.com/odoo/odoo/pull/99829, , and previously implicitly installed by `pyopenssl` on https://github.com/odoo/odoo/commit/a4d513034ea8.

I'm therefore closing this and avoiding removing it from similar PRs, at least in v14.

Regards,

PR #761, opened 2023-05-03
[16.0][MIG] bi_view_editor

huguesdk:
## `BVEEditor` widget

the `BVEEditor` widget (in the “query builder” tab) has been converted to owl. this is quite a big change, almost a rewrite. it should work in exactly the same way as in 15.0 (except for one slight difference, as explained below, and little styling changes).

the only difference that i’m aware of is that if a model is expanded in the left column and the model filter is changed, in 15.0 all models are collapsed, while now the ones that are still visible stay expanded if they were. i don’t know if the previous behavior was desired or was due to a technical limitation, and i think the current way makes sense, so i left it like that. doing it with owl in the same way as in 15.0 would be possible but cumbersome, because parent components cannot access their children components in an easy way (at least i didn’t find how to do it except by making children register themselves with their parent using a callback).

i tried to keep the existing logic as much as possible to lower the amount of changes. for example, the `FieldList` component still manages the data of the field list, and has `.get()` and `.set()` methods called directly by the main widget, which in a reactive framework like owl is a little weird.

this is the first time i work with owl (learning it along the way), so there are probably things that are not done exactly like they should. for example, i didn’t succeed in using custom events to notify parent components of changes, so i used [callbacks](https://github.com/odoo/owl/blob/master/doc/reference/props.md#binding-function-props) instead. all advice and comments are welcome.

## known issues

in 16.0, translations are handled differently than before. the `ir.translation` table doesn’t exist anymore. because of this, i removed the translation smart button that was available on created bi views. the translations can be edited directly on the values themselves (name of the bi view and field names in the details tab).

additionally, there are 2 things that don’t work anymore like they do in 15.0 because of how translations are handled:

1.  translations of the name of a bi view and of the field descriptions do not work as expected: the translated strings are selected (by the user’s language) when the view is generated (and stored as their `en_US` value) instead of when it is displayed.
2.  ~translated fields are not supported in bi views.~

the 1st issue is due to how the models and fields are created: only one value is set as their name field, while internally the field is a json object. currently, the user’s language defines the value for the `en_US` language, regardless of which language it actually is.

~the 2nd issue is due to the fact that the code handles translated fields as regular fields while they need to be handled in a special way. here is how to reproduce the problem:~

1.  ~create a new bi view.~
2.  ~give it a name (or it won’t save, as it’s required).~
3.  ~add the “country” → “country code” and “country” → “country name” fields.~
4.  ~click on “generate bi view”.~
5.  ~click on “open bi view”.~
6.  ~problem: the “country name” column displays empty values.~
7.  ~click on the pivot button on the top right.~
8.  ~click on a “total” cell and select “country name”.~
9.  ~the values displayed are all “[object Object]”.~
10. ~click on any of those values and select “country code”.~
11. ~problem: the following error is raised:~
    ```
    psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type json
    LINE 4: ...       WHERE ("x_bve_test0"."x_bve_t1_name" = '{''en_US'...
                                                             ^
    DETAIL: Token "'" is invalid.
    CONTEXT: JSON data, line 1: {'...
    ```

any help to solve this is welcome.

## edit

translated fields are now correctly supported.


legalsylvain:
Hi @huguesdk. Do you know @pierrickbrun from akretion is working on it since some weeks ? 

https://github.com/OCA/reporting-engine/pull/755

huguesdk:
@legalsylvain ouch, this is unfortunate. i started working on this a while ago, and didn’t check since then. 🤦 @PierrickBrun, since you dived into the code on your side, what do you think of this version?

PierrickBrun:
I feel like this version is more finished than mine (ie. I did not look into translations at all atm, the drag/drop functionnality, the css on your side looks better)

about the owl migration of the widget, it looks like we took similar paths. I made a first attempt at using a store to put the fields but I did not like that the children components would directly make changes to it so I used callbacks, like you did

maybe we can use this one and take things that may be better in mine to complete it ?

I will squash my commits to make it more readable

huguesdk:
thanks for your reply, @PierrickBrun. indeed, this version is supposed to be finished (except for the limitations regarding translations), so i think like you that it makes sense to try to improve this one.

i have to work on other things at the moment, but any change suggestion is welcome.

PierrickBrun:
> thanks for your reply, @PierrickBrun. indeed, this version is supposed to be finished (except for the limitations regarding translations), so i think like you that it makes sense to try to improve this one.
> 
> i have to work on other things at the moment, but any change suggestion is welcome.

From what I've read, I'd reuse the class Fields and the fact that all the logic is on the BiViewEditor component (the subcomponents are "dumb" on my implementation). I find it better than putting this logic on the `FieldList`component, as you pointed out

This is the only thing I think is worth reusing but feel free to read my code and reuse what you like

huguesdk:
during the development, i was also tempted to move the logic back from `FieldList` to `BiViewEditor`. what decided me to not do it was:

1.  during [the migration to 11.0](https://github.com/OCA/reporting-engine/commit/7e398034ba6c835f443eaa4bf3ef699cded82535), the widget was refactored from one widget doing everything to separate widgets having their own responsibility, which i found more logical.
2.  to limit the amount of changes, to avoid breaking too much stuff (possibly introducing new bugs), and spending less time on it.

this said, in 15.0 the `FieldList` widget was storing and getting all its data into/from the dom, while here the dom is generated dynamically from the data.

so i think it makes sense indeed to move the data management back to the main component (preferably encapsulating it in another class, like you did), while leaving the `FieldList` component only to display and interact with the data, but not storing or modifying it.

PierrickBrun:
@huguesdk Can you reuse Odoo's dashboard menu like it is done in bi_sql_editor ? 

https://github.com/OCA/reporting-engine/commit/6be6f7d02f8612ec0d2cfa836eef987f63457281 for reference 

Also I noticed that the button "Create a Button" is now called "Date", I don't think it is defined in this module but there is something odd there

huguesdk:
@PierrickBrun i just moved the menus to spreadsheet dashboards, like `bi_sql_editor`. the only problem i see with this is that it adds `spreadsheet` as an unnecessary (and quite big) dependency. do you think this is ok?

i don’t see the button you’re mentioning. where is it?

(the failing tests are due to a problem in `bi_sql_editor` apparently, unrelated to this module.)

PierrickBrun:
> the only problem i see with this is that it adds `spreadsheet` as an unnecessary (and quite big) dependency. do you think this is ok?

You can use `spreadsheet_dashboard` as a dependency instead :smile:

huguesdk:
> You can use `spreadsheet_dashboard` as a dependency instead :smile:

that’s what i did, but that one depends on `spreadsheet`.

PierrickBrun:
> > You can use `spreadsheet_dashboard` as a dependency instead smile
> 
> that’s what i did, but that one depends on `spreadsheet`.

Oh I did not understand that, I don't understand the use for `spreadsheet_dashboard` then...

I think that we can keep it like this since spreadsheet is an odoo module

huguesdk:
@PierrickBrun i moved the fields’ data handling code from `FieldList` to `BiViewEditor`. i didn’t use a separate class like you did because it’s actually just an array with some methods, and the field array would be at `state.fields.fields` which seems a little strange. another way would be to inherit from `Array`, but somehow this also seems strange to me. what do you think?

bi_view_editor/static/src/components/bi_view_editor/field_list.esm.js
```diff
@@ -0,0 +1,131 @@
...
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+import {Component, onMounted, useRef, useState} from "@odoo/owl";
+
+class FieldListItem extends Component {
+    // This is only to simplify the template (using field instead of
+    // props.field).
+    get field() {
+        return this.props.field;
+    }
```
remytms: I think that this is not really useful. Having `props.field` in the template is common. This act as an exception for only some of the component.

bi_view_editor/static/src/components/bi_view_editor/field_list.esm.js
```diff
@@ -0,0 +1,131 @@
...
+    setDescription: Function,
+    readonly: Boolean,
+};
+
+class JoinListItem extends Component {
+    // This is only to simplify the template (using field instead of
+    // props.field).
+    get field() {
+        return this.props.field;
+    }
```
remytms: Idem.

remytms (approved):
LGTM. Some comment, but not blocking IMO.

mrm-david:
Just out of curiosity, what's holding this PR ?

huguesdk:
@mrm-david i’ve just made the slight change suggested by @remytms. i think it should be good now. another review could help, though.

yankinmax:
hello @huguesdk, could you pls forward-port this commit to 16.0. It adds a group to manage `bi_view_editor`:
- [ ] https://github.com/OCA/reporting-engine/pull/801

huguesdk:
@yankinmax sure, i’ll do it once it is accepted and merged in 15.

bi_view_editor/__manifest__.py
```diff
@@ -11,7 +11,9 @@
     "category": "Productivity",
     "version": "16.0.1.0.0",
     "development_status": "Beta",
-    "depends": ["web"],
+    "depends": [
+        "spreadsheet_dashboard",
```
ivantodorovich: ⚠️ do we really want to add the `spreadsheet` dependency just for this menu?

ivantodorovich (commented):
btw, could you squash or reword your commits with meaningful messages before merging?

bi_view_editor/__manifest__.py
huguesdk: i also think that this is a quite big and useless dependency. what do you suggest? put it in `base.menu_board_root` by default and add a glue module (auto-install) that depend on `spreadsheet_dashboard` and moves the menu entry there?

what do you think, @PierrickBrun?

huguesdk (commented):
@ivantodorovich should i just squash all commits after (and including) the migration commit to one?

bi_view_editor/__manifest__.py
ivantodorovich: That sounds good to me 👍🏻

ivantodorovich:
> should i just squash all commits after (and including) the migration commit to one?

Yes, you can do that. If all the changes are strictly due to the migration process, then it makes sense to have all of them in a single "migration" commit.

Though, if you have other improvements you can decide to put them in separate commits (optional, they can be on the migration commit, too). However if you do this, please write more descriptive commit messages and try to stick to [the guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message) for formatting it. Messages like "add comment" don't say much 😓

Camille0907 (approved):
I'm not totally at ease with owl, but LGTM from what I understand :)

huguesdk:
@ivantodorovich:

*   removed dependency on `spreadsheet_dashboard`.
*   added glue module to move menu items in case `spreadsheet_dashboard` is installed.
*   squashed all commits to one.

please note that in the odoo master branch, the legacy `base.menu_board_root` menu has been removed, so eventually, either the dependency on `spreadsheet_dashboard` will be necessary, or a simple custom dashboard menu will need to be defined in this module.

ivantodorovich (approved):
thanks!

HviorForgeFlow:
@huguesdk seems that pre-commit is failing https://github.com/OCA/reporting-engine/actions/runs/6781745670/job/18432619470#step:7:19

huguesdk:
@HviorForgeFlow i’ve just added the missing file. could you please retry?

edit: i’m surprised that pre-commit forces the `__init__.py` file to be there. the module is working well without it. maybe it was necessary in previous versions and not anymore?

HviorForgeFlow (approved):

PR #766, opened 2023-06-05
[16.0][FIX] Fix warning removing the useless param nodestroy in sql export wizard

florian-dacosta:


legalsylvain (approved):
Naive question : What is the objective of this old param ?

florian-dacosta:
> Naive question : What is the objective of this old param ?

I am not even sure...It was used in v7...

PR #697, opened 2022-12-13
[15.0][ADD] report_qweb_decimal_place

AungKoKoLin1997:
This module intends to provide the base function for currencies to adjust the number of decimal places for the unit price in QWeb reports.
@qrtl

PR #769, opened 2023-06-15
[15.0][ADD] report_qweb_decimal_place

AungKoKoLin1997:
This PR is recreated from https://github.com/OCA/reporting-engine/pull/697 that was closed as stale.
This module intends to provide the base function for currencies to adjust the number of decimal places for the unit price in QWeb reports.

@qrtl


yostashiro (approved):

kanda999 (approved):

rinaldifirdaus (approved):
Functional test: work as expected!

report_qweb_decimal_place/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,28 @@
...
+Background:
+~~~~~~~~~~~
+
+Odoo default reports display price unit with the decimal accuracy of product price configuration.
+However, globally applying the decimal accuracy setting is sometimes not appropriate under multi-currency settings
+where how unit prices should be presented differs depending on the currency.
+
+For example, unit prices in JPY usually do not have decimals (with some exceptions depending on the industry),
+while those in USD may require up to 2-4 decimals.  If we configure the decimal accuracy based on USD, the unit price
+presentation on PDF reports for JPY transactions may appear  a bit unconventional.
```
thomaspaulb, line 28: Hm, but if you look at for example the display of subtotal in Odoo 15's [sale report](https://github.com/OCA/OCB/blob/15.0/addons/sale/report/sale_report_templates.xml#L134), it uses `<span t-esc="current_subtotal" t-options='{"widget": "monetary", "display_currency": doc.pricelist_id.currency_id}'/>`.

Under the hood, the [monetary widget](https://github.com/OCA/OCB/blob/15.0/odoo/addons/base/models/ir_qweb_fields.py#L433) uses the currency to determine the amount of decimal places to display.

So, in the above example, why wouldn't you go for:

    <xpath expr="//td[@name='td_priceunit']/span" position="replace">
        <t t-esc="line.price_unit" t-options='{"widget": "monetary", "display_currency": doc.currency_id}' />
    </xpath>

AungKoKoLin1997: Thanks for the review. But the purpose of this module is not only for using same decimal place with the currency. Also for dynamic decimal place for price unit in pdf reports.

thomaspaulb (approved):

PR #770, opened 2023-06-15
[FIX] unlink bi_sql_view without cron

PierrickBrun:
@legalsylvain trivial fix PR

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -263,7 +263,8 @@ class BiSQLView(models.Model):
                     "If you want to delete them, first set them to draft."
                 )
             )
-        self.cron_id.unlink()
+        if self.cron_id:
+            self.cron_id.unlink()
         return super().unlink()
 
     def copy(self, default=None):


bi_sql_editor/models/bi_sql_view.py
```diff
@@ -263,7 +263,8 @@ def unlink(self):
                     "If you want to delete them, first set them to draft."
                 )
             )
-        self.cron_id.unlink()
+        if self.cron_id:
+            self.cron_id.unlink()
```
legalsylvain, line 267: Maybe replace two lines by

self.mapped("cron_id").unlink()

Otherwise, LGTM

PierrickBrun: I've just tested, it does not work either.

I think the real problem is that unlink() crashes if called from an empty recordset

PierrickBrun: probably only on ir_cron, that said: https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/ir_cron.py#L401

PierrickBrun: I've created this PR: https://github.com/odoo/odoo/pull/125232

legalsylvain: ah yes. makes sense.

leemannd: Hello, open question, why not setting the field as `ondelete=cascade` ?

PR #771, opened 2023-06-19
[16][FIX] sql_export_mail : sql users need read right on cron to access the query

florian-dacosta:
@legalsylvain One more patch needed for sql_export_mail in v16.
The users have to be able to read the cron model to be able to access a sql report, since it maybe linked to a cron to be sent by email automatically.

### DIFF
--- a/sql_export_mail/__manifest__.py
+++ b/sql_export_mail/__manifest__.py
@@ -13,6 +13,7 @@
     ],
     "license": "AGPL-3",
     "data": [
+        "security/ir.model.access.csv",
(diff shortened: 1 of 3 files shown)


legalsylvain:
Hi @florian-dacosta. I don't think giving access to all user to the read of cron is a good idea. (can be a security issue + it's not the design of odoo for the time being).
proposing alternative fix here : https://github.com/OCA/reporting-engine/pull/772

florian-dacosta:
Ok, I have no strong opinion about the way to solve this issue, I'll close this one

PR #772, opened 2023-06-20
[FIX] sql_export_mail : do not raise an error if non admin user want to access to sql export

legalsylvain:
Alternative implementation of https://github.com/OCA/reporting-engine/pull/771

### DIFF
--- a/sql_export_mail/views/sql_export_view.xml
+++ b/sql_export_mail/views/sql_export_view.xml
@@ -18,6 +18,7 @@
                 <field
                     name="mail_condition"
                     attrs="{'invisible': [('cron_ids', '=', [(6, False, [])])]}"
+                    groups="base.group_system"
                 />
             </field>
             <page name="page_sql" position="after">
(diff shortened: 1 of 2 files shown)


sql_export_mail/views/sql_export_view.xml
```diff
@@ -36,6 +37,7 @@
                             nolabel="1"
                             colspan="2"
                             domain="[('model_id', '=', 'sql.export')]"
+                            groups="base.group_system"
```
florian-dacosta: I would put the cron on the group level (or both). Else, the user  without group will just see the group name with nothing else and it is kind of strange.
Apart from this, (and pre-commit small diff)  LGTM, thanks

legalsylvain:
Thanks ! 
done.

florian-dacosta (approved):

bguillot (approved):

bguillot:
Thanks !

PR #773, opened 2023-06-26
[FIX] report_csv: Assign 'reportname' variable before accessing it

rousseldenis:
@AnizR @adrienpeiffer

### DIFF
--- a/report_csv/controllers/main.py
+++ b/report_csv/controllers/main.py
@@ -54,6 +54,7 @@ class ReportController(report.ReportController):
     def report_download(self, data, context=None):
         requestcontent = json.loads(data)
         url, report_type = requestcontent[0], requestcontent[1]
+        reportname = ""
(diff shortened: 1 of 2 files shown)


rousseldenis:
@adrienpeiffer In addition of this fix, shouldn't an error re-raise werkzeug InternalServerError instead ?

Like this : https://github.com/odoo/odoo/blob/16.0/addons/stock/controllers/main.py#L38

ioans73 (approved):

AnizR (approved):
Great!

adrienpeiffer:
@rousseldenis Totally agree with you !

rousseldenis:
@adrienpeiffer @AnizR @ioans73 In addition, I've added the InternalServerError return and a test for controller to increase coverage

AnizR (approved):

PR #776, opened 2023-07-10
[13.0][IMP]report_xlsx: posibility to use boilerplate template on xlsx report

GuillemCForgeFlow:
These changes add the possibility to define a boilerplate template from which to start working from. From there, you can add the needed extra information for the inherited report.

cc @ForgeFlow


GuillemCForgeFlow:
@adrienpeiffer @sebalix @sbidoul @etobella 
Perhaps I could get a review here? 😃

JordiMForgeFlow (approved):
Code review + functionally tested 👍🏼

HviorForgeFlow (approved):
Code review :+1:

report_xlsx/report/report_abstract_xlsx.py
```diff
@@ -5,7 +5,14 @@
 import re
 from io import BytesIO
 
-from odoo import models
+import openpyxl
```
etobella, line 8: you should add a try except just to not break old installations

pedrobaeza (changes requested):
This change is not valid for a stable version. It's changing the dependent library. It should be added on an extra module, not here.

GuillemCForgeFlow:
> This change is not valid for a stable version. It's changing the dependent library. It should be added on an extra module, not here.

Ok, then I will add the feature in a new module.

sbidoul:
> This change is not valid for a stable version. It's changing the dependent library. It should be added on an extra module, not here.

I don't think we have such policy in OCA. With a major version bump we can change dependencies.

pedrobaeza:
No, Stéphane, if this is merged, it's breaking existing installations, so that's not acceptable. One thing is to change datamodel providing migration scripts, and another to totally break the module to existing users.

sbidoul:
Then we disagree on this. Such dependency refactorings did happen in the past and should continue be allowed IMO.

pedrobaeza:
This module is the base for a lot of other modules (account_financial_report, account_asset_management, etc), and one thing is to add an extra feature that doesn't work if you don't have the dependency, and another to break the existing behavior on a version that lasts 4 years. I think there are degrees of acceptance, and this one is not acceptable.

sbidoul:
Yes, I agree we must exercise common sense. I was talking about general policy, not specifics of this PR.

pedrobaeza:
I haven't talked in any moment about a general OCA policy. I started saying "This change"...

GuillemCForgeFlow:
closing this as the changes are not relevant anymore after the new module adding the functionality got created/merged: https://github.com/OCA/reporting-engine/pull/798

PR #778, opened 2023-07-19
[15.0][FIX] base_comment_template: search when multiple models in a template

AaronHForgeFlow:
I have comment templates that are shared between several models:

image

I think the module allows to do so in the specifications and the _search_model_ids method needs to change it will output an error

@ForgeFlow

### DIFF
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -120,6 +120,8 @@ class BaseCommentTemplate(models.Model):
     def _search_model_ids(self, operator, value):
         # We cannot use model_ids.model in search() method to avoid recursion.
         allowed_items = (
-            self.sudo().search([]).filtered(lambda x: x.model_ids.model == value)
+            self.sudo()
+            .search([])
+            .filtered(lambda x: value in x.model_ids.mapped("model"))
         )
         return [("id", "in", allowed_items.ids)]
--- a/base_comment_template/models/comment_template.py
+++ b/base_comment_template/models/comment_template.py
@@ -24,7 +24,7 @@ class CommentTemplate(models.AbstractModel):
         compute_sudo=True,
         comodel_name="base.comment.template",
         string="Comment Template",
-        domain=lambda self: [("model_ids", "=", self._name)],
+        domain=lambda self: [("model_ids", "in", self._name)],
         store=True,
         readonly=False,
     )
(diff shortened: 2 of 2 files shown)


JordiMForgeFlow (approved):
Makes sense 👍🏼

base_comment_template/models/base_comment_template.py
```diff
@@ -120,6 +120,8 @@ def name_get(self):
     def _search_model_ids(self, operator, value):
         # We cannot use model_ids.model in search() method to avoid recursion.
         allowed_items = (
-            self.sudo().search([]).filtered(lambda x: x.model_ids.model == value)
+            self.sudo()
+            .search([])
+            .filtered(lambda x: value in x.model_ids.mapped("model"))
```
pedrobaeza, line 125: @victoralmau this was prepared for models split by commas?

@AaronHForgeFlow in any case, you should split the values by comma, and do exact match. If not, the model `purchase.order` will match with for example `purchase.order.line`

AaronHForgeFlow: model_ids is computed correctly, it has the 3 ir.model records. But then I get singleton error without this change.

AaronHForgeFlow: The mapped is a list, The filtered is checking the string with the model is in the list. The complete string must match an element in the list, otherwise it will be filtered out, I think:

```
>>> list = ['sale.order', 'purchase.order']
>>> 'purchase.order' in list
True
>>> 'purchase.order.line' in list
False
>>> list = ['sale.order.line', 'purchase.order.line']
>>> 'purchase.order' in list
False
```

victoralmau: Yes, the original idea is to be able to define several models separated by comma (`model_ids` will be self-defined).

Although the proposal is correct, I think this https://github.com/OCA/reporting-engine/blob/15.0/base_comment_template/models/comment_template.py#L35 would have to be adapted as well.

AaronHForgeFlow: Thank you @victoralmau, you are right. I adapted that part as well and it seems to be working when putting several models in comment templates and trying to get them in records for one of those models. They don't show if the record is for a model that is not in the template.

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to v16 if applicable

PR #779, opened 2023-07-24
[16.0][MIG] report_qweb_encrypt: Migration to 16.0

payen000:


payen000:
Hello @luisg123v, could you review this, please?

report_qweb_encrypt/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "depends": [
+        "web",
+    ],
+    "data": [
+        "views/ir_actions_report.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "report_qweb_encrypt/static/src/js/report/action_manager_report.esm.js",
+            "report_qweb_encrypt/static/src/js/report/encrypt_dialog.xml",
```
luisg123v: You're placing a XML file under the js directory.

I think what you should do instead is removing the js directory and place both files inside report, because, starting from OWL structure, files are not placed according to their languages but according to their purpose.

report_qweb_encrypt/i18n/es.po
```diff
@@ -0,0 +1,81 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* report_qweb_encrypt
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
+"Report-Msgid-Bugs-To: \n"
+"Last-Translator: Automatically generated\n"
```
luisg123v: Don't touch translation files.

You may regenerate pot file if you would like to. Both pot and po will be regenerated by the bot anyway.

report_qweb_encrypt/models/ir_actions_report.py
```diff
@@ -0,0 +1,78 @@
...
+                encrypt_password=encrypt_password
+            )
+            password = report._get_pdf_password(res_ids[:1])
+            document = self._encrypt_pdf(document, password)
+        return document, ttype
+
+    def _get_pdf_password(self, res_ids):
+        encrypt_password = False
+        if self.encrypt == "manual":
+            encrypt_password = self._context.get("encrypt_password", False)
```
luisg123v, line 49: You're removing comments above this line

report_qweb_encrypt/static/src/js/report/action_manager_report.esm.js
```diff
@@ -0,0 +1,79 @@
...
+import {download} from "@web/core/network/download";
+import {registry} from "@web/core/registry";
+const {Component} = owl;
+
+async function download_function(action, options, env) {
+    const type = action.report_type === "qweb-pdf" ? "pdf" : action.report_type;
+    let url = `/report/${type}/${action.report_name}`;
+    const actionContext = action.context || {};
+    const userContext = Object.assign(env.services.user.context, {
+        encrypt_password: action.context.encrypt_password,
```
luisg123v: did you mean encrypted?

luisg123v (changes requested):
There is only one commit, you're not bringing commit history.

Please follow [these instructions](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#technical-method-to-migrate-a-module-from-150-to-160-branch) to bring commit history.

This is important, because otherwise:
- It's hard for reviewer to review the actuall migration changes
- Authorship for all existing changes is lost

report_qweb_encrypt/__manifest__.py
payen000: I think this makes sense and is consistent with other modules' structures, I've corrected this, thanks!

report_qweb_encrypt/i18n/es.po
payen000: Fixed, thanks!

report_qweb_encrypt/models/ir_actions_report.py
payen000: I originally deleted them because I believed they were no longer relevant, but they still are, so I've added the back, thanks

report_qweb_encrypt/static/src/js/report/action_manager_report.esm.js
payen000: No, `encrypt_password` is the context key that points to the password that is fetched in the backend to encrypt the pdf - here:

https://github.com/OCA/reporting-engine/blob/e9536926133fc3f0a79bcdd226b40cdd56059675/report_qweb_encrypt/models/ir_actions_report.py#L49

so the name was intentional.

payen000:
@luisg123v I've updated this PR, could you review again, please?

report_qweb_encrypt/models/ir_actions_report.py
```diff
@@ -0,0 +1,81 @@
...
+            Model = self.env[self_sudo.model]
+            record_ids = Model.browse(res_ids)
+            try:
+                encrypt_password = safe_eval(
+                    self.encrypt_password, {"object": record_ids}
+                )
+            except Exception as e:
+                raise ValidationError(
+                    _("Python code used for encryption password is invalid.\n%s")
+                    % self.encrypt_password
```
luisg123v: Previous version was the correct one, you should use the lazy syntax:

`_("...%s..", param)`

Instead of:

`_("...%s..") % param`

luisg123v (changes requested):
I noticed you brought commit history, you just missed one commit: 556e60b4c3d2faa5096067082e4d28fc7db56981

requirements.txt
```diff
@@ -1,4 +1,5 @@
 # generated from manifests external_dependencies
 lxml
+PyPDF2
```
luisg123v: Please remove this dependency from manifest, this is already considered by Odoo's requirements [here](https://github.com/odoo/odoo/blob/253479633b32383e141fe2ca2ca7395f22651485/requirements.txt#L42).

Please take into account this requirements file is autogenerated from module's external dependencies. So, once you remove it from there, it should also be removed from here.

luisg123v (changes requested):

requirements.txt
payen000: Done :)

report_qweb_encrypt/models/ir_actions_report.py
payen000: Done :)

payen000:
@luisg123v I've added the missing commit and applied the rest of your suggestions, please let me know if there's anything else missing, thanks!

report_qweb_encrypt/models/ir_actions_report.py
luisg123v: Done but in the incorrect commit, it's placed in the pre-commit commit

requirements.txt
luisg123v: Instead of adding it into the pre-commit commit to then removing it into the migration commit, just don't include it.

report_qweb_encrypt/models/ir_actions_report.py
payen000: I figured this was something "pre-commit-ish", so I placed it there, but I've moved it to the other commit, thanks!

requirements.txt
payen000: That makes sense, I've updated the code, thanks!

luisg123v (approved):
LGTM :+1: 


@moylop260 could you review and merge, please?

moylop260 (approved):
👍

PR #783, opened 2023-08-16
[16.0][MIG] report_context: Migration to 16.0

AlexPForgeFlow:
Migration to 16.0

[ForgeFlow](https://github.com/ForgeFlow)


mariadforgeflow (approved):
LGTM!

gpothier:
Hi, I tried the original v14 version of this module in a v16 install and it did not work out of the box, I had to copy the `route` params in the controller from the base module for it to work:
```python
    @route([
        '/report/<converter>/<reportname>',
        '/report/<converter>/<reportname>/<docids>',
    ], type='http', auth='user', website=True)
    def report_routes(self, reportname, docids=None, converter=None, **data):
```

(the v14 version just has `route()`  with no params)

I'm not sure if I'm looking at this MR right, because it appears as if all the files were new, so I don't see a diff wrt v14, but I see that the `route` params aren't there, so it is not clear to me why this would work. Maybe there were recent changes in v16 that prevent the empty `route()`  to work?

AlexPForgeFlow:
> ```python
> @route([
>         '/report/<converter>/<reportname>',
>         '/report/<converter>/<reportname>/<docids>',
>     ],
> ```

Hi @gpothier , I think they are not necessary as this method is inherited from the import "from odoo.addons.web.controllers import report" and the parent method already define the route. So in this module is only changing the body of the method.

gurneyalex (approved):

PR #787, opened 2023-08-25
[IMP] Avoiding ERROR in log with tests of report_qweb_pdf_watermark

BT-rmartin:
When having 0 pages it will raise and error
https://github.com/BT-rmartin/reporting-engine/blob/16.0-1/report_qweb_pdf_watermark/models/report.py#L56
We are in this way adding a context manager to test that at least one error message is logged on the logger with at least the given level.


BT-rmartin:
@bosd @bvanjeelharia @hbrunn @robin5210 Could you please review and approve this tiny change?

BT-anieto (approved):
LGTM

bosd (changes requested):
Please fix pre-commit

BT-rmartin:
@bosd Fixed

bosd (approved):

PR #788, opened 2023-08-29
[16.0][IMP] base_comment_template: allow to select rendering engine

alan196:
I made this PR based on comment https://github.com/OCA/sale-reporting/pull/227#issuecomment-1696503543

I have added a new field `engine` to allow to set the engine that will be used to render the template.

- I have set the default engine to `inline_template` in order to not break current installations.
- I updated the method `render_comment ` to work as it was designed and if the engine is nos passed to the method it will take the one from the template in order to not break current installations.
- I have added the field `engine` in the view only shown with developer mode.

### DIFF
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -72,6 +72,16 @@ class BaseCommentTemplate(models.Model):
     sequence = fields.Integer(
         required=True, default=10, help="The smaller number = The higher priority"
     )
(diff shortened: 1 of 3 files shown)


alan196:
@pedrobaeza could you help me to review this PR 🙏

pedrobaeza (approved):
Thanks

EduardoJarsa (approved):
functional review check, its works well  👍🏼

PR #795, opened 2023-10-10
Update fr.po

yassine-rhimi:


pedrobaeza:
Please use Weblate for translations: https://odoo-community.org/page/translate

PR #798, opened 2023-10-13
[13.0][ADD]report_xlsx_boilerplate

GuillemCForgeFlow:
These changes add the possibility to define a boilerplate template from which to start working from. From there, you can add the needed extra information for the inherited report.

related with https://github.com/OCA/reporting-engine/pull/776 --> Creating new module based on feedback

cc @ForgeFlow


GuillemCForgeFlow:
@pedrobaeza @etobella @JordiMForgeFlow @HviorForgeFlow any kind review here? 😊 

Comes from the feedback on https://github.com/OCA/reporting-engine/pull/776

JordiMForgeFlow (approved):
Code review LGTM 👍🏼

HviorForgeFlow (approved):
Code review :+1: 
Seems promising

report_xlsx/__manifest__.py
```diff
@@ -10,6 +10,7 @@
     "development_status": "Production/Stable",
     "license": "AGPL-3",
     "depends": ["base", "web"],
+    "external_dependencies": {"python": ["xlsxwriter"]},
```
pedrobaeza: Why changing this?

report_xlsx/report/report_abstract_xlsx.py
```diff
@@ -116,3 +116,15 @@ def get_workbook_options(self):
 
     def generate_xlsx_report(self, workbook, data, objs):
         raise NotImplementedError()
+
+    @api.model
+    def _get_new_workbook(self, file_data):
```
pedrobaeza, line 121: Better to add these hooks in a separate PR. Anyway, not sure this is the best option, as you are assuming the rest of the manipulation is handling the proper `workbook` object, which may not be. Why not directly hook `generate_xlsx_report` and if you have a boilerplate, process it with the other library?

report_xlsx/__manifest__.py
GuillemCForgeFlow: I could see that `report_xlsx` was using the `xlsxwriter` library already (https://github.com/OCA/reporting-engine/blob/13.0/report_xlsx/report/report_abstract_xlsx.py#L13), but the external dependency was not explicitly in the manifest file. I thought that this should be added so that the library is also added to the `requirements.txt` file. Perhaps this is not needed?

pedrobaeza: It's already in Odoo requirements, so that's why is not needed.

report_xlsx/report/report_abstract_xlsx.py
GuillemCForgeFlow: I see your point, the thing is that I cannot hook the `generate_xlsx_report` method as this is simply overwritten on the modules that are creating the new xlsx report.

Perhaps what I can do is trigger an error if the boilerplate is not correctly converted, instead of simply returning an empty workbook here: https://github.com/OCA/reporting-engine/pull/798/files#diff-0dc6d617191c2199fc13e1c7670f57d71f99f354c11d49cceb2868625dc4d668R60

Perhaps I misunderstood something?

report_xlsx/__manifest__.py
GuillemCForgeFlow: Oh, I see. I will remove it then. Thank you 😊

report_xlsx/report/report_abstract_xlsx.py
pedrobaeza: OK, got it.

pedrobaeza (approved):
/ocabot merge nobump

PR #801, opened 2023-10-19
[15.0][IMP] `bi_view_editor`: Restrict access to non BI View Editor Manager users

yankinmax:


legalsylvain (changes requested):
too much files changed. Could you fix it ? 

thanks !

yankinmax:
@legalsylvain It's because I regenerated translations. The group name was added to the translation files. I can remove them.

yankinmax:
@legalsylvain sorry, I forgot translations will be regenerated automatically, too much translations generated on projects recently :smile:

yankinmax:
@legalsylvain is this ok now?

legalsylvain:
I don't use personaly this module. I prefer to have a point of view of the initial developers.

bi_view_editor/security/ir.model.access.csv
```diff
@@ -1,3 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_bve_view_everyone,bve.view,bi_view_editor.model_bve_view,,1,1,1,1
-access_bve_view_line,access_bve_view_line,model_bve_view_line,,1,1,1,1
+access_bve_view_everyone_user,bve.view.user,bi_view_editor.model_bve_view,,0,0,0,0
```
ivantodorovich: Are you sure this is ok, from a funcitonal pov? I'm not very familiar with this module, but like this the "user" has no permissions, not even read permissions (`0,0,0,0`). Will the user be able to read the generated reports?

From a technical pov, an `ir.access.model` record with 0 permissions is the same as no record, so assuming the functional behaviour is wanted, I'd just remove these lines

yankinmax: yep of course, my bad. Thanks @ivantodorovich, I would go with usual user and manager rights. So, `base.group_user` can read, `group_bve_manager` can do CRUD

ivantodorovich (approved):

vvrossem (approved):
LGTM

yankinmax:
@astirpe @monen17 can you pls take a look?

bi_view_editor/security/res_groups.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
```
monen17:
```suggestion
<odoo noupdate="1">
```
I think groups usually are created as not-updateable, otherwise if the user edits the group their modifications are overwritten by the module

monen17 (changes requested):
Thanks for the PR!
Right now all users can edit the views so, in order to avoid unexpected behaviors when the module is updated, I think a migration is needed to assign the new group to everyone; what do you think?

legalsylvain:
>  Right now all users can edit the views so, in order to avoid unexpected behaviors when the module is updated, I think a migration is needed to assign the new group to everyone; what do you think?

Not sure. It will generate errors, if base_user_role (other OCA module) is installed.
If you fix a security issue, it's "normal" that the accesses change. don't you think ?

monen17:
> >  Right now all users can edit the views so, in order to avoid unexpected behaviors when the module is updated, I think a migration is needed to assign the new group to everyone; what do you think?
> 
> Not sure. It will generate errors, if base_user_role (other OCA module) is installed.
> If you fix a security issue, it's "normal" that the accesses change. don't you think ? 
> 

Is this fixing a security issue?
Whoever has this module installed right now is expecting that everyone can work on the views.
It would be at least unexpected that only admin and Odoobot can access the views after an update.
I still think that current behaviour should be kept for whoever has the module already installed.

New installs should be aware of the module's behaviour, and it should be mentioned in the README; otherwise a normal user installing the module wouldn't be able to access the menu mentioned in the **Usage** section.

yankinmax:
@monen17 I think it's normal when some functionality of module changes/new feature appears.
When someone is updating the module of course he can expect that something can change.
New fields added for example.
It's the way OCA modules evolve.
Assigning this group to everyone is not what we need.
What I propose:
1. We can restrict access not to menu, but to `bve.view` records CRUD operations. So, everyone will still have access to the menu, but only allowed Users to create custom reports.
2. I agree with you we should update README.
3. I'll update PR, add `noupdate`.

leemannd (approved):
Seems legit. Thank you

Camille0907 (approved):
LGTM :)

yankinmax:
@monen17 is https://github.com/OCA/reporting-engine/pull/801/commits/4df35fa0d6108dc952eabd3bcaebb857deabc35c commit ok for you?

bi_view_editor/README.rst
monen17, line 1: **README** file is overwritten by files in the **readme** folder, so those files should be edited instead (there is also a warning about it at the top of this file).
You can have a look at https://github.com/OCA/maintainer-tools#readme-generator for more details

bi_view_editor/README.rst
```diff
@@ -76,6 +76,8 @@ Usage
 To graphically design your analysis data-set:
 
 - From the Dashboards menu, select "Custom BI Views"
+- BI Views creation is restricted to admin users and members of "BI View Editor Manager" group.
```
monen17:
```suggestion
- BI Views creation is restricted to members of "BI View Editor Manager" group.
```

bi_view_editor/README.rst
```diff
@@ -88,6 +90,7 @@ To access the created BI View with a dedicated menu:
 
 - If module Dashboard (board) is installed, the standard "Add to My Dashboard"
   functionality would be available
+- "Create a menu" is restricted to admin users and members of "BI View Editor Manager" group.
```
monen17:
```suggestion
- "Create a menu" is restricted to members of "BI View Editor Manager" group.
```

monen17 (changes requested):
Thanks for taking care of https://github.com/OCA/reporting-engine/pull/801#pullrequestreview-1695840126, everyone seeing the menu is a good compromise between not seeing anything and everyone being a view manager.

Here you have just a couple of notes, the only blocking comment is the fact that changes to the README are overwritten.

Also please fix the commits:

- Fix up the **fixup!** commit.
  You probably did it to let me see the clean changes (thanks!) but the commit history is part of what has to be reviewed so I can't really approve a PR having a **fixup!** commit.
  For the future, in my opinion you should always organize the commits as you want them to be merged.
- The commit message is too long, see https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message
  I suggest **[IMP] bi_view_editor: Restrict access to BI View Editor Managers**
  Then you can write any explanation in the 2nd and following lines fo the message.

yankinmax:
@monen17 thanks for the review, your suggestions are applied and PR is updated

yankinmax:
@monen17 @legalsylvain can we finish this PR review and merge?

yankinmax:
@HviorForgeFlow can you pls also pay attention on this PR? It's updated with @monen17 suggestions as maintainer and it will also be forwarded to v16

HviorForgeFlow (approved):
Code review :+1:

HviorForgeFlow:
@yankinmax ready to be forward ported

PR #820, opened 2023-11-28
[16][FIX] sql_export : remove unused action parameter

florian-dacosta:
Same as here : https://github.com/OCA/reporting-engine/pull/766

But it has been introduced again by mistake by https://github.com/OCA/reporting-engine/pull/718, which was merged after the PR 766...


legalsylvain (approved):
/ocabot merge patch

PR #823, opened 2023-12-05
[16.0][IMP] bi_sql_editor

legalsylvain:
* [IMP] Make installation idem potens
* [REF] simplify tests
* [FIX] allow to pass default values in copy() function
* [FIX] do not allow to create model (and sql views) if related model are not set on many2one fields. It prevents the error 'AttributeError: '_unknown' object has no attribute 'id''
* [FIX] Use correct key word 'optional'. (and not 'option') in tree view
* [IMP]  allow to reset to the previous state, and not only into 'draft' state
* [REF]  simplification : Remove conditional display related to field_description, as this field is required and so, allways defined
* [REM] option_context_field field, as it is now possible to implement that feature with optional='hide' on the field list view
* [IMP] Allow to change fields settings if state is 'Model Valid'. Avoid to reset to draft and to have to recreate view that can take a while, if the view is materialized
* [IMP] Improve code coverage, adding various settings on field items

### DIFF
--- a/bi_sql_editor/demo/bi_sql_view_demo.xml
+++ b/bi_sql_editor/demo/bi_sql_view_demo.xml
@@ -20,6 +20,7 @@ ORDER BY unexisting_field
     <record id="partner_sql_view" model="bi.sql.view">
         <field name="name">Partners View</field>
         <field name="technical_name">partners_view</field>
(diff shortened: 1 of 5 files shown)


AMEA-LYON (approved):
Fonctional testing : OK

PR #825, opened 2023-12-12
16.0][FIX] report_label fix a security issue

nayatec:
Odoo V16 does not allow 'regular' users (ie users with no Admin/Access Right group) to acces to ir.model at all and ir.ui.view either (only for group Admin/Settings for the latter), therefore i added some sudoes to allow users with no such right to be able to actually print labels.
Did not find a more elegant solution.


legalsylvain (approved):

ivantodorovich (approved):
Makes sense, thanks for fixing it!

PR #826, opened 2023-12-13
[14.0][ADD] report_qweb_table_pagebreak

petrus-v:
Control Page Breaks in `<table>`s within QWeb Reports

used by:
* https://github.com/OCA/account-invoice-reporting/pull/297
* https://github.com/OCA/sale-reporting/pull/243


report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+    @api.model
+    def _split_tables(self, html_content):
+        html_document = html.document_fromstring(html_content.decode("utf-8"))
+
+        for fragment in html_document:
+            table = None
+            for row in fragment.iterfind(".//tr[@data-page-break]"):
+                # while generating pdf this method is called twice so we prevent
+                # table to be split twice by removing data-page-break once
+                # table has been spited
```
SirAionTech:
```suggestion
                # table has been splitted
```

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+    def set(self, key, value, default=None):
+        if value is None:
+            value = default
+        if value is None:
+            self._styles.pop(key, None)
+        else:
+            self._styles[key] = value
+
+
+class IrQWeb(models.AbstractModel, QWeb):
```
SirAionTech: Is `QWeb` needed too?

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+                # this make sure we avoid break page inside the latest table parts
+                # it's not a matter if this table is split in the next loop
+                self._set_page_break_style(table, False)
+
+        return html.tostring(
+            html_document,
+        )
+
+    @api.model
+    def _set_page_break_style(self, table, force_break):
```
SirAionTech:
```suggestion
    def _set_page_break_style(self, table, force_break=False):
```
and avoid `False` parameter above.
In my opinion the developer should not be required to set parameters that _force_ a behavior: the normal usage of a method should require as few parameters as possible.

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+            for row in fragment.iterfind(".//tr[@data-page-break]"):
+                # while generating pdf this method is called twice so we prevent
+                # table to be split twice by removing data-page-break once
+                # table has been spited
+                pos = row.attrib.pop("data-page-break")
+                if pos not in ("before", "after", "avoid"):
+                    continue
+                table = next(row.iterancestors("table"))
+                tbody = next(row.iterancestors("tbody"))
+                newtable = html.Element("table", attrib=dict(table.attrib))
```
SirAionTech, line 71: Adding a new `table`/`tbody` might lead to unexpected behavior when printing the PDF if the `table`/`tbody` tag has some specific style from the CSS; maybe this is the only way, but this might be worth considering.

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+
+class IrQWeb(models.AbstractModel, QWeb):
+    _inherit = "ir.qweb"
+    _description = "Qweb"
+
+    @api.model
+    def _render(self, id_or_xml_id, values=None, **options):
+        result = super()._render(id_or_xml_id, values=values, **options)
+
+        if b"data-page-break=" not in result:
```
SirAionTech: `data-*` attributes have a [particular meaning in HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) and I always thought this was because they described the _data_ represented by the node; maybe something like `qweb-page-break` or `oca-page-break` might be more appropriate and avoid clashing with other custom parameters, what do you think?

SirAionTech:
```suggestion
        if b"data-page-break-inside=" not in result:
```
might be more readable: `data-page-break-inside="avoid"` vs `data-page-break="always"`.
This would also be more similar to the attribute `page-break-inside` used under the hood.
What do you think?

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+                    newtable.append(new_thead)
+                newtable.append(newtbody)
+                for sibling in row.getparent().iterchildren("tr"):
+                    if sibling is row:
+                        if pos == "after":
+                            newtbody.append(sibling)
+                        break
+                    newtbody.append(sibling)
+                table.addprevious(newtable)
+                self._set_page_break_style(newtable, pos != "avoid")
```
SirAionTech, line 85: If I set `data-page-break="before"`, then inside this method a `page-break-after="always"` is injected: I am a little confused, shouldn't it inject a `page-break-before="always"`?

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+                if pos not in ("before", "after", "avoid"):
+                    continue
+                table = next(row.iterancestors("table"))
+                tbody = next(row.iterancestors("tbody"))
+                newtable = html.Element("table", attrib=dict(table.attrib))
+                newtbody = html.Element("tbody", attrib=dict(tbody.attrib))
+                thead = table.find("thead")
+                if thead:
+                    new_thead = html.fromstring(html.tostring(thead))
+                    newtable.append(new_thead)
```
SirAionTech, line 76: I haven't tried this but what if the table fits in one page? Is this repeating the `thead` for each row having `data-page-break`?

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,104 @@
...
+            value = default
+        if value is None:
+            self._styles.pop(key, None)
+        else:
+            self._styles[key] = value
+
+
+class IrQWeb(models.AbstractModel, QWeb):
+    _inherit = "ir.qweb"
+    _description = "Qweb"
```
SirAionTech:
```suggestion
```
Why overriding the `_description` of the base model with the same value?

SirAionTech (changes requested):
Thanks for the PR, this is some awesome and inspiring work!
As far as I know the _page-break-in-PDF_ is a known issue (due to **wkhtmltopdf**) in Odoo from ages and this might help mitigating that.
Below some concerns, let me know what you think

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: Good catch, I don't think so !

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: :pray: too large copy/past

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: why not ! let's have a try using `qweb-page-break`

petrus-v: I prefer to use only one attribute such as `qweb-page-break` or `qweb-table-page-break` because:

* it keeps things different from css attributes (code is using `page-break-inside` and `page-break-after`) and avoid to mix-up both
* I feel easier to use only one attribute and manage behavior according the value 
* In qweb template that would manage the value I suppose code will be also easier to manage

I'll add a comment that any other value (thant `avoid`, `before` or `after`) are ignored

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: `style` attribute and all other  `tbody`/`table` attributes are copied on new element while creating the element with the  `attrib=dict(tbody.attrib)`. There is a unit test on `table` element, I'll add it on tbody as well

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: Yes, for any row that define a valid value for `data-page-break` this will split table repeating the header on each tables. When using `avoid` you can get two tables on the same page, column width can be different in such case. 

ps: you can test it on runboat with this PR https://github.com/OCA/sale-reporting/pull/243 which split table before each sections !

report_qweb_table_pagebreak/models/ir_qweb.py
SirAionTech: Sorry I was not clear enough, I was talking about CSS rules like
```
table {
  ...
}
```
that are applied to any `table`; on a second thought that also applies to all the attributes of the `table` node.

What I meant is: the user has added some style to the main table of the report, maybe it is not expected to see that same style applied to another table _inside_ the list of rows?
Maybe the new table should be as invisible as possible to mimic the visual appearance of normal rows?
I wouldn't know how to do that but I was just saying that it might be something to consider.

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: no because newtable is add before the existing tab, each row are previously moved to this new table, before/after is used to know if the current row should be part of the new table or keep it on the existing table

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: probably the name of the parameter is wrongly name I'll rename to `add_break_page`, thanks

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: yeah hard to find the best solution, I'll add a comment to warn developer that there are some magic that generate multiple tables instead of one.

And just realized that this code is also called while templates are shared between PDF report and portail views (likes for customer invoices)

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: fixed by 9f1724a7

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: done by 348eacfe

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: changed by 48aa0d8d

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: test added by e63bb184
some warning in documentaiton added by d1b9d84c

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: I did the change but I've the feeling it's also a documented HTML way to add custom attribute, I'm wondering if it gives valid html... but as the code remove the attribute in final html it doesn't matter that much !

done by 7b494564

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: done within 348eacfe

report_qweb_table_pagebreak/models/ir_qweb.py
SirAionTech: Thanks for the explanation, now it is more clear how it is working.

report_qweb_table_pagebreak/models/ir_qweb.py
```diff
@@ -0,0 +1,101 @@
...
+        return html.tostring(
+            html_document,
+        )
+
+    @api.model
+    def _set_page_break_style(self, table, add_page_break):
+        styles = StyleParser(table.get("style", ""))
+        styles.set("page-break-inside", "avoid")
+        if add_page_break:
+            styles.set("page-break-after", "always")
```
SirAionTech:
```suggestion
    def _set_page_break_style(self, table, break_page_after):
        styles = StyleParser(table.get("style", ""))
        styles.set("page-break-inside", "avoid")
        if break_page_after:
            styles.set("page-break-after", "always")
```
Seems more appropriate to what it is doing, in my opinion it also helps understanding https://github.com/OCA/reporting-engine/pull/826/files#r1426721337

SirAionTech (changes requested):
Thanks for all the answers, could you please squash everything in one commit?
I think that should be enough for adding a new module.

The other below comments are not blocking.

report_qweb_table_pagebreak/models/ir_qweb.py
petrus-v: done !

petrus-v:
> Thanks for all the answers, could you please squash everything in one commit? I think that should be enough for adding a new module.
> 
> The other below comments are not blocking.

Thanks for your reviews very well cumming :)

SirAionTech:
@petrus-v let me know when I can update my review:
> Navigate to Reviewers in the right sidebar and click image next to the reviewer's name whose review you'd like.

(ref. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review)

Right now I'm not sure if I have to update the review because you have made a change (https://github.com/OCA/reporting-engine/compare/39a5aec24d54349ba885c12652b690ec451c4477..fea8068f9da90856ace4b52f804606a9bac66750) and rolled it back shortly after (https://github.com/OCA/reporting-engine/compare/fea8068f9da90856ace4b52f804606a9bac66750..e27f64bbeb5cca59e5cd1db005edeace5c97f793), might be a mistake due to rebasing commits.

petrus-v:
sorry about the mess ! yeah I've merge your suggestion to change the variable name then squash all my commit but forgot the one merged using github suggestion so I cherry-pick your suggestion and merged it as well and force push again... ! so it's ready for review :sweat_smile:

SirAionTech (approved):

SirAionTech (approved):
Only code review

PR #827, opened 2023-12-14
[ADD] report_generate_helper

matthieusaison:
:wave:  @marielejeune, the get_report helper is here!


report_generate_helper/models/report_helper.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo.tools.safe_eval import safe_eval, time
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _get_report_converter(self):
+        return f"_render_{self.report_type.replace('-', '_')}"
+
+    def get_report(self, report_name):
```
lmignon: It would be nice to add a minimal documentation that at least specify that a call to the method will return a tuple filename / content....

lmignon (approved):
LGTM (Code review only)

report_generate_helper/models/report_helper.py
lmignon: A better name could be `generate_report`.... `get_report` could mean 'give me the report record for the given report name'

sbidoul: Is it intended that this method is public (xml-rpc callable?).

sbidoul (commented):
Some more documentation would be nice (use case, how to use).

report_generate_helper/models/report_helper.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import models
+from odoo.tools.safe_eval import safe_eval, time
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _get_report_converter(self):
+        return f"_render_{self.report_type.replace('-', '_')}"
```
sbidoul: I would add this method to another model that extends `ir.action.report` since that is where `report_type` is declared. Adding it to a subclass of `base` is misleading.

report_generate_helper/models/report_helper.py
sebastienbeau: No it's not necessary, I change to _generate_report

report_generate_helper/models/report_helper.py
sebastienbeau: fixed

sbidoul (approved):

PR #832, opened 2023-12-20
[17.0][MIG]  sql_request_abstract + bi_sql_editor: Migration to 17.0

nguyenminhchien:
Ref: BSRD-736


legalsylvain:
Hi @nguyenminhchien. Thanks for porting this module ! 
could you cherry pick the last commit from that V16 PR ? 
thanks ! 
https://github.com/OCA/reporting-engine/pull/823

nguyenminhchien:
> Hi @nguyenminhchien. Thanks for porting this module ! could you cherry pick the last commit from that V16 PR ? thanks ! #823

Updated

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -639,7 +615,7 @@ def _check_execution(self):
         field_ids = []
         for column in columns:
             existing_field = self.bi_sql_view_field_ids.filtered(
-                lambda x: x.name == column[1]
+                lambda x: x.name == column[1]  # noqa: B023
```
legalsylvain: Why this line. I dont understand B2023. What is dangerous here ?

nguyenminhchien: yep. "**B2023**" looks very dangerous. luckily we got `B023` here instead, and, it's a ruff failed code. any idea to fix it is welcome.

legalsylvain: > "B2023" looks very dangerous.

sorry, it's a typo.

> any idea to fix it is welcome.

Let's discuss here https://github.com/OCA/oca-addons-repo-template/issues/241

rven: It should be defined like this
`lambda x, c=column: x.name == c[1]`

https://docs.astral.sh/ruff/rules/function-uses-loop-variable/

gurneyalex (approved):

bi_sql_editor/models/bi_sql_view.py
legalsylvain: Thanks @rven ! 

@guyenminhchien could you use this syntax ? 

Thanks !

nguyenminhchien: updated

gurneyalex:
@legalsylvain can I merge this PR?

legalsylvain (changes requested):
Hi. @nguyenminhchien. thanks for porting this module ! 

Blocking point : I don't understand why you changed all the readonly state defined by python in readonly state defined by xml. Could you elaborate ? 

nitpicking : could you squash the two "MIG V17" commits ? it ll make the review easier.

thanks !

nguyenminhchien:
> Blocking point : I don't understand why you changed all the readonly state defined by python in readonly state defined by xml. Could you elaborate ?

Hi @legalsylvain `states` is not allowed anymore, so, `readonly` attribute in python is useless. 

> could you squash the two "MIG V17" commits ? it ll make the review easier.

There is only 1 "MIG V17" commit for each module, following the guideline from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#tasks-to-do-in-the-migration . I think we should not squash them just for making review easier .

legalsylvain (approved):
Thanks for the explanation.

/ocabot merge patch

PR #839, opened 2024-01-12
[IMP] bi_sql_editor: allow to generate form view + use sum / avg on tree view

legalsylvain:
* allow to generate form view, so It is possible to see data detail, when clicking on an item in a tree view.

* use group_operator to display sum / avg function in tree view, if defined.

image

CC : @AMEA-LYON, @nguyenminhchien

Note : add migration script to create form views and update tree views.


legalsylvain:
used in production since monthes.

/ocabot merge nobump

PR #831, opened 2023-12-20
[17.0][MIG] report xlsx helper: Migration to 17.0

FernandoRomera:

PR #847, opened 2024-01-19
[17.0][MIG] report_xlsx_helper

alexis-via:


legalsylvain (approved):
Diff review.

FernandoRomera:
Duplicate #831

alexis-via:
@FernandoRomera If you had written the EXACT module name in the title of your PR, I would have found it and it would have avoided duplicate work !

legalsylvain:
> @FernandoRomera If you had written the EXACT module name in the title of your PR, I would have found it and it would have avoided duplicate work !

Indeed. 
The error is common in the OCA PRs. I've done it several times. The PR titles generated by github are based on the last commit text, but github remove special chars like underscores, that should be added by contributors manually.
(``[17.0][MIG] report_xlsx_helper`` become ``17.0 MIG report_xlsx_helper``)
I don't know if there is a way to configure github differently.

rvalyi:
> > @FernandoRomera If you had written the EXACT module name in the title of your PR, I would have found it and it would have avoided duplicate work !
> 
> Indeed. 
> The error is common in the OCA PRs. I've done it several times. The PR titles generated by github are based on the last commit text, but github remove special chars like underscores, that should be added by contributors manually.
> (``[17.0][MIG] report_xlsx_helper`` become ``17.0 MIG report_xlsx_helper``)
> I don't know if there is a way to configure github differently.

it actually becomes "17.0 MIG report xlsx helper" that is removing the underscore chars so module name won't match. I think we cannot customize this is GitHub. Either we we warn the contributors, either we could have a bot detecting it and fixing the PR titles. But well many things probably more important to do first...

FernandoRomera:
> > > @FernandoRomera If you had written the EXACT module name in the title of your PR, I would have found it and it would have avoided duplicate work !
> > 
> > 
> > Indeed.
> > The error is common in the OCA PRs. I've done it several times. The PR titles generated by github are based on the last commit text, but github remove special chars like underscores, that should be added by contributors manually.
> > (`[17.0][MIG] report_xlsx_helper` become `17.0 MIG report_xlsx_helper`)
> > I don't know if there is a way to configure github differently.
> 
> it actually becomes "17.0 MIG report xlsx helper" that is removing the underscore chars so module name won't match. I think we cannot customize this is GitHub. Either we we warn the contributors, either we could have a bot detecting it and fixing the PR titles. But well many things probably more important to do first...

To solve this case, I think the simplest solution would be to allow EVERYONE who creates PR, to be able to specify "/ocabot migration report_xlsx_helper", only the first time. To overwrite it to be a PSC with permissions.

rvalyi:
> > > > @FernandoRomera If you had written the EXACT module name in the title of your PR, I would have found it and it would have avoided duplicate work !
> > > 
> > > 
> > > Indeed.
> > > The error is common in the OCA PRs. I've done it several times. The PR titles generated by github are based on the last commit text, but github remove special chars like underscores, that should be added by contributors manually.
> > > (`[17.0][MIG] report_xlsx_helper` become `17.0 MIG report_xlsx_helper`)
> > > I don't know if there is a way to configure github differently.
> > 
> > it actually becomes "17.0 MIG report xlsx helper" that is removing the underscore chars so module name won't match. I think we cannot customize this is GitHub. Either we we warn the contributors, either we could have a bot detecting it and fixing the PR titles. But well many things probably more important to do first...
> 
> To solve this case, I think the simplest solution would be to allow EVERYONE who creates PR, to be able to specify "/ocabot migration report_xlsx_helper", only the first time. To overwrite it to be a PSC with permissions. 

I don't think that would be a good solution because you generally do this command too late, after you already spent time doing the hard migration work and pushing your PR...

So for me the easiest is to warn OCA contributors about this (in the migration wiki page?) and possibly having a bot action fixing the bad PR titles.

PR #853, opened 2024-02-07
[16.0][FIX] report_async: replaced parent menu

ArnauCForgeFlow:
Changed the parent from Report Center menu item since odoo is using now spreadsheet_dashboard_menu_root instead of menu_board_root


Saran440 (approved):
Functional Test :+1:

lmignon (approved):

PR #856, opened 2024-02-22
[17.0][MIG] report_csv

pedroguirao:


report_csv/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "author": "Creu Blanca, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/reporting-engine",
     "category": "Reporting",
-    "version": "16.0.2.0.0",
+    "version": "17.0.2.0.0",
```
john-herholz-dt: Should be 17.0.1.0.0, right?

john-herholz-dt (commented):
Seems good except the version number.

report_csv/__manifest__.py
pedroguirao: Yes! thanks!

ridge-works-fs:
Looks good and I was able to generate a CSV report against another model not already demonstrated. 

Are there other tests that I can assist with before this PR is merged?

john-herholz-dt (approved):

yostashiro:
We have this refactoring PR https://github.com/OCA/reporting-engine/pull/864 for 16.0 which will affect this PR if accepted. Would be nice if we could get comments on the refactoring PR.

bealdav:
Yes @yostashiro @pedroguirao 

Could you depend on this new module report_format_option please ? Thanks

https://github.com/OCA/reporting-engine/pull/889

PR #842, opened 2024-01-13
[16.0][ADD] report_qweb_decimal_precision

yostashiro:
This module allows administrators to define the decimal precision of float fields for QWeb report presentation.

Functionality wise, this basically replaces https://github.com/OCA/reporting-engine/tree/16.0/report_qweb_decimal_place.

~Tests need to be added.~

Configuration:
image

Result:
image

@qrtl

PR #857, opened 2024-02-26
[16.0][ADD] report_qweb_field_option

AungKoKoLin1997:
Related: https://github.com/OCA/reporting-engine/pull/842

@qrtl QT5035


report_qweb_field_converter/models/ir_qweb.py
```diff
@@ -0,0 +1,36 @@
...
+    @api.model
+    def _get_field(
+        self, record, field_name, expression, tagName, field_options, values
+    ):
+        report_type = values.get("report_type")
+        if not report_type or report_type != "pdf":
+            return super()._get_field(
+                record, field_name, expression, tagName, field_options, values
+            )
+        qweb_recs = self.env["qweb.field.converter"].search(
```
yostashiro:
```suggestion
        qweb_recs = self.sudo().env["qweb.field.converter"].search(
```
We should either do this or make `qweb.field.converter` globally accessible. Please check and suggest which way is better. Portal users get 403 forbidden error when accessing their documents otherwise.

report_qweb_field_options/tests/test_report_qweb_field_options.py
```diff
@@ -0,0 +1,61 @@
...
+            self.test_currency, "rounding", False, False, {}, values
+        )
+        self.assertEqual(content, "0.010")
+
+        # Test with widget
+        self.qweb_options_rec.field_options = "{'widget': 'monetary'}"
+        _, content, _ = self.IrQweb._get_field(
+            self.test_currency, "rounding", False, False, {}, values
+        )
+        self.assertIn(self.env.company.currency_id.symbol, content)
```
yostashiro:
```suggestion
        self.assertIn("$", content)
```

report_qweb_field_options/tests/test_report_qweb_field_options.py
```diff
@@ -0,0 +1,61 @@
...
+from odoo.tests.common import TransactionCase
+
+
+class TestQwebFieldOptions(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.IrQweb = cls.env["ir.qweb"]
+        cls.test_currency = cls.env["res.currency"].create(
+            {"name": "Test Currency", "symbol": "T", "rounding": 0.01}
```
yostashiro:
```suggestion
            {"name": "Test Currency", "symbol": "$", "rounding": 0.01}
```

report_qweb_field_options/models/ir_qweb.py
```diff
@@ -0,0 +1,25 @@
...
+class IrQweb(models.AbstractModel):
+    _inherit = "ir.qweb"
+
+    @api.model
+    def _get_field(
+        self, record, field_name, expression, tagName, field_options, values
+    ):
+        if values.get("report_type") == "pdf":
+            options_recs = self.env["qweb.field.options"].search(
+                [("res_model_name", "=", record._name), ("field_name", "=", field_name)]
```
yostashiro: Please consider the company as well. Use `record.company_id` or `self.env.company`. Filter records where the company matches or is false.

report_qweb_field_options/models/qweb_field_options.py
```diff
@@ -0,0 +1,117 @@
...
+                ) from e
+            if not isinstance(field_options, dict):
+                raise ValidationError(
+                    _("Options must be a dictionary, but got %s") % type(field_options)
+                )
+
+    def _get_score(self, record):
+        self.ensure_one()
+        score = 1
+        if self.company_id:
```
yostashiro: Adjust the condition according to the upstream.

report_qweb_field_options/models/qweb_field_options.py
```diff
@@ -0,0 +1,117 @@
...
+            if record.company_id == self.company_id:
+                score += 1
+            else:
+                return -1
```
yostashiro: This case should not happen, so let's remove it.
```suggestion
```

yostashiro (changes requested):

yostashiro (approved):
LGTM.

nobuQuartile (approved):
LGTM
Functional test

pedrobaeza:
We should avoid plurals in module names.

yostashiro:
@pedrobaeza The module name comes from manipulating with `field_options` arg in `_get_field()`, so it makes sense to me (similar to the tone of web_m2x_options). But we are open to change the module name if that's better for the community.

pedrobaeza:
Yeah, `web_m2x_options` is a legacy name that comes from very old before setting this guideline. The guideline is set for avoiding constant typos when you start doing derivative modules, as the `s` in the middle of the module name is very confusing. Example `report_qweb_field_options_whatever`, that you mispell easily as `report_qweb_fields_option_whatever`, `report_qweb_fields_options_whatever`, etc.

AungKoKoLin1997:
I changed the module name.

PR #863, opened 2024-03-15
[FIX] base_comment_template: Search by model_ids.model

SirAionTech:
**Steps**:

1. Search comment templates using the domain [("model_ids.model", "=", "your_model")]

**Actual behavior**:
Nothing is found

**Expected behavior**:
Comments for model **your_model** are found

### DIFF
--- a/base_comment_template/README.rst
+++ b/base_comment_template/README.rst
@@ -169,6 +169,10 @@ Contributors
 
   * Achraf Mhadhbi <machraf@bloopark.de>
 
+*  `Aion Tech <https://aiontech.company/>`__:
+
+  *  Simone Rubino <simone.rubino@aion-tech.it>
+
 Maintainers
 ~~~~~~~~~~~
 
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -2,9 +2,11 @@
 # Copyright 2013-2014 Nicolas Bessi (Camptocamp SA)
 # Copyright 2020 NextERP Romania SRL
 # Copyright 2021-2022 Tecnativa - Víctor Martínez
+# Copyright 2024 Simone Rubino - Aion Tech
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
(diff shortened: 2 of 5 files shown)


SirAionTech:
The modified method comes from https://github.com/OCA/reporting-engine/commit/d9d2ef5b2731ba312e6274ea162caa2183d0ccbb and https://github.com/OCA/reporting-engine/commit/1f6bb667db1105d7cc33bbad83f27c2aeda7ccb7, maybe the authors @victoralmau and @AaronHForgeFlow would like to have a look?

SirAionTech:
Rebased on `16.0`

SirAionTech:
Rebased on `16.0`

AaronHForgeFlow:
How you do the searching? I tested in the comment template list and I can search by model successfully without this change.

SirAionTech:
> How you do the searching? I tested in the comment template list and I can search by model successfully without this change.

Thanks for having a look!
I'm searching with 
```py
self.env["base.comment.template"].search(
            [
                ("model_ids.model", "=", user_model_name),
            ]
        )
```
as it's done in the added test; you can also check that the test fails without the other changes

AaronHForgeFlow:
Sure, I just wanted to know if there was a way to test this as a user in Odoo. In any case as long as I can still search by model in the comment template list view in Odoo I am fine with this change.

AaronHForgeFlow:
Just one thing, I think you can add yourself as contributor in the CONTRIBUTORS.rst file, but changing the copyright requires big changes on the module,I don't think thi sis the case.

SirAionTech:
> Just one thing, I think you can add yourself as contributor in the CONTRIBUTORS.rst file

Right, thanks, I added myself to CONTRIBUTORS.
> changing the copyright requires big changes on the module,I don't think thi sis the case.

I usually add myself in any file I edit, no matter how big the change is; I believe it means _these people agree with its copyright_ and even if I added only one line to this file I need to agree with its copyright.

AaronHForgeFlow:
I don't mind it. But there's some discussion is the OCA in regards that: https://github.com/orgs/OCA/discussions/185

Indeed the copyright lines does not mean much: https://github.com/orgs/OCA/discussions/185#discussioncomment-10656634

PR #864, opened 2024-03-28
[16.0][REF][ADD] report_csv, report_format_option

AungKoKoLin1997:
This PR refactors the encoding part of the report_csv module into a new module to enable the use of encoding features across multiple reporting modules by adding this new module as a dependency.

@qrtl QT4120


report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+    encode_error_handling = fields.Selection(
+        selection=[("ignore", "Ignore"), ("replace", "Replace")],
+        help="If nothing is selected, CSV export will fail with an error message when "
+        "there is a character that fail to be encoded.",
+    )
+    show_encoding = fields.Boolean(compute="_compute_show_encoding")
+    line_ending = fields.Char(
+        help="Line ending to be applied to the generated report. e.g., crlf"
+    )
```
yostashiro, line 30: Shouldn't we have `show_line_ending` in a similar manner as `show_encoding`?

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,25 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class IrActionReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    encoding = fields.Char(
+        help="Encoding to be applied to the generated CSV file. " "e.g. cp932"
```
yostashiro:
```suggestion
        help="Encoding to be applied to the generated CSV file. e.g. cp932"
```

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+    encode_error_handling = fields.Selection(
+        selection=[("ignore", "Ignore"), ("replace", "Replace")],
+        help="If nothing is selected, CSV export will fail with an error message when "
+        "there is a character that fail to be encoded.",
+    )
+    show_encoding = fields.Boolean(compute="_compute_show_encoding")
+    line_ending = fields.Char(
+        help="Line ending to be applied to the generated report. e.g., crlf"
```
yostashiro: nitpick
```suggestion
        help="Line ending to be applied to the generated report. e.g. crlf"
```

report_format_option/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This is a technical module to add encoding fields to ir.actions.report model and it does nothing by itself.
+The module is expected to be a dependency of other reporting modules (e.g., report_csv).
```
yostashiro, line 2: nitpick
```suggestion
The module is expected to be a dependency of other reporting modules (e.g. report_csv).
```

yostashiro (approved):
Code review + checked presentation in runboat.

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,38 @@
...
+    encode_error_handling = fields.Selection(
+        selection=[("ignore", "Ignore"), ("replace", "Replace")],
+        help="If nothing is selected, CSV export will fail with an error message when "
+        "there is a character that fail to be encoded.",
+    )
+    show_encoding = fields.Boolean(
+        compute="_compute_show_encoding",
+        help="Technical field to control the visibility of the encoding field.",
+    )
+    line_ending = fields.Char(
```
yostashiro: @AungKoKoLin1997 Can we actually make it a selection field, pre-defining 'LF', 'CRLF' and 'CR'?

I think we can also extend method `_render_qweb_text()` in this module to handle the conversion of line ending and encoding.

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    show_line_ending = fields.Boolean(
+        compute="_compute_show_line_ending",
+        help="Technical field to control the visibility of the line ending field.",
+    )
+
+    def _compute_show_encoding(self):
+        """Extend this method to show the encoding field in the report form."""
+        for report in self:
+            report.show_encoding = False
```
yostashiro, line 35:
```suggestion
            report.show_encoding = False
            if report.report_type == "qweb_text":
                report.show_encoding = True
```

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _compute_show_line_ending(self):
+        """Extend this method to show the line ending field in the report form."""
+        for report in self:
+            report.show_line_ending = False
```
yostashiro, line 42:
```suggestion
            report.show_line_ending = False
            if report.report_type == "qweb-text":
                report.show_line_ending = True
```

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,66 @@
...
+        "there is a character that fail to be encoded.",
+    )
+    show_encoding = fields.Boolean(
+        compute="_compute_show_encoding",
+        help="Technical field to control the visibility of the encoding field.",
+    )
+    line_ending = fields.Selection(
+        [("LF", "LF (\\n)"), ("CRLF", "CRLF (\\r\\n)"), ("CR", "CR (\\r)")],
+        default="LF",
+        help="Select the type of line ending.",
```
yostashiro:
```suggestion
        [("crlf", "CRLF (\\r\\n)"), ("cr", "CR (\\r)")],
        help="Select the type of line ending in case the report needs to be output with other line ending than 'LF'.",
```

report_format_option/models/ir_actions_report.py
```diff
@@ -0,0 +1,66 @@
...
+            content_str = content_str.replace("\n", "\r\n")
+        elif report.line_ending == "CR":
+            content_str = content_str.replace("\n", "\r")
+        # Then encode with the desired encoding
+        encode_options = {}
+        if report.encode_error_handling:
+            encode_options["errors"] = report.encode_error_handling
+        # Re-encode the content with or without the errors parameter
+        content_encoded = content_str.encode(encoding, **encode_options)
+        return content_encoded, content_type
```
yostashiro:
```suggestion
        if not report.encoding and not report.line_ending:
            return content, content_type
        report = self._get_report(report_ref)
        content_str = content.decode("utf-8")
        if report.line_ending == "crlf":
            content_str = content_str.replace("\n", "\r\n")
        elif report.line_ending == "cr":
            content_str = content_str.replace("\n", "\r")
        # If specific encoding is set on the report, use it; otherwise, fallback to utf-8
        encoding = report.encoding or "utf-8"
        encode_options = {}
        if report.encode_error_handling:
            encode_options["errors"] = report.encode_error_handling
        content = content_str.encode(encoding, **encode_options)
        return content, content_type
```

yostashiro (changes requested):

yostashiro:
@AungKoKoLin1997 Can you please add tests to `report_format_option`?

report_format_option/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,12 @@
...
+
+* Encoding: set an encoding system (such as cp932)
+* Encode Error Handling: select 'Ignore' or 'Replace' as necessary.
+  * 'Ignore': in case of an encoding error, the problematic character will be removed from the exported file.
+  * 'Replace': in case of an encoding error, the problematic character will be replaced with '?' symbol.
+  * Leaving the field blank: in case of an encoding error, the report generation fails with an error message.
+* Line Ending: Select the type of line ending—'CRLF', 'CR' or leave blank, as necessary.
+  * 'CRLF': Stands for 'Carriage Return' + 'Line Feed'. It's used as a line ending in Windows, signifying the end of a line and the start of a new one.
+  * 'CR': Stands for 'Carriage Return'. It's used in classic Mac OS as a line ending, moving the cursor to the beginning of the line.
+  * Leaving this field blank will default to using 'LF' (Line Feed), which is the standard line ending in Unix/Linux systems, denoting the end of a line.
```
yostashiro: Let's make it less verbose (and more maintainable).

```suggestion
* Line Ending: Select the type of line ending, 'CRLF' or 'CR', as necessary.
  * 'CRLF': 'Carriage Return' + 'Line Feed' (Windows)
  * 'CR': 'Carriage Return' (classic Mac OS)
  * Leaving this field blank defaults to using 'LF' (Line Feed), the default behavior of Odoo.
```

yostashiro (commented):
Please apply the suggested change, squash commits and update the commit message with rationale.

yostashiro (approved):
Code review. LGTM.

report_format_option/demo/report_demo.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="model">res.partner</field>
+        <field name="report_type">qweb-text</field>
+        <field name="report_name">report_format_option.demo_report_template</field>
+        <field name="report_file">report_format_option.demo_report_template</field>
+        <field name="binding_model_id" ref="base.model_res_partner" />
+    </record>
+
+    <!-- Demo report template -->
+    <template id="demo_report_template">
+        <t t-call="web.html_container">
```
kanda999: Does this text report need to be html report?

kanda999 (approved):
Functinal and code review

report_format_option/demo/report_demo.xml
AungKoKoLin1997: Yes. You right. We don't need it.

HviorForgeFlow:
Question @pedrobaeza 
How should we merge this PR? Seems that report_csv needs some major version bump while the new module does not.

report_csv/__manifest__.py
```diff
@@ -8,9 +8,8 @@
     "category": "Reporting",
     "version": "16.0.2.0.0",
     "license": "AGPL-3",
-    "depends": ["base", "web"],
+    "depends": ["base", "web", "report_format_option"],
```
pedrobaeza, line 11: This is incorrect, as you can't change dependencies of a merged module.

pedrobaeza (changes requested):
Well, there's no debate, as you can't do such change.

report_csv/__manifest__.py
HviorForgeFlow: AFAIK the purpose is to split a feature to be reusable by other report modules out of the scope of the CSV.
There should be a way to achieve this change.

yostashiro: @pedrobaeza Do you mean that this goes against the stable policy and report_csv should stay as is?  In that case, we will try to do this in the 17.0 migration, and we may just propose to add report_format_option in 16.0 anyway.

pedrobaeza: Yes, you can't do it on an existing version, or you will break existing installations. I also see that the extracted code is not worth for such "reusing", but that's only my opinion.

yostashiro: Fine by me. We need report_format_option anyway for a client to apply some format conversions on text reports. That's why we proposed to split the common feature off report_csv to avoid redundancy. We will try to do this in 17.0.

@AungKoKoLin1997 Can you please close this PR and create another one just to add report_format_option?

bealdav:
Sad this PR was closed because report_format_option is a really good idea. But yes report_csv shouldn't depends on it for now @AungKoKoLin1997 @yostashiro  @HviorForgeFlow @pedrobaeza

AungKoKoLin1997:
@bealdav I created new PR only for report_format_option but I renamed the module name and fields name to avoid the UI breakout because of having same fields name in both `report_csv` and `report_text_format_option`.
https://github.com/OCA/reporting-engine/pull/884

PR #865, opened 2024-04-04
[16.0][IMP] remove python libs included in odoo requirements

CRogos:
We observed, that some installed python libraries are not in the version as they are listed in the odoo [requirements.txt](https://github.com/odoo/odoo/blob/16.0/requirements.txt) file. This causes some warnings and errors in the logs of our odoo.sh installation. 

The analysis showed, that this is due to some OCA repositories we've added to our installation (e.g. reporting-engine). These repositories contain python libs in their requirements which are not locked to a specific version. This triggers the installation of the newest version. 

In this PR, I've removed the libs, which are already installed by the odoo requirements file, as these libs should already be installed in a odoo compatible version.


CRogos:
@pedrobaeza may I ask you on your opinion on this? 

[server-tools ](https://github.com/OCA/server-tools/blob/16.0/requirements.txt) is the other repository I have the same problem with.

pedrobaeza:
I agree with this, as they are already implicit, but others say than "explicit is better than implicit". In our deployments, we simply ignore `requirements.txt` in OCA repos, and that's why this is not a problem for us. Why? Because we don't deploy whole repo, so these requirements is for the whole one, and may have libraries that we don't need.

CRogos:
Thanks for the quick response. 

We are using odoo.sh and include the oca repositories as git submodules. It looks like odoo.sh is automatically installing all requirements of all submodules. (even though the module which requires a lib is not installed)

Odoo SA explained to me, that they have to stick to certain libs, because newer libs require newer python versions and Odoo 16 supports older versions. I think this is a valid argument, even though some users use newer python versions and could therefore also use newer lib versions.  

I've already gave feedback to Odoo SA, that they should give an overview of non-standard library versions and warn, before updating the versions, but as you may know this might take a while until they have a fix for this problem.

pedrobaeza:
I run away from Odoo.sh...

legalsylvain (approved):
> , but others say than "explicit is better than implicit". 

Well. i guess that we don't explicit all lib. so better to have a rule that says : "add only requirements that are not present in Odoo". don't you think ? 

Can we merge ? Or someone see problem with such change ?

pedrobaeza (approved):

PR #868, opened 2024-04-11
ADD New Report engine base on pylatex

OmniaGit:
With this commit I would like to add my developed pylatex report to the main oca repo.

Thanks in advance
Matteo


len-foss:
Another approach to Latex reports: https://github.com/OCA/reporting-engine/pull/1064

PR #880, opened 2024-05-16
[16] sql_export : Add last_run datetime field

florian-dacosta:
Very small addition, which can become quite usefull when there are hundreds of queries and some may not be used anymore.
In case of migration for instance, it may allow to priorize which should be migrated or not, etc...

@legalsylvain

### DIFF
--- a/sql_export/wizard/wizard_file.py
+++ b/sql_export/wizard/wizard_file.py
@@ -63,7 +63,21 @@ class SqlFileWizard(models.TransientModel):
                 % {"name": sql_export.name, "date": date, "extension": extension},
             }
         )
-        sql_export.write({"last_run": fields.Datetime.now()})
+        # Bypass ORM to avoid changing the write_date/uid from sql query on a simple
+        # execution. This also avoid error if user has no update right on the
+        # sql.export object.
+        self.env.cr.execute(
+            """
+            UPDATE sql_export
+            SET last_execution_date = %s, last_execution_uid = %s
+            WHERE id = %s
+        """,
+            (
+                fields.Datetime.to_string(fields.Datetime.now()),
+                self.env.user.id,
+                sql_export.id,
+            ),
+        )
         action = {
             "name": "SQL Export",
             "type": "ir.actions.act_url",
--- a/sql_export/wizard/wizard_file.py
+++ b/sql_export/wizard/wizard_file.py
@@ -63,6 +63,7 @@ class SqlFileWizard(models.TransientModel):
                 % {"name": sql_export.name, "date": date, "extension": extension},
             }
         )
+        sql_export.write({"last_run": fields.Datetime.now()})
         action = {
             "name": "SQL Export",
             "type": "ir.actions.act_url",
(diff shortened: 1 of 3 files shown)


legalsylvain (commented):
Hi. Could you add it also in the form viewsql_export_view_form ?

sql_export/wizard/wizard_file.py
```diff
@@ -63,6 +63,7 @@ def export_sql(self):
                 % {"name": sql_export.name, "date": date, "extension": extension},
             }
         )
+        sql_export.write({"last_run": fields.Datetime.now()})
```
legalsylvain: It could be great to avoid to impact write date / write uid here  . in fact the user that executes the request doesn't "modify" the item.
Also for people that can read but not write on this model by IR.rule it will raise an access errror. 

I think that it should be done by SQL, then.

Don't you think ?

legalsylvain (changes requested):

sql_export/wizard/wizard_file.py
florian-dacosta: Thanks for the review, I did not think of that indeed !

I was first wondering if I should put it in an new table, linked to the query with 2 main motives : 
1) Avoid writing on the query (I did not think on the write_date/write_uid but rather on conccurent update if 2 users execute the query at the same time, but I finally thought it was quite unlikely, not so problematic)
2) To be able to keep an history so I can retrieve both the number of execution of a query + the evolution of the time to execute the query. (I have some big queries made by the users that sometimes can take a huge time...

Do you think it could be usefull ? Then maybe I'll go for this (and it will solve the issues you have highlighted)
But then maybe I have to think on a way to delete these history after a while...

Else, I'll just make the update by SQL and add a comment as an explanation, as suggested.

legalsylvain: Hi. In a simple way, I think we could just : 
- add two new field last_run_date ; last_run_uid 
- make the request by SQL instead of ORM

-> So that, the PR is valid, and can be merged.



Then, if we want to have some logs regarding execution of a function, we could : 
- create a generic module ``base_log_function`` that adds:
    - a decorator @log_function you can use for any function.
    - an history table as you propose with column execution_duration (and create_date / create_uid)
    - a cron to delete old logs

- add a custom module where you use the decorator for the function you want to log.
something like : 
````
    @log_function()
    def export_sql(self):
        return super().export_sql()
````

florian-dacosta:
> Hi. Could you add it also in the form viewsql_export_view_form ?

It does not seem necessary to me. I think it is more something you want on tree view to filter, order, and have a global vision rather than have this information on the form.
I guess we could put it on form view too, but there are no obvious place to put it either. The first group container exclusively configuration fields and the different tab are not adequate. 
The cleaner way would maybe to add one more tab for these 2 information, if you insist, but on my side I don't think we need it.

legalsylvain (approved):
thanks !

bguillot (approved):
Thanks

PR #884, opened 2024-05-28
[16.0][ADD] report_text_format_option

AungKoKoLin1997:
This module is designed to add encoding fields to the ir.actions.report model and is applied to text type reports.

This module was initially proposed in https://github.com/OCA/reporting-engine/pull/864, and `report_csv` depends on it. According to the discussion in this PR, this module should be used separately and specifically target text reports. Other modules (e.g., `report_xml`, `report_csv`) have their own encoding fields, and adding this module as a dependency could disrupt existing installations.

In this PR, I changed the module name from `report_format_option` to `report_text_format_option` to better reflect its primary focus on text report formatting. Additionally, I adjusted the encoding-related field names to avoid duplication with the `report_csv` encoding fields and to prevent breaking the UI.

@qrtl


bealdav (approved):

yostashiro (approved):
Functional test. 👍

HviorForgeFlow (approved):

PR #887, opened 2024-05-31
[FIX] bi_sql_editor : prevent to raise an error when creating a cron for a materialized view,

legalsylvain:
if it is the last day of the month.

Backport of recent improvment.

when the current date is May, 31 2024, the current prepare function will fail, when trying to cast a date with the value 2024-05-32.


rvalyi (approved):

yostashiro (approved):

legalsylvain:
/ocabot merge patch

(no need to port this patch in more recent version)

Thanks for your review !

PR #891, opened 2024-06-04
[13.0][IMP]bi_sql_editor: add parent menu as an editable parameter when creating UI elements

GuillemCForgeFlow:
When creating the UI for the SQL view, it is now possible to set a specific Parent Menu which will be used when creating the UI Menu for the report itself. The default value will be the `SQL Reports` menu.
Also, added the explanation on the `CONFIGURE.rst` file to have the description updated.

cc @ForgeFlow

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -153,7 +153,31 @@ class BiSQLView(models.Model):
     model_id = fields.Many2one(
         string="Odoo Model", comodel_name="ir.model", readonly=True
     )
+    # UI related fields
+    # 1. Editable fields, which can be set by the user (optional) before
+    # creating the UI elements
 
+    @api.model
+    def _default_parent_menu_id(self):
+        return self.env.ref("bi_sql_editor.menu_bi_sql_editor")
+
+    parent_menu_id = fields.Many2one(
+        string="Parent Odoo Menu",
+        comodel_name="ir.ui.menu",
+        required=True,
+        readonly=True,
+        default=lambda self: self._default_parent_menu_id(),
+        states={
+            "draft": [("readonly", False)],
+            "sql_valid": [("readonly", False)],
+            "model_valid": [("readonly", False)],
+        },
+        help="By assigning a value to this field before manually creating the "
+        "UI, you're overwriting the parent menu on which the menu related to "
+        "the SQL report will be created.",
+    )
+
+    # 2. Readonly fields, non editable by the user
     tree_view_id = fields.Many2one(
         string="Odoo Tree View", comodel_name="ir.ui.view", readonly=True
     )
@@ -492,7 +516,7 @@ class BiSQLView(models.Model):
         self.ensure_one()
         return {
             "name": self.name,
-            "parent_id": self.env.ref("bi_sql_editor.menu_bi_sql_editor").id,
+            "parent_id": self.parent_menu_id.id,
             "action": "ir.actions.act_window,%s" % self.action_id.id,
             "sequence": self.sequence,
         }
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (approved):
Interesting idea for usability

JordiMForgeFlow (approved):

GuillemCForgeFlow:
@LoisRForgeFlow FYI just pushed some improvements here:
- Add explanation on the `CONFIGURE.rst` file to have it updated based on the new change.
- Assigned the Parent Menu when creating the UI Menu to have it always assigned based on the current Parent.

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -490,9 +507,14 @@ def _prepare_action_name(self):
 
     def _prepare_menu(self):
         self.ensure_one()
+        parent_id = (
+            self.parent_menu_id.id
+            or self.env.ref("bi_sql_editor.menu_bi_sql_editor").id
+        )
+        self.write({"parent_menu_id": parent_id})
```
legalsylvain: a write in a prepare is not the best design IMO.

I guess we could simply remove that lines, and set 

```
parent_menu_id = fields.Many2one(
    ...
    required=True,
    default=lambda x: x._default_parent_menu_id(),
    help="The parent menu on which the menu related to "
        "the SQL report will be created."
    ...
)

def _default_parent_menu_id(self):
    return self.env.ref("bi_sql_editor.menu_bi_sql_editor")

```

This will be more explicit.

What do you think ?

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -490,9 +507,14 @@ def _prepare_action_name(self):
...
         return {
             "name": self.name,
-            "parent_id": self.env.ref("bi_sql_editor.menu_bi_sql_editor").id,
+            "parent_id": parent_id,
```
legalsylvain:
```suggestion
            "parent_id": self.parent_menu_id.id,
```

legalsylvain (changes requested):
Some minor remark inline. 

Otherwise, Nice improvment ! 

Thanks !

bi_sql_editor/models/bi_sql_view.py
GuillemCForgeFlow: Great, I totally agree that this looks like a cleaner solution. I will update 👍🏿

bi_sql_editor/models/bi_sql_view.py
GuillemCForgeFlow: same as above, thank you for the feedback 🙂

GuillemCForgeFlow:
> Some minor remark inline.
> 
> Otherwise, Nice improvment !
> 
> Thanks !

Just applied the changes you were suggesting + made the field required so that we don't have possible errors when creating the UI elements. 👍🏿

legalsylvain (approved):
Thanks for your contribution ! 

/ocabot merge minor

hum... I'd like to have some @robodoo r+ command to propagate that improvment on 14.0 / 15.0 / 16.0 / 17.0 ...

MiquelRForgeFlow:
> hum... I'd like to have some @robodoo r+ command to propagate that improvment on 14.0 / 15.0 / 16.0 / 17.0 ...

Don't worry, @GuillemCForgeFlow is a good person I am sure he will do it 🙂

GuillemCForgeFlow:
> > hum... I'd like to have some @robodoo r+ command to propagate that improvment on 14.0 / 15.0 / 16.0 / 17.0 ...
> 
> Don't worry, @GuillemCForgeFlow is a good person I am sure he will do it 🙂

Not sure about the first thing 🙂, but for sure about forward porting the changes, I can substitute the command this time 👍🏿

PR #892, opened 2024-06-04
[13.0][IMP]bi_sql_editor: many2one fields clickable on the list view if setting is enabled

GuillemCForgeFlow:
cc @ForgeFlow

### DIFF
--- a/bi_sql_editor/views/view_bi_sql_view.xml
+++ b/bi_sql_editor/views/view_bi_sql_view.xml
@@ -163,6 +163,10 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
                                             ('ttype', '=', 'many2one')]}"
                                     />
                                     <field
+                                        name="is_many2one_clickable"
+                                        attrs="{'invisible': ['|', ('field_description', '=', False), ('ttype', '!=', 'many2one')]}"
+                                    />
+                                    <field
                                         name="selection"
                                         attrs="{
                                         'invisible': [('ttype', '!=', 'selection')],
--- a/bi_sql_editor/models/bi_sql_view_field.py
+++ b/bi_sql_editor/models/bi_sql_view_field.py
@@ -74,6 +74,12 @@ class BiSQLViewField(models.Model):
         " a 'group by' option in the search view",
     )
 
+    is_many2one_clickable = fields.Boolean(
+        default=True,
+        help="If the field is of type Many2one and the option is enabled, the"
+        "field will be clickable on the list view.",
+    )
+
     index_name = fields.Char(string="Index Name", compute="_compute_index_name")
 
     graph_type = fields.Selection(string="Graph Type", selection=_GRAPH_TYPE_SELECTION)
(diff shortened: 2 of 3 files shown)


GuillemCForgeFlow:
@legalsylvain perhaps you can give me some feedback on this one as well 🙂

bi_sql_editor/views/view_bi_sql_view.xml
```diff
@@ -178,6 +178,10 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
                                         name="is_group_by"
                                         attrs="{'invisible': [('field_description', '=', False)]}"
                                     />
+                                    <field
+                                        name="is_many2one_clickable"
```
MiquelRForgeFlow: I would add this after many2one_model_id, as it's related to many2one field type too.

GuillemCForgeFlow: I have to say that sometimes you are right, and this is one of those few times. thank you for the input 🙂

MiquelRForgeFlow (approved):

BernatPForgeFlow (approved):
Functional test on Guillem's computer :computer: :slightly_smiling_face:

GuillemCForgeFlow:
@legalsylvain it would be nice to also get your input on this feature whenever you can 🙂

legalsylvain:
Hi thanks for the ping ! 

In fact, I don't understand what is widget="many2one". I rgrepped Odoo, and there is only a fiew occurences in the whole project. Coud you elaborate what is changing ?

GuillemCForgeFlow:
> Hi thanks for the ping !
> 
> In fact, I don't understand what is widget="many2one". I rgrepped Odoo, and there is only a fiew occurences in the whole project. Coud you elaborate what is changing ?

Sure, it is actually a widget added in odoo's core. See here: https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/fields/relational_fields.js#L96 and here https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/fields/field_registry.js#L72.

It's not widely used, but it gives the possibility for Many2one records to be clickable so you can access the specific record.

legalsylvain (changes requested):
Well, I just tested on runboat : 

**checkbox disabled :**
 
image

**checkbox enabled :** 

image

(Add possibility to click on the company to enter in the company form).


Well, I'm not very in favor of adding another extra setting (extra column in the configuration of the view of bi.sql.view.field model). 

Why not just install the OCA/web module web_tree_many2one_clickable
(https://github.com/OCA/web/tree/16.0/web_tree_many2one_clickable)

What do you think ? 

thanks !

GuillemCForgeFlow:
> Well, I just tested on runboat :
> 
> **checkbox disabled :**
> 
> image
> 
> **checkbox enabled :**
> 
> image
> 
> (Add possibility to click on the company to enter in the company form).
> 
> Well, I'm not very in favor of adding another extra setting (extra column in the configuration of the view of bi.sql.view.field model).
> 
> Why not just install the OCA/web module web_tree_many2one_clickable (https://github.com/OCA/web/tree/16.0/web_tree_many2one_clickable)
> 
> What do you think ?
> 
> thanks !

Understandable, I agree we can still use the https://github.com/OCA/web/tree/16.0/web_tree_many2one_clickable module if needed. We can close it 👍🏿

PR #893, opened 2024-06-06
[14.0][FW][IMP]bi_sql_editor: add parent menu as an editable parameter when creating UI elements

GuillemCForgeFlow:
When creating the UI for the SQL view, it is now possible to set a specific Parent Menu which will be used when creating the UI Menu for the report itself. The default value will be the `SQL Reports` menu. Also, added the explanation on the `CONFIGURE.rst` file to have the description updated.

FW of https://github.com/OCA/reporting-engine/pull/891

cc @ForgeFlow


legalsylvain (approved):
/ocabot merge minor

thanks for porting this patch.

PR #899, opened 2024-06-17
[IMP] Add the possibility to use a custom name if no docids, through context

lmarion-source:


report_xlsx/controllers/main.py
```diff
@@ -90,6 +90,9 @@ def report_download(self, data, context=None, token=None):
                             report.print_report_name, {"object": obj, "time": time}
                         )
                         filename = "%s.%s" % (report_name, "xlsx")
+                ctx = json.loads(context)
+                if not docids and "custom_report_name" in ctx.keys():
+                    filename = ctx["custom_report_name"]
```
acsonefho: As discussed, IMO the best solution is to extract the filename into a dedicated function. Then that make you an inherit point for any custom module.

acsonefho (changes requested):

acsonefho (approved):

PR #900, opened 2024-06-17
[16][FIX] sql_export_mail : avoid trying to send mail for a not valid query

florian-dacosta:


legalsylvain (approved):
Make sense ! nice catch.

/ocabot merge patch

PR #903, opened 2024-06-26
[17.0][IMP]bi_sql_editor: scheduled action periodicity settings

GuillemCForgeFlow:
refactor `Action Settings` page to have all settings, renamed to `Settings`. Added the settings to be able to customize the periodicity of the Scheduled Action that will refresh the Materialized view.

cc @ForgeFlow

### DIFF
--- a/bi_sql_editor/README.rst
+++ b/bi_sql_editor/README.rst
@@ -117,11 +117,16 @@ to make reporting depending on the current companies of the user.
 
    |image4|
 
--  Before applying the final step, you will need to add a specific
-   Parent Menu to use when creating the UI Menu for the report. By
-   default, it will be set with the ``SQL Views`` menu, which can be
-   changed before creating the UI elements in order to have the report
-   accessible from a different place within Odoo.
+-  Before creating the UI elements: you can modify two specific settings
+   based on your needs:
+
+   -  **Parent Menu**: Apply a Parent Menu to use for when creating the
+      UI elements. By default, it will be set with the ``SQL Views``
+      menu, which can be changed in order to have the report accessible
+      from a different place within Odoo.
+   -  **Scheduled Action periodicity**: By going to the Settings page,
+      you can customize the frequency for which you want to run the
+      Scheduled Action that will refresh the Materialized view.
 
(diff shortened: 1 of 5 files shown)


GuillemCForgeFlow:
Hi @legalsylvain, I'd like to get your input on this whenever you have some time 
thank you in advance 🙂

legalsylvain:
Hi @GuillemCForgeFlow. Thanks for contributing. 
I don't get the interest of this change. (Except that all cron information are managed in the SQL view.)
Could you elaborate ?
Thanks !

GuillemCForgeFlow:
> Hi @GuillemCForgeFlow. Thanks for contributing. I don't get the interest of this change. (Except that all cron information are managed in the SQL view.) Could you elaborate ? Thanks !

Sure, the use case for this is that the periodicity of the refresh was too low. It could be changed at Scheduled Action level, but each time the view gets recreated (because the SQL definition changes) the Scheduled Action gets back to the original parameters.

I thought it would be great to have a general settings tab, in order to fulfill this kind of information if it needs to be changed based on specific needs. We can add more settings' groups on that tab if required in the future.

What do you think?

legalsylvain (changes requested):
> but each time the view gets recreated (because the SQL definition changes) the Scheduled Action gets back to the original parameters

You're right. That's the point. It should be changed, for exemple, disabling the cron instead of deleting it. What do you think ?
It avoid to duplicate fields in both modèls.

GuillemCForgeFlow:
> > but each time the view gets recreated (because the SQL definition changes) the Scheduled Action gets back to the original parameters
> 
> You're right. That's the point. It should be changed, for exemple, disabling the cron instead of deleting it. What do you think ? It avoid to duplicate fields in both modèls.

I agree, that seems to solve the issue for both of us. I will adjust the changes then to have this in place. Thank you for the input 😄

GuillemCForgeFlow:
closing, I will reference the implemented solution once done 👍🏿 @legalsylvain

GuillemCForgeFlow:
Implemented agreed changes on https://github.com/OCA/reporting-engine/pull/928

PR #904, opened 2024-07-01
[12.0][IMP] report_async: add schedule_date field

RicardCForgeFlow:
Change schedule_time to Datetime field

@ForgeFlow

### DIFF
--- a/report_async/models/report_async.py
+++ b/report_async/models/report_async.py
@@ -2,7 +2,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
 
 import base64
-from datetime import datetime, timedelta
 from odoo import api, fields, models, _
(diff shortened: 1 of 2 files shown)


HviorForgeFlow (approved):

report_async/models/report_async.py
```diff
@@ -71,7 +70,7 @@ class ReportAsync(models.Model):
         help="List all files created by this report background process",
     )
 
-    schedule_time = fields.Char(string='Schedule time')
+    schedule_time = fields.Datetime(string='Schedule Time')
```
MiquelRForgeFlow: This would need migration scripts. Can we adapt the code in a way to avoid this change?

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow (approved):

PR #905, opened 2024-07-04
[12.0][IMP] report_async: change job_ids to One2many with queue.job

RicardCForgeFlow:
Due to performance considerations, the string-based search has been replaced. We have optimized this by directly linking rec.job_ids to queue.job records using a direct relationship defined as a One2many field in the report.async model.

cc @ForgeFlow

### DIFF
--- a/report_async/migrations/12.0.1.2.0/post-migration.py
+++ b/report_async/migrations/12.0.1.2.0/post-migration.py
@@ -1,0 +1,19 @@
+from openupgradelib import openupgrade
+
+
+def populate_report_async_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE queue_job qj
+        SET report_async_id = ra.id
+        FROM report_async ra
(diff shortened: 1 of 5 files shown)


MiquelRForgeFlow (approved):

report_async/migrations/12.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def populate_report_async_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE queue_job qj
+        SET report_async_id = ra.id
+        FROM report_async ra
+        WHERE qj.func_string LIKE '%%report.async(%%' || ra.id || ',%%'
```
MiquelRForgeFlow:
```suggestion
        WHERE qj.func_string LIKE '%report.async(' || ra.id || ',%'
```

RicardCForgeFlow:
Should merge as minor or major :smiley:

PR #909, opened 2024-07-23
16.0 imp bi sql editor extra info display

legalsylvain:
### Before
- configuration "parent Odoo menu" is not in the correct place. (is in extra information tabs)
- group are bad displayed
- space is not optimized in the top of the form 

image

### After

image


quentinDupont (approved):
Code review :heavy_check_mark:

PR #835, opened 2024-01-08
[16.0] FIX] report_py3o: fix error in py3o_context_parser

jdoutreloux:

PR #910, opened 2024-07-24
[16.0] FIX] report_py3o: fix error in py3o_context_parser

ThomasBinsfeld:
Supersedes #835 

The definition of the _format_date function is `def _format_date(self, value, lang_code=False, date_format=False):` no env must be passed when calling it.


benwillig (approved):

ThomasBinsfeld:
@OCA/reporting-engine-maintainers this one is a bug and should be merged asap.

HviorForgeFlow (approved):

PR #840, opened 2024-01-12
[17.0][MIG] report_qweb_element_page_visibility

sanderlienaerts:

PR #911, opened 2024-07-24
[17.0][MIG] report_qweb_element_page_visibility

Algosse:
Tested on V17 and worked just fine.
Also compared the custom 'subst' function with the [official one](https://github.com/odoo/odoo/blob/f80094f43bcc464cc99fba6aa59f64659fe659b6/addons/web/views/report_templates.xml#L242) and they are the same


ioans73:
@Algosse https://github.com/OCA/reporting-engine/pull/840

Algosse:
... --' 
I'm dumb sorry...

PR #912, opened 2024-07-24
[16.0][FIX] report_substitute: adding change of main method

matiasperalta1:
The main method that belongs to odoo changed in the migration and was not taken into account when migrating, however the name had been changed in the super methods. This never allowed me to enter the method and there were errors when rendering templates with reports that had substitution rules.


matiasperalta1:
@HviorForgeFlow Hi! Could you review this PR?

HviorForgeFlow (approved):

PR #920, opened 2024-08-19
[16.0][MIG] kpi: Migration to 16.0

matiasperalta1:


kpi/models/kpi.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    next_execution_date = fields.Datetime(readonly=True)
+    value = fields.Float(compute="_compute_display_last_kpi_value")
+    color = fields.Text(compute="_compute_display_last_kpi_value")
+    last_execution = fields.Datetime(compute="_compute_display_last_kpi_value")
+    kpi_type = fields.Selection(
+        [
+            ("python", "Python"),
+            ("local", "SQL - Local DB"),
```
jok-adhoc: according L122 this must be 'local' --> 'sql'

kpi/models/kpi.py
```diff
@@ -0,0 +1,187 @@
...
+            ("external", "SQL - External DB"),
```
jok-adhoc: remove

kpi/models/kpi.py
```diff
@@ -0,0 +1,187 @@
...
+            ("week", "Week"),
+            ("month", "Month"),
+        ],
+        required=True,
+        default="day",
+    )
+
+    next_execution_date = fields.Datetime(readonly=True)
+    value = fields.Float(compute="_compute_display_last_kpi_value")
+    color = fields.Text(compute="_compute_display_last_kpi_value")
```
jok-adhoc: because we're removing threshold to simplify the module, color should not exist

kpi/models/kpi.py
```diff
@@ -0,0 +1,187 @@
...
+            ("hour", "Hour"),
+            ("day", "Day"),
+            ("week", "Week"),
+            ("month", "Month"),
+        ],
+        required=True,
+        default="day",
+    )
+
+    next_execution_date = fields.Datetime(readonly=True)
```
jok-adhoc, line 74: default=fields.Date.today ?

kpi/views/kpi_views.xml
```diff
@@ -0,0 +1,158 @@
...
+                            <field name="history_ids" readonly="1" nolabel="1" />
+                        </page>
+                        <page string="Computation" groups="kpi.group_kpi_manager">
+                            <group col="6">
+                                <field name="periodicity" colspan="3" />
+                                <field name="periodicity_uom" colspan="3" />
+                                <field name="next_execution_date" colspan="3" />
+                                <separator string="KPI Computation" colspan="6" />
+                                <newline />
+                                <field name="kpi_type" colspan="2" />
```
jok-adhoc, line 121: It would be better to have colspan equal to "6" in order to have both kpi_type and kpi_code one bellow the other. Another option could be creating two diff groups 

With colspan = "2" , the code's label is in the same line as kpi_type and the editable field bellow, what makes the label confusing.

PR #927, opened 2024-08-30
[FIX] report_substitute: handle reports without xml_id

lef-adhoc:
The change replaces self._for_xml_id(action_report.xml_id) with action_report.read()[0] to avoid errors when reports do not have an xml_id, such as when they are created via the interface.


sbejaoui (approved):
LGTM, code review

HviorForgeFlow (approved):

PR #928, opened 2024-09-02
[13.0][IMP]bi_sql_editor: archive Scheduled Action to not recreate when resetting the SQL View

GuillemCForgeFlow:
Based on the agreed solution: https://github.com/OCA/reporting-engine/pull/903#pullrequestreview-2273835493. It is best to archive the Scheduled Action every time you set back to the draft state. Then, when creating the SQL view, if it is a materialized view, unarchive the Scheduled Action if already exists, otherwise create it from scratch.

Then, if the SQL View is deleted, also make sure to remove the Scheduled Action if there is one assigned as it has not been done in the `button_set_draft` method

cc @ForgeFlow

### DIFF
--- a/bi_sql_editor/models/bi_sql_view.py
+++ b/bi_sql_editor/models/bi_sql_view.py
@@ -18,8 +18,9 @@ _logger = logging.getLogger(__name__)
 
 @api.model
 def _instanciate(self, model_data):
-    """ Return a class for the custom model given by
-    parameters ``model_data``. """
+    """Return a class for the custom model given by
+    parameters ``model_data``."""
+
     # This monkey patch is meant to avoid create/search tables for those
     # materialized views. Doing "super" doesn't work.
     class CustomModel(models.Model):
@@ -281,6 +282,8 @@ class BiSQLView(models.Model):
                     "If you want to delete them, first set them to draft."
                 )
             )
+        # Remove Scheduled Actions, as they were not removed in `button_set_draft`
+        self.mapped("cron_id").unlink()
         return super(BiSQLView, self).unlink()
 
     def copy(self, default=None):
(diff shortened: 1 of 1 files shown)


bi_sql_editor/models/bi_sql_view.py
```diff
@@ -324,7 +329,7 @@ def button_set_draft(self):
             sql_view.pivot_view_id.unlink()
             sql_view.search_view_id.unlink()
             if sql_view.cron_id:
-                sql_view.cron_id.unlink()
```
legalsylvain, line 327: As you remove unlink here, you should add if the bi.sql.view is deleted I think.

Apart that point, LGTM. Very ligther solution, and nice improvment ! thanks !

GuillemCForgeFlow: sure, thank you for noticing 👍🏿 just applied the changes, we should be good now 😄

legalsylvain (approved):
Thanks !

bi_sql_editor/models/bi_sql_view.py
```diff
@@ -281,6 +282,9 @@ def unlink(self):
                     "If you want to delete them, first set them to draft."
                 )
             )
+        for sql_view in self:
+            if sql_view.cron_id:
+                sql_view.cron_id.unlink()
```
legalsylvain: Nitpicking : The 3 lines could be replaced by seld.mapped("cron_id").unlink() i think.

GuillemCForgeFlow: no problem, I also think it looks cleaner like this 👍🏿

legalsylvain:
/ocabot merge minor

Thanks !

legalsylvain:
@GuillemCForgeFlow could you port this patch upstream ? (17.0 and ideally 16.0)

thanks !

GuillemCForgeFlow:
> @GuillemCForgeFlow could you port this patch upstream ? (17.0 and ideally 16.0)
> 
> thanks !

sure, I will try to allocate some time during the next few days, will keep you posted 👍🏿

GuillemCForgeFlow:
Hi @legalsylvain, I've seen that you already added some similar changes back in the day from v16 on: https://github.com/OCA/reporting-engine/commit/680bfdfdada245716307e49687312717960d2f88

So I guess we should just leave this change in v13 😄

legalsylvain:
> Hi @legalsylvain, I've seen that you already added some similar changes back in the day from v16 on: https://github.com/OCA/reporting-engine/commit/680bfdfdada245716307e49687312717960d2f88

ahah ! I didn't remember.

PR #931, opened 2024-09-14
[17.0] [MIG] report_py3o: Migration to 17.0

bug2point0:


angelmoya (approved):
:+1:  LGTM tested on runbot ok

report_py3o/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "ODT -> PDF, ODT -> DOC, ODT -> DOCX, ODS -> ODS, etc.)",
+    "version": "17.0.1.0.0",
+    "category": "Reporting",
+    "license": "AGPL-3",
+    "author": "XCG Consulting, ACSONE SA/NV, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/reporting-engine",
+    "depends": ["web"],
+    "external_dependencies": {
+        "python": ["py3o.template", "py3o.formats"],
+        "deb": ["libreoffice"],
```
Tardo, line 15: More info: https://github.com/odoo/odoo/blob/d759264c6e9b7a8cc8b6cffdc5509df654cd434c/odoo/modules/module.py#L508

```suggestion
        "bin": ["libreoffice"],
```

I think this should be changed on all versions... and other modules...

Tardo (changes requested):

report_py3o/__manifest__.py
bug2point0: [odoo docs](https://www.odoo.com/documentation/17.0/developer/reference/backend/module.html?highlight=external_dependencies) are pretty clear on this. Nice catch.

Tardo: Umm... looks like OCA "unofficial" way (i don't know where is documented) to handle dependencies in CI env: https://github.com/OCA/oca-ci/blob/c10c90835d3baf6576faaae03f593099b9995009/bin/oca_install_addons#L16

Perhaps a solution could be to use ```bin``` to have the Odoo check and ```deb``` to indicate how to install the package with the 'apt package manager'?

ping @sbidoul

sbidoul: You can add `bin` if you want, but please keep `deb` so libreoffice is installed in ci and runboat.

That said, I'm not sure libreoffice is a hard dependency. IIRC it's optional when using `report_py3o_fusion_server`, so in that case we should not add `bin`.

sersanchus:
I think I've found an issue. There is a demo report for res.users model that is defined in the file 'demo/res_users_report_py3o.xml'. However, it is not added in the 'demo' section of the '\_\_manifest\_\_.py' but rather in 'data'. Therefore, it is always installed even if demos are not required for the module.

report_py3o/__manifest__.py
Tardo: @bug2point0 I think there should be both “bin” and “deb”. 
As I see it, actually it all depends on which module you give more importance to 'report_py3o_fusion_server' vs 'report_py3o'... Everything else is a mess of dependencies (if you choose X with Y but you have Z then disable W... :/). At least from what I could see.

You decide, if you choose only “deb” it seems right to me... and if you choose “bin” and “deb” the same. Little more I can say.

Tardo:
@bug2point0 please, can you rebase? thanks!

Tardo:
@bug2point0 please, can you rebase? thanks! :B ... and please, check the commits because it looks like you're touching a lot more files than the migration requires.

pilarvargas-tecnativa:
superseded by https://github.com/OCA/reporting-engine/pull/1021

PR #933, opened 2024-10-03
[17.0][FIX] report_xlsx: Preserve custom context by using context from report instead of user context

DucTruongKomit:
**Steps to Reproduce:**
1. Try to set with_context when print the XLSX report. for example with_context(uom_id=1).
2. when print the report the context uom_id never show up.

**Observer Behavior**

**Expected Behavior**

### DIFF
--- a/report_xlsx/static/src/js/report/action_manager_report.esm.js
+++ b/report_xlsx/static/src/js/report/action_manager_report.esm.js
@@ -33,7 +33,7 @@ registry
                     url: "/report/download",
                     data: {
                         data: JSON.stringify([url, action.report_type]),
-                        context: JSON.stringify(env.services.user.context),
+                        context: JSON.stringify(actionContext),
                     },
                 });
             } finally {


cuongnmtm (approved):
The code LGTM.

I think you should adjust the commit msg to include the module name. Maybe "[FIX] report_xlsx: preserve custom context by using context from report instead of user context"

DucTruongKomit:
> The code LGTM.
> 
> I think you should adjust the commit msg to include the module name. Maybe "[FIX] report_xlsx: preserve custom context by using context from report instead of user context"

Done, thank you!

DucTruongKomit:
> The code LGTM.
> 
> I think you should adjust the commit msg to include the module name. Maybe "[FIX] report_xlsx: preserve custom context by using context from report instead of user context"

Thank you. I did the update.

cuongnmtm (approved):
LGTM. Code review.

HviorForgeFlow (approved):

PR #935, opened 2024-10-09
[17.0][IMP] report_qweb_parameter: Remove warnings

victoralmau:
Remove warning Found deprecated directive `@t-raw`

`WARNING devel odoo.addons.base.models.ir_qweb: Found deprecated directive @t-raw='docs.env["ir.qweb"].check_length(docs[0].vat, 10, False)[:10]' in template 237. Replace by @t-out, and explicitely wrap content in `Markup` if necessary (which likely is not the case)`

`WARNING devel odoo.addons.base.models.ir_qweb: Found deprecated directive @t-raw='docs.env["ir.qweb"].check_length(docs[0].company_registry, False, 10)' in template 237. Replace by @t-out, and explicitely wrap content in `Markup` if necessary (which likely is not the case)`

Execute pre-commit to remove the existing warning.

`WARNING prod py.warnings: /opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_module.py:178: DeprecationWarning: XML declarations in HTML module descriptions are deprecated since Odoo 17, report_qweb_parameter can just have a UTF8 description with not need for a declaration.`

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #937, opened 2024-10-10
[18.0][MIG] sql_request_abstract: Migration to 18.0

thienvh332:
### Ported PRs
- https://github.com/OCA/server-tools/pull/2291
- https://github.com/OCA/server-tools/pull/2317


legalsylvain:
hi. could you take a look on red CI ? thanks !

thienvh332:
Hi @legalsylvain ,
I updated it. Could you have a look?

checklog-odoo.cfg
```diff
@@ -0,0 +1,3 @@
+[checklog-odoo]
+ignore=
+    WARNING.*0 failed, 0 error\(s\).*
```
legalsylvain, line 3: Why this addition ?

thienvh332: This module originally had no test case and the new oca-ci test case was raising errors even with WARNING. So I added it to avoid being labeled as ERROR

legalsylvain: hum. something is wrong if CI is red if there are no test, it is a valid use case. I think that something should be fixed in checklog-odoo tools instead. 

@sbidoul : what do you think ?

sbidoul: > something should be fixed in checklog-odoo tools instead.

We don't want to special case this in checklog-odoo, because we have had situations where for some reason tests did not run had that error was actually helpful.

This problem will only occur when the first module migrated has no test, so in practice this should not occur often, I guess.

sebalix (approved):

rvalyi (approved):
Ok. Now this was a pretty trivial migration, not sure it is fair to be listed in the contributors and put other credits. Let the module authors be the judges...

HviorForgeFlow:
trivial or not, an effort to move this module to v18 has been done. Commit history will put everybody in their place. Let's merge it :)

HviorForgeFlow:
@thienvh332 pre-commit failed

legalsylvain (approved):

PR #941, opened 2024-10-16
[18.0][ADD] sheet_dataframe_process module: convert file to polars dataframe and…

bealdav:
… process it

image

Based on an imported spreadsheet like file, this module allow to create
Polars dataframe and process them according to rules in order to:

- display filtered data
- obtain another dataframe with only the expected data to use in Odoo
- import data into Odoo TO IMPLEMENT

Why dataframe ?

- a dataframe is a kind of in-memory dataset on which you can operate
- you can operates on your entire dataset a bit like with a database but in memory: you don't need to iterate on each line to perform operations
- the operations are powerful: filter, add column resulting from calculation , select a subset of data

Why Polars ?

- performance: code in rust
- environment consideration
- dynamic project

TODO

- [x] implement populate in file config
- [ ] check matching columns
- [x] check matching according to partner
- [ ] raise if required column missing
- [ ] raise if date is invalid


sheet_dataframe_process/README.rst
```diff
@@ -0,0 +1,99 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/reporting-engine-18-0/reporting-engine-18-0-sheet_dataframe_process
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/reporting-engine&target_branch=18.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+Based on an imported spreadsheet like file, this module allow to create
```
rvalyi: s/allow/allows/

PR #942, opened 2024-10-23
[12.0][ADD] report_pdf_zip_download

AungKoKoLin1997:
This module generates one PDF report per record, and the reports are compiled into a ZIP file if the user prints multiple records at once. If the user prints a single record, it will follow Odoo's standard behavior and will generate just a single PDF file.

@qrtl QT4419

### DIFF
--- a/report_pdf_zip_download/__manifest__.py
+++ b/report_pdf_zip_download/__manifest__.py
@@ -1,0 +1,17 @@
+# © 2017 Creu Blanca
+# Copyright 2024 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Report PDF ZIP Download",
+    "category": "Report",
+    "version": "12.0.1.0.0",
+    "author": "Quartile, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/reporting-engine",
+    "license": "AGPL-3",
+    "depends": ["web"],
+    "data": [
+        "views/ir_actions_report_views.xml",
+        "views/web_client_templates.xml",
+    ],
+    "installable": True,
+}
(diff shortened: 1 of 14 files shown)


batch_report_pdf_zip/models/ir_actions_report.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    is_zip = fields.Boolean(
```
yostashiro: To be a bit more descriptive.
```suggestion
    zip_download = fields.Boolean(
```

yostashiro (commented):
How about `report_pdf_zip_download` as the module name?

report_pdf_zip_download/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# Copyright 2024 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Report PDF ZIP Download",
+    "category": "Report",
+    "version": "12.0.1.0.0",
+    "author": "Quartile, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/reporting-engine",
+    "license": "AGPL-3",
+    "depends": ["base"],
```
yostashiro:
```suggestion
    "depends": ["web"],
```

report_pdf_zip_download/static/src/js/action_manager_report.js
```diff
@@ -0,0 +1,88 @@
...
+    var core = require("web.core");
+    var ActionManager = require("web.ActionManager");
+    var crash_manager = require("web.crash_manager");
+    var framework = require("web.framework");
+    var session = require("web.session");
+    var _t = core._t;
+
+    ActionManager.include({
+
+        _downloadReportZIP: function (url, actions) {
```
yostashiro, line 16: I feel that there must be a better way to handle zip downloads, but I don't have a concrete suggestion...

report_pdf_zip_download/__manifest__.py
AungKoKoLin1997: I updated the code.

yostashiro (approved):
Code review.

kanda999 (approved):
Functional review

PR #873, opened 2024-04-27
[16.0][ADD] sql_export_delta

hbrunn:
This module extends the functionality of ``sql_export`` to allow you to
only export changes between export runs.

I added a [second commit](https://github.com/OCA/reporting-engine/pull/873/commits/0e13eda3b1b4d6790514e20fd6d0ca5cf34380c7) that allows users to keep the results of an sql export, that's useful independently of deltas, and very useful with deltas.

PR #955, opened 2024-12-05
[16][FIX] sql_export : rights for no "system" users

florian-dacosta:
Since this PR https://github.com/OCA/reporting-engine/pull/873
There is access issues if the user is not par of the `base.group_system` group.

Because of this https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_attachment.py#L460

I also add myself as maintainer since I am the original author and still participate actively to its maintenance/migration.

@legalsylvain @hbrunn


legalsylvain (approved):
makes sense.

hbrunn (approved):
thanks!

PR #957, opened 2024-12-19
[14.0] report_xlsx: Fix currency format

mmequignon:
Some xlsx readers doesn't like mixed formats on the same column.
Here we sometimes have the sign on the left, sometimes on the right.
(change introduced in this commit on `account-financial-report` https://github.com/OCA/account-financial-reporting/commit/f088aa6405aac8527cfbd6da2d63dd6aebaf7951#diff-8237378723eb85bb5c6d6e16f339b55840b762029442371284fe1ad71d08c1e2
When it happens, we get `###` displayed (even though value is ok)
-> No, this isn't because of the column being too narrow.

image

Among the tested readers, I was able to reproduce on `excel` (asking to repair the file), `onlyoffice` (not proposing any fix) and `gsheet`.

Reverting to a simple number format.

Also, the only report on which this method is `open items` in `account_financial_report`, which also displays the currency in a dedicated column. This information is redundant.

### DIFF
--- a/report_xlsx/report/report_abstract_xlsx.py
+++ b/report_xlsx/report/report_abstract_xlsx.py
@@ -92,11 +92,9 @@ class ReportXlsxAbstract(models.AbstractModel):
         return self.env[self.env.context.get("active_model")].browse(ids)
 
     def _report_xlsx_currency_format(self, currency):
-        """Get the format to be used in cells (symbol included).
+        """Get currency format to be used in cells.
         Used in account_financial_report addon"""
-        s_before = currency.symbol if currency.position == "before" else ""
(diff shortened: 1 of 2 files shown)


simahawk:
@victoralmau you introduced this change if I'm not mistaken. Could you please have a look?

victoralmau (changes requested):
The existing change (actually introduced in 13.0 https://github.com/OCA/reporting-engine/pull/647 +1 year ago) is intentional, the idea is to not need to add a currency column and display the value with the currency (as appropriate). I have (and no one has reported it so far) not had (or known) any problems viewing the generated excel files, but I think the right thing to do is to make it compatible and not remove the existing utility.

PR #962, opened 2025-01-08
[FIX] report_csv: make it working if docids is False

maisim:
len() is used on docids, but in some cases, docids is False wich raised a TypeError

### DIFF
--- a/report_csv/models/ir_report.py
+++ b/report_csv/models/ir_report.py
@@ -55,7 +55,7 @@ class ReportAction(models.Model):
         report_sudo = self._get_report(report_ref)
         report_model_name = "report.%s" % report_sudo.report_name
         report_model = self.env[report_model_name]
-        res_id = len(docids) == 1 and docids[0]
+        res_id = docids[0] if docids and len(docids) == 1 else None
         if not res_id or not report_sudo.attachment or not report_sudo.attachment_use:
             return report_model.with_context(
                 **{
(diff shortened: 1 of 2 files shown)


remi-filament (approved):
Indeed, as per code in controller, docids is None by default, and in this case you will get a TypeError.

See : 
https://github.com/OCA/reporting-engine/blob/a2240527a2f65f578970fcb5d57e5969a622af9d/report_csv/controllers/main.py#L26
Called from https://github.com/OCA/reporting-engine/blob/a2240527a2f65f578970fcb5d57e5969a622af9d/report_csv/controllers/main.py#L80-L82 if not docids

petrus-v (approved):
could be nice to add unittest to avoid regressions !

other wise code looks good to me

report_csv/models/ir_report.py
```diff
@@ -55,7 +55,7 @@ def _render_csv(self, report_ref, docids, data):
         report_sudo = self._get_report(report_ref)
         report_model_name = "report.%s" % report_sudo.report_name
         report_model = self.env[report_model_name]
-        res_id = len(docids) == 1 and docids[0]
+        res_id = docids[0] if docids and len(docids) == 1 else None
```
maisim, line 58: just replaced 
`res_id = (len(docids) == 1 and docids[0]) if docids else False`
with
`res_id = docids[0] if docids and len(docids) == 1 else None`

wich is more readable

maisim: @petrus-v @remi-filament 
Just added a test and the little change above

HviorForgeFlow (approved):

PR #964, opened 2025-01-15
[17.0][FIX] report_qweb_parameter: out and raw doesn't work with conditionals

ValentinVinagre:


ValentinVinagre:
@etobella @HaraldPanten

etobella (approved):
/ocabot merge patch

Fast tracking

PR #979, opened 2025-02-18
[IMP] sql_export_mail, add partner as export email recipient

joepsanders:
Adds partner_ids additional to user_ids to send the result of an sql_export by email.
- Doesn't allow the usage of dynamic parameters "user_id" or "company_id" in combination with sending the report to a partner_id. This to prevent unwanted / unfiltered results to recipients.
- Adds the email addresses of selected partners to the email addresses of selected users (if any)

### DIFF
--- a/sql_export_mail/models/sql_export.py
+++ b/sql_export_mail/models/sql_export.py
@@ -19,6 +19,14 @@ class SqlExport(models.Model):
         help="Add the users who want to receive the report by e-mail. You "
         "need to link the sql query with a cron to send mail automatically",
     )
+    mail_partner_ids = fields.Many2many(
+        "res.partner",
+        "mail_partner_sqlquery_rel",
+        "sql_id",
+        "partner_id",
+        help="Add the partners who wants to receive the report by e-mail. You "
+        "need to link the sql query with a cron to send a mail automatically",
+    )
     cron_ids = fields.Many2many(
         "ir.cron",
         "cron_sqlquery_rel",
@@ -138,13 +146,43 @@ class SqlExport(models.Model):
                 if not user.email:
                     raise UserError(_("The user does not have any e-mail address."))
 
+    @api.constrains("mail_partner_ids", "query")
+    def _check_mail_partner(self):
+        for export in self:
+            if export.mail_partner_ids and (
+                "%(company_id)s" in export.query or "%(user_id)s" in export.query
+            ):
+                raise UserError(
+                    _(
+                        "A query that uses the company_id or user_id parameter "
+                        "cannot be directly sent to a partner."
+                    )
+                )
+            missing_email_partners = export.mail_partner_ids.filtered(
+                lambda partner: not partner.email
+            )
+            if missing_email_partners:
+                raise UserError(
+                    _(
+                        "Missing email address for partner(s): %(names)s",
+                        names=", ".join(missing_email_partners.mapped("name")),
+                    )
+                )
+
     def get_email_address_for_template(self):
         """
-        Called from mail template
+        Called from mail template.
+        Collects email addresses from both users and partners.
         """
         self.ensure_one()
         if self.env.context.get("mail_to"):
             mail_users = self.env["res.users"].browse(self.env.context.get("mail_to"))
+            mail_partners = self.env["res.partner"]
         else:
             mail_users = self.mail_user_ids
-        return ",".join([x.email for x in mail_users if x.email])
+            mail_partners = self.mail_partner_ids
+        email_addresses = set(
+            mail_users.mapped("email") + mail_partners.mapped("email")
+        )
+        email_addresses.discard(False)
+        return ",".join(email_addresses)
--- a/sql_export_mail/views/sql_export_view.xml
+++ b/sql_export_mail/views/sql_export_view.xml
@@ -31,6 +31,14 @@
                             colspan="2"
                         />
                         </group>
+                        <group string="Partners Notified by e-mail">
+                            <field
+                            name="mail_partner_ids"
+                            nolabel="1"
+                            widget="many2many_tags"
+                            colspan="2"
+                        />
+                        </group>
                         <group string="Crons" groups="base.group_system">
                             <field
                             name="cron_ids"
--- a/sql_export_mail/static/description/index.html
+++ b/sql_export_mail/static/description/index.html
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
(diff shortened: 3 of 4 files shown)


sql_export_mail/models/sql_export.py
```diff
@@ -138,13 +146,38 @@ def check_mail_user(self):
                 if not user.email:
                     raise UserError(_("The user does not have any e-mail address."))
 
+    @api.constrains("mail_partner_ids")
+    def check_mail_partner(self):
```
sanderlienaerts: This should be a private method, `_check_mail_partner_ids`

sql_export_mail/models/sql_export.py
```diff
@@ -138,13 +146,38 @@ def check_mail_user(self):
...
+            mail_partners = self.mail_partner_ids
+
+        email_addresses = []
+        for user in mail_users:
+            if user.email:
+                email_addresses.append(user.email)
+        for partner in mail_partners:
+            if partner.email:
+                email_addresses.append(partner.email)
+        return ",".join(email_addresses)
```
sanderlienaerts, line 188: Simplify (should probably also be private):
```python
   def _get_email_address_for_template(self):
        """"""
        self.ensure_one()
         if self.env.context.get("mail_to"):
             mail_users = self.env["res.users"].browse(self.env.context.get("mail_to"))
             mail_partners = self.env["res.partner"]
         else:
             mail_users = self.mail_user_ids
             mail_partners = self.mail_partner_ids
        email_addresses = mail_users.mapped("email") + mail_partners.mapped("email")
        return ",".join(email_addresses)
```

sql_export_mail/models/sql_export.py
```diff
@@ -138,13 +146,38 @@ def check_mail_user(self):
...
+    def check_mail_partner(self):
+        for export in self:
+            if "%(company_id)s" in export.query or "%(user_id)s" in export.query:
+                raise UserError(
+                    _(
+                        "A query that uses the company_id or user_id parameter "
+                        "cannot be directly sent to a partner."
+                    )
+                )
+            for partner in export.mail_partner_ids:
```
sanderlienaerts: Generalize:
```python 
if export.mail_partner_ids.filtered(lambda partner: not partner.email):
    raise .....
```
or specify what partner is missing an email address:
```python 
missing_email_partners = export.mail_partner_ids.filtered(lambda partner: not partner.email):
if missing_email_partners:
   raise UserError(_("Missing email address for partner(s): %(names)s", names= ", ".join(missing_email_partners.mapped("name")))
```

sql_export_mail/views/sql_export_view.xml
```diff
@@ -31,6 +31,14 @@
                             colspan="2"
                         />
                         </group>
+                        <group string="Partners Notified by e-mail">
```
sanderlienaerts, line 34: `Partners Notified by Email`

sanderlienaerts (changes requested):

sanderlienaerts (approved):

joepsanders:
squashed PR feedback @legalsylvain, could you review this addition?

sql_export_mail/__manifest__.py
```diff
@@ -1,7 +1,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "SQL Export Mail",
-    "version": "16.0.2.0.2",
+    "version": "16.0.2.1.0",
```
legalsylvain: no need to bump the version manually, as there is no migration script.

sql_export_mail/models/sql_export.py
```diff
@@ -138,13 +146,38 @@ def check_mail_user(self):
                 if not user.email:
                     raise UserError(_("The user does not have any e-mail address."))
 
+    @api.constrains("mail_partner_ids")
```
legalsylvain: I guess the constrains should be raised, when query is changed. Or did I missed something ?

sql_export_mail/models/sql_export.py
```diff
@@ -138,13 +146,38 @@ def check_mail_user(self):
...
+        Called from mail template.
+        Collects email addresses from both users and partners.
         """
         self.ensure_one()
         if self.env.context.get("mail_to"):
             mail_users = self.env["res.users"].browse(self.env.context.get("mail_to"))
+            mail_partners = self.env["res.partner"]
         else:
             mail_users = self.mail_user_ids
-        return ",".join([x.email for x in mail_users if x.email])
```
legalsylvain, line 150: there was a test if user email was defined (or not). Why is the test removed ?

legalsylvain (changes requested):
hi @joepsanders Thanks for this addition.
Some minor remarks inline. Otherwise LGTM.

sql_export_mail/__manifest__.py
joepsanders: isn't this version update following the OCA guidelines? https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#12version-numbers

"y (Minor): Increments when new features are added that do not break backward compatibility. A module upgrade will likely be necessary."

joepsanders: isn't this version update following the OCA guidelines? https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#12version-numbers

"y (Minor): Increments when new features are added that do not break backward compatibility. A module upgrade will likely be necessary."

sql_export_mail/models/sql_export.py
joepsanders: The mapped function skips the missing emails by design: https://github.com/odoo/odoo/blob/3443fc707e9794f5fd329b1ca0cb1f95a47cda0b/odoo/models.py#L6401

joepsanders:
> hi @joepsanders Thanks for this addition. Some minor remarks inline. Otherwise LGTM.

Hi @legalsylvain, thanks for your swift review! See my answers / questions to your remarks and a fixup commit (to be squashed).

sql_export_mail/models/sql_export.py
legalsylvain: I don't think so.

Just tested in a shell : 

<code>
(Console)
>>> env["res.partner"].create({"name": "bob"}).mapped("email")
[False]
>>> ",".join([False])
TypeError: sequence item 0: expected str instance, bool found
</code>

sql_export_mail/__manifest__.py
legalsylvain: That's not my point.
The point is : Do not propose a change of version. When merging, maintainers will call "ocabot merge patch /minor / major". And then it will bump the version at this step.
Otherwise, if another PR bump also the version in the meantime, you'll face a conflict.

sql_export_mail/models/sql_export.py
joepsanders: I stand corrected, thanks ;). See 2nd fixup commit

sql_export_mail/__manifest__.py
joepsanders: thanks for explaining!

HviorForgeFlow:
ping @legalsylvain

legalsylvain (approved):

HviorForgeFlow:
@joepsanders please squash fixup commits

joepsanders:
> @joepsanders please squash fixup commits

Done!

HviorForgeFlow (approved):

PR #837, opened 2024-01-10
[16.0][MIG] report_py3o_signer

benwillig:
Based on https://github.com/OCA/reporting-engine/pull/836

PR #987, opened 2025-03-10
[16.0][MIG] report_py3o_signer

len-foss:
Note: directly integrates https://github.com/OCA/reporting-engine/pull/988
(split for ease of review and migration)


yvaucher:
Duplicate of #837

PR #988, opened 2025-03-10
[16.0][REF] report_qweb_signer: make signing functions reusable

len-foss:
This is a simple refactoring to make it easy to reuse the signing code.
Without this, `report_py3o_signer` would have to duplicate a lot of code. 
See: https://github.com/OCA/reporting-engine/pull/987

### DIFF
--- a/report_qweb_signer/models/ir_actions_report.py
+++ b/report_qweb_signer/models/ir_actions_report.py
@@ -211,9 +211,8 @@ class IrActionsReport(models.Model):
             self._signer_endesive(params, p12, pdf, pdfsigned, passwd)
         return pdfsigned
 
-    def _render_qweb_pdf(self, report_ref, res_ids=None, data=None):
-        report = self._get_report(report_ref)
-        certificate = self._certificate_get(report, res_ids)
+    def _read_attached_signed_content(self, report, res_ids, certificate):
+        signed_content, ext = False, False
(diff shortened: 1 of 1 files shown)


report_qweb_signer/models/ir_actions_report.py
```diff
@@ -223,29 +222,43 @@ def _render_qweb_pdf(self, report_ref, res_ids=None, data=None):
                     res_ids,
                 )
                 return signed_content, "pdf"
+        return signed_content, ext
+
+    def _render_qweb_pdf(self, report_ref, res_ids=None, data=None):
+        report = self._get_report(report_ref)
+        certificate = self._certificate_get(report, res_ids)
```
pedrobaeza, line 229: I think this should go in the method `_sign_pdf_and_attach`, don't you think?

len-foss: Hi @pedrobaeza :-)
You really need to methods, one that goes before the super, and one that goes after.
Since the super is not the same if you're operating on another report type, that wouldn't work. (I'm not 100% sure I interpreted your remark in the correct way, let me know if I'm wrong)

pedrobaeza (approved):
OK, I see the certificate is also used in the method before the super. Forget it.

carlos-lopez-tecnativa (approved):
LGTM

PR #992, opened 2025-03-14
[16.0][IMP] report_qweb_field_option: remove odoo_test_helper from manifest

AungKoKoLin1997:
This PR removes the `odoo_test_helper` from the `manifest ` to avoid raising error when install the module in the environment that doesn't install this library and this library is needed for the test cases. So, I added `odoo_test_helper` in `test-requirements.txt` instead.

@qrtl QT5035


yostashiro (approved):
LGTM.

pedrobaeza (approved):
/ocabot merge patch

Issue #254, opened 2018-11-14
[RFC] using Weasyprint to generate pdf


legalsylvain:
Hello, 

This issue is here to talk about the opportunity to integrate Weasyprint into odoo. feel free to ask question.

Reference : 
- [original talk on OCA mailing list]( https://odoo-community.org/groups/contributors-15/contributors-119658)
- official website : https://weasyprint.org/
- FR : [video on pyconfr]( https://www.youtube.com/watch?v=vbXLX1eUrls&index=3&list=PL_dGzOUQ0cyiBA8tJXDDZPnx7ityIYZpy)
- FR : [presentation on linuxfr.org website](https://linuxfr.org/news/premiere-version-stable-pour-weasyprint)

CC : @liZe, alias guillaume Ayoub, CEO of Kozea that has developed the librairy.

hbrunn:
yes, writing `report_qweb_weasyprint_renderer` is on my list for a while already. You can just take https://github.com/OCA/reporting-engine/pull/159 and replace chrome by weasyprint.

This will get us beautiful reports with full page backgrounds and proper page margin boxes, amongst others. What I think will be a little bit of a challenge is to add some compatibility layer so that we can actually replace wkhtmltopdf with it. But I think for starters the flag (or maybe a second selection field `renderer` so that modules can just add to that) as I did in the chrome version will be fine, then we can already use it and add some extra module coming with the proper templates to really replace wkhtmltopdf

alexis-via:
I see in the documentation that weasyprint is able to produce PDF with attachment, which is really cool (cf https://weasyprint.readthedocs.io/en/latest/api.html) ! But it doesn't seem to have an option to produce PDF/A (which would make it the perfect tool to generate compliant Factur-X invoices !)... i didn't see anything about it in the documentation. @liZe do you confirm that ?

yajo:
The docs say:

> It is based on various libraries but not on a full rendering engine like WebKit or Gecko.

IMHO it should be better then to try to use a full rendering engine, one that matches a good browser. That should lead to least surprises. Odoo v12+ uses chrome/chromium to run JS tests. Maybe we can use it to render PDFs. Of course this is a general idea, and I have not hacked on it.

Also I'd appreciate it if you open the proposal for Odoo master, since this is a general problem. It can be backported then. They claim to be open to suggestions: https://github.com/odoo/odoo/issues/21255#issuecomment-347124683

hbrunn:
[...]

PR #993, opened 2025-03-15
[16.0][ADD] report_qweb_weasyprint_renderer

hbrunn:
This module enables you to use [WeasyPrint](https://weasyprint.org) to render PDFs, which allows for more print related HTML features than using Odoo's standard wkhtmltopdf.

Have a look at the [demo report](https://github.com/user-attachments/files/19265957/WeasyPrint.demo.report.full.custom.pdf).

It's not the main focus of this module, but I also experimented a bit with supporting reports made for wkhtmltopdf. [This](https://github.com/user-attachments/files/19265959/INV_2025_00004-1.pdf) is an invoice generated after changing reporting engine to weasyprint. Not good enough yet, but also not horrible. As said, that's not interesting for the moment anyways.

Usage
=====

To use this module, you need to set ``qweb_pdf_engine`` to ``weasyprint``

```xml
   <record id="demo_report" model="ir.actions.report">
       <field name="qweb_pdf_engine">weasyprint</field>
   </record>
```

In your report template, you have multiple possibilities:

1. ``t-call`` the template ``report_qweb_weasyprint_renderer.html_wrapper`` and add your html code inside it. This takes care of defining metadata, includes some of Odoo's standard assets like fonts and fontawesome and defines some helpful classes:

    `header`/`footer`: divs with this class go into the page margin. Put those divs *before* any content

    `page`/`topage`: elements with this class get the current page number (`page`) or the page count (`topage`) prepended

    Note that bootstrap classes are not available, because bootstrap doesn't go well with weasyprint.

2. Generate a complete html tree (starting with `<html>`), this is the most flexible way to use this addon
3. ``t-call`` one of the the existing ``web.internal_layout`` and ``web.external_layout`` - this tries to have similar behavior as wkhtmltopdf reports wrt headers, footers, document style, but isn't well supported as of yet and very slow because this includes a lot of assets that weasyprint doesn't deal with speedily.

Footers are implemented as margin boxes, which are of fixed height. You can edit the header and footer height per report in the WeasyPrint tab.

Known issues / Roadmap
======================

* improve compatibility layer for wkhtmltopdf
* add weasyprint-specific versions of the layout styles (bold, boxed etc)


rvalyi:
cc @legalsylvain I think you experimented with weasyprint reports for Odoo, may be you could provide feedback here...
cc @sebastienbeau

hbrunn:
Sylvain experimented with my code from v12, and I'll be happy for the feedback ~~as soon as I found out what causes the problem on runbot~~

legalsylvain:
Hi. 

> [This](https://github.com/user-attachments/files/19265959/INV_2025_00004-1.pdf) is an invoice generated after changing reporting engine to weasyprint. Not good enough yet, but also not horrible

great ! 

> may be you could provide feedback here...

My work was different from the holger one. I made a PoC to replace wkhtmltopdf, not to use weasyprint as an extra module. It was a fork of Odoo. (It looks I lost my branch...) (See [comment](https://github.com/OCA/reporting-engine/issues/254#issuecomment-1625283117 and https://github.com/OCA/reporting-engine/issues/254#issuecomment-1626347401))

I'm convinced that Odoo should leave wkhtmltopdf, and use another framework. I promoted weasyprint, because it works great, it is a simple python library, very easy to install, and I personaly know the funder of this tool. That said, I gave up because I couldn't find a serious contact at Odoo to talk about it. Also because all the commenters said that weasyprint was horribly slow, without understanding why. Indeed run ``weasyprint https://www.odoo.com`` is slow, but the problem is in the second word of this command, not in the first one. ;-)

For the time being, In Odoo, since wkhtmltopdf has been archived in 2023 : 

- first try : [fork of wkhtmltopdf](https://github.com/odoo/wkhtmltopdf/). Failed. Repo archived on August 2024.
- Second try : Since September 2024, creation of the project [Paper-Muncher](https://github.com/odoo/paper-muncher/). Today in a "Not yet ready for use state", 6 months later.
- let's see how it goes...

rvalyi:
Hello @legalsylvain as for Odoo making weasyprint the main engine of Odoo, I think the problem is not only the relative slowness. It's also that the frontend and document templating tends to converges (think portal, framework...) and weasyprint will miss the latest CSS tricks used by the frontend and might also miss the Javascript engine (OWL templates might become more ubiquitous).

That being said, having weasyprint as an optionnal rendering engine is very welcome.

Finally, Odoo went for yet another internal tool, this time a huge blob of C++: Paper Muncher: https://github.com/odoo/paper-muncher
No matter the skills and good intentions I don't put too much faith into Odoo maintaining a decent CSS engine and I fear customizing Odoo reports will keep being a nightmare if they make this new engine the default. Future will tell it. Browsers are simply the biggest codebase around in the world and even the rendering engine is something crazy to maintain.

cc @pedrobaeza @sebastienbeau @alexis-via

hbrunn:
please don't make this a general ranting-about-wkhtmltopdf-and-odoo-sa-thread and forget reviewing the code over it.

my vision with this is:

- get this pr merged
- add a v16 version of https://github.com/OCA/reporting-engine/pull/159
- have addons evolve like `invoice_report_weasyprint`, `account_financial_report_chrome` so that eventually you can drop wkhtmltopdf in your installation if you like

in my experience if stuff gets traction in the oca, eventually odoo sa comes up with their own, usually slightly more convoluted version of it

pedrobaeza:
I won't follow this path, sorry. I trust paper-muncher will be a thing soon, and anyway, having the overhead of first designing replacement reports for each of the standard ones, and then extending them is not what I want. It's the same of report_py3o. You can see that it doesn't get traction many years later.

rvalyi (approved):
the code looks great an I tested it on Runboat.

Sure it's really slower: like 6 seconds for an invoice on wkhtmltopdf vs some 20 seconds with Weasyprint (disclaimer: mental count). It renders a bit differently but that's not horrible and that's expected.

For reference:

wkhtmltopdf:


Weasyprint:



@pedrobaeza not sure you got it, but here this module is only about being able to choose Weasyprint as an alternative engine for a given report. It's not a generic replacement... As for py3o, it has its downside sure, but it certainly did a great service for so many companies during a decade. I'm sure there will be use cases for Weasyprint or other report engines.

And yes, I'm curious to see Chrome or other browser based engines...

pedrobaeza:
I understood that it's not a drop-in replacement, as you have proved with your screenshots, and that's why it needs some extra modules replacing the standard reports with "optimized" ones, so I stand in my previous comment.

PR #994, opened 2025-03-17
[FIX] report_substitute: avoid write over the mail template

augusto-weiss:

PR #991, opened 2025-03-11
[17.0][MIG] report_qweb_signer: Migration to version 17.0

carlos-lopez-tecnativa:
TT55397

@Tecnativa @pedrobaeza @chienandalu @etobella could you please review this?

PR #1001, opened 2025-03-27
[FIX] report_substitute: avoid write over the mail template

augusto-weiss:


augusto-weiss:
FW of https://github.com/OCA/reporting-engine/pull/994

amkarthik (approved):

augusto-weiss:
@sbejaoui @jjscarafia @HviorForgeFlow Hello guys, could you review this fordwar-port of #994 ? Thanks a lot

zaoral (approved):

zaoral:
Thank you for the manual FW @augusto-weiss !

I confirm this is needed in the 17 and 18 versions, and it is an important fix, because it will avoid a concurrency problem. For example, when validating and auto-sending invoices.
 
Best Regards

HviorForgeFlow:
@pedrobaeza could you take a quick look here? A solution does not come me up. It looks like an error with a cryptography dependency but, in the past the related module was merge with any problem.

https://github.com/OCA/reporting-engine/pull/991

pedrobaeza:
Try to pin the library version

PR #1002, opened 2025-03-28
[15.0][IMP] report_py3o: add new mixin model

pilarvargas-tecnativa:
Extending a model with this mixin allows to set a py3o template at model and record level.

Using a common report it is possible to print a custom py3o template by setting that template at model and record level.
A use case could be to want to print the invoice reports of a contact by customising those reports using a py3o template. You would extend the res.partner model to be able to add the template to that contact and have a related on the invoices to print that custom report.

cc @Tecnativa TT55652

@chienandalu @pedrobaeza @victoralmau please review

### DIFF
--- a/report_py3o/models/py3o_report.py
+++ b/report_py3o/models/py3o_report.py
@@ -164,10 +164,16 @@ class Py3oReport(models.TransientModel):
         if report_xml.py3o_template_id.py3o_template_data:
             # if a user gave a report template
             tmpl_data = b64decode(report_xml.py3o_template_id.py3o_template_data)
-
         else:
             tmpl_data = self._get_template_fallback(model_instance)
-
+            if (
+                not tmpl_data
+                and "py3o_template_id" in model_instance._fields
+                and model_instance.py3o_template_id
+            ):
+                tmpl_data = b64decode(
+                    model_instance.py3o_template_id.py3o_template_data
+                )
         if tmpl_data is None:
             # if for any reason the template is not found
             raise TemplateNotFound(_("No template found. Aborting."), sys.exc_info())
--- a/report_py3o/models/py3o_template.py
+++ b/report_py3o/models/py3o_template.py
@@ -1,4 +1,5 @@
 # Copyright 2013 XCG Consulting (http://odoo.consulting)
+# Copyright 2025 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from odoo import fields, models
 
@@ -22,3 +23,12 @@ class Py3oTemplate(models.Model):
         required=True,
         default="odt",
     )
+
+
+class Py3oTemplateMixin(models.AbstractModel):
+    _name = "py3o.template.mixin"
+    _description = "Py3o template mixin"
+
+    # Extending a model with this mixin allows to establish a py3o template at model
+    # and record level.
+    py3o_template_id = fields.Many2one("py3o.template", "Template")
--- a/report_py3o/README.rst
+++ b/report_py3o/README.rst
@@ -50,6 +50,22 @@ This reporting engine is an alternative to `Aeroo <https://github.com/aeroo-comm
 .. contents::
    :local:
 
+Use Cases / Context
+===================
+
+The mixin in the report_py3o module allows to set up a Py3o template at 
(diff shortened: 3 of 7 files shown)


report_py3o/models/py3o_template.py
```diff
@@ -22,3 +23,12 @@ class Py3oTemplate(models.Model):
         required=True,
         default="odt",
     )
+
+
+class Py3oTemplateMixin(models.AbstractModel):
```
chienandalu, line 28: Better use a different file and make sure you load it before the others in the `__init__` to avoid issues

pilarvargas-tecnativa: A separate file causes a recurrence error
```
ImportError: cannot import name 'py3o_template_mixin' from partially initialized module 'odoo.addons.report_py3o.models' (most likely due to a circular import) (/opt/odoo/auto/addons/report_py3o/models/__init__.py)
```

victoralmau (approved):

pilarvargas-tecnativa:
@carlos-lopez-tecnativa can you take a look?

carlos-lopez-tecnativa (approved):

report_py3o/models/py3o_report.py
```diff
@@ -165,6 +165,12 @@ def get_template(self, model_instance):
             # if a user gave a report template
             tmpl_data = b64decode(report_xml.py3o_template_id.py3o_template_data)
 
+        elif (
```
pedrobaeza: Shouldn't you give more priority to model templates instead of the report one?

pilarvargas-tecnativa: The main purpose of the module is to generate reports with custom templates. When a report is defined for a model with a specific template or template path, it should take precedence over any template set at model level.

The main reason is that the model template is intended to generate reports in a general way, allowing dynamic styling without a predefined template in the report. However, when a specific template is defined for a report, it is assumed that it should be applied in priority.

Therefore, giving priority to the report template ensures that the intended layout is used for that specific case, while the model template acts as a flexible fallback when there is no report template defined.

pedrobaeza: I understand it the contrary:

- You have specific template on each record. Example: the certification template for each survey.
- If not defined at survey level, you have a generic template on the report that is the "generic certification".

pilarvargas-tecnativa: But then you would be changing the behaviour of the module. I can have two reports, or more, defined for the same model, as in the example you give me, generate an internal certificate with a specific template defined in the report and another one without template because it depends on the template defined in the model which can be different for each record. If I prioritise on the template of the model then it would be useless to be able to define a report with the informed template because it will always take first the template of the model.

pedrobaeza: It's not changing the behavior, as by default, there won't be any value in the model, so it fallbacks to the report template. But when a value is set on the model records, it should have priority.

pilarvargas-tecnativa: Again, you lose the functionality of being able to create other reports with a defined template. Moreover, looking at the code again, it should be the last check after fallback. For example, I define in res.partner a template at model level to have a dynamic report that takes the template of the model to simply get the data of a contact with a design that will depend on the contact (although it is not a case applicable to something real). Now I define another report for res.partner with a template defined to print a credential with a red layout and I define another report for res.partner with another template defined to print another type of credential in blue. Those two reports have the template defined in the py3o_template_id field or in py3o_template_fallback, as in this example: https://github.com/OCA/reporting-engine/blob/748884f88144611b51870105e9dc6a32d1b01a26/report_py3o/demo/report_py3o.xml#L13C22-L13C44

image
image

Now we print the red report, it will check the model first and if it has a template defined (which it would have) it would not print the red report. I don't know if I'm explaining myself well

It really should stay that way:

```     
if report_xml.py3o_template_id.py3o_template_data:
    # if a user gave a report template
    tmpl_data = b64decode(report_xml.py3o_template_id.py3o_template_data)
else:
    tmpl_data = self._get_template_fallback(model_instance)
    if not tmpl_data and "py3o_template_id" in model_instance._fields and model_instance.py3o_template_id:
        tmpl_data = b64decode(model_instance.py3o_template_id.py3o_template_data)

pilarvargas-tecnativa:
ping @pedrobaeza

Issue #1014, opened 2025-05-05
[16.0] bi_sql_editor - duplicate sql materialized view - cron id is the same


SupportSNDC:
Hi,

When you duplicate a V1 materialized SQL view to make a new V2 version, for example, the cron remains the same.

However, in the cron line model._refresh_materialized_view_cron([id]), the id is that of the old V1 version and so our V2 view is never refreshed.

legalsylvain:
Makes sense. The field cron_id should be copy=False. Could you make a PR ? 
Ideally upstream also. (On V18 branch)

Thanks !

PR #1017, opened 2025-05-12
[16.0][FIX] The field cron_id should be copy=False

SupportSNDC:
Hi,

When you duplicate a V1 materialized SQL view to make a new V2 version, for example, the cron remains the same.
However, in the cron line model._refresh_materialized_view_cron([id]), the id is that of the old V1 version and so our V2 view is never refreshed.

Makes sense. The field cron_id should be copy=False. Could you make a PR ?
Ideally upstream also. (On V18 branch)
Thanks !


legalsylvain:
Thanks !

Fixes : #1014 

/ocabot merge patch

Could you port this patch in latest  (V18) branch ?

Thanks !

PR #952, opened 2024-11-27
[17.0][IMP] report_qweb_pdf_watermark: support pypdf >= 2.0

hbrunn:

PR #1023, opened 2025-05-23
[18.0][IMP] report_qweb_pdf_watermark: support pypdf >= 2.0

hbrunn:
forward port of https://github.com/OCA/reporting-engine/pull/952

### DIFF
--- a/report_qweb_pdf_watermark/models/report.py
+++ b/report_qweb_pdf_watermark/models/report.py
@@ -19,10 +19,19 @@ except ImportError:
     logger.error("ImportError: The PdfImagePlugin could not be imported")
 
 try:
-    from PyPDF2 import PdfFileReader, PdfFileWriter  # pylint: disable=W0404
-    from PyPDF2.utils import PdfReadError  # pylint: disable=W0404
+    try:
+        from PyPDF2 import PdfReader, PdfWriter  # pylint: disable=W0404
+        from PyPDF2.errors import PdfReadError  # pypdf >= 2.0
+    except ImportError:
+        from PyPDF2 import (  # pylint: disable=W0404
+            PdfFileReader as PdfReader,
+        )
+        from PyPDF2 import (
+            PdfFileWriter as PdfWriter,
+        )
+        from PyPDF2.utils import PdfReadError  # pypdf < 2.0
 except ImportError:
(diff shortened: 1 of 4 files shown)


rvalyi (approved):

report_qweb_pdf_watermark/models/report.py
```diff
@@ -19,8 +19,17 @@
...
+    except ImportError:
+        from PyPDF2 import (  # pylint: disable=W0404
+            PdfFileReader as PdfReader,
+        )
+        from PyPDF2 import (
+            PdfFileWriter as PdfWriter,
+        )
+        from PyPDF2.utils import PdfReadError  # pypdf < 2.0
 except ImportError:
     logger.debug("Can not import PyPDF2")
```
StefanRijnhart: Would you consider making this a `warning`? With 2.12.1 we have everything until PdfReadError in which case setting the watermark goes fine until any exception occurs.

StefanRijnhart (commented):
Thanks!

report_qweb_pdf_watermark/models/report.py
hbrunn: happy to oblige, even though I don't think I understand what you wrote. Is there a version of the 2.x branch where the above fails? In 2.12.1 the import [should](https://github.com/py-pdf/pypdf/blob/2.12.1/PyPDF2/errors.py) work it seems to me.

StefanRijnhart (approved):

report_qweb_pdf_watermark/models/report.py
StefanRijnhart: Thanks! It's the `utils` bit, before the change from this PR.

```
>>> import PyPDF2
>>> PyPDF2.__version__
'2.12.1'
>>> from PyPDF2 import PdfFileReader, PdfFileWriter
>>> from PyPDF2.utils import PdfReadError
ModuleNotFoundError: No module named 'PyPDF2.utils'
```

hbrunn: but how do you end up there with 2.12.1, as there the first imports should work?

StefanRijnhart: That was the older version, from before this PR, and it had gone unnoticed which is why I suggested to raise the logging level.

PR #1030, opened 2025-06-12
[17.0][IMP] base_comment_template: Allow creation of global templates

etobella:
On 14, we had that comments not assigned to partners where global. After 15, it was removed.

I agree that the change makes a lot of sense, but we lost an interesting functionality and I think it can be restored by using a boolean field. WDYT?

@victoralmau @pedrobaeza

### DIFF
--- a/base_comment_template/models/comment_template.py
+++ b/base_comment_template/models/comment_template.py
@@ -41,7 +41,11 @@ class CommentTemplate(models.AbstractModel):
             templates = template_model.search(
                 expression.AND(
                     [
-                        [("id", "in", partner.base_comment_template_ids.ids)],
+                        [
+                            "|",
+                            ("id", "in", partner.base_comment_template_ids.ids),
+                            ("global_template", "=", True),
+                        ],
                         template_domain,
                     ]
                 )
--- a/base_comment_template/models/base_comment_template.py
+++ b/base_comment_template/models/base_comment_template.py
@@ -42,6 +42,7 @@ class BaseCommentTemplate(models.Model):
         help="If set, the comment template will be available only for the selected "
         "company.",
     )
+    global_template = fields.Boolean()
     partner_ids = fields.Many2many(
         comodel_name="res.partner",
         relation="base_comment_template_res_partner_rel",
--- a/base_comment_template/models/res_partner.py
+++ b/base_comment_template/models/res_partner.py
@@ -14,6 +14,7 @@ class ResPartner(models.Model):
         column1="res_partner_id",
         column2="base_comment_template_id",
         string="Comment Templates",
+        domain=[("global_template", "=", False)],
         help="Specific partner comments that can be included in reports",
     )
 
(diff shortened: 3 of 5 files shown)


pedrobaeza:
I'm not understanding the flag. You have a m2m field, so you can put there all the models you want, and the module is not designed for auto-applying comments to any arbitrary model. Or do you have something on top of it that we are missing?

etobella:
On 14 there was an option to apply templates to all records, but it was lost because you couldn't create generic non applied templates. With this change, you recover it. For example, you want a template to be applied to all moves with the product A. With this you can do it

pedrobaeza:
Sorry, but I still not understanding. One thing is the domain for records, and another for the models, and the case you are mentioning is for records, or am I wrong?

etobella:
On 14 we had this query:

https://github.com/OCA/reporting-engine/blob/14.0/base_comment_template/models/comment_template.py#L35-L40

As you can see, it applied all the templates assigned to the partner and all templates that had no partner assigned.

but on 15 it was changed for the current one.

When we split it, we obtain the templates that includes the model and are aligned with the partner. 

We are suggest to do the templates that includes the model and (are aligned with the partner or are global).

In all cases, later on, we filter by domain.

base_comment_template/models/comment_template.py
```diff
@@ -26,7 +26,10 @@ class CommentTemplate(models.AbstractModel):
         compute_sudo=True,
         comodel_name="base.comment.template",
         string="Comment Template",
-        domain=lambda self: [("model_ids", "in", self._name)],
+        domain=lambda self: [
+            ("model_ids", "in", self._name),
+            ("global_template", "=", False),
```
pedrobaeza: What confuses me is this change... You need to remove it and leave only the other.

pedrobaeza:
I don't think we need a new field. Just that if you leave `partner_ids` empty, no partner filtering is applied, and I think this is currently working, isn't it? If not, you have to make it work instead of adding the field.

base_comment_template/models/comment_template.py
etobella: the idea is to avoid selecting global templates on partners, it makes no sense.

etobella: Certainly. I made it in the wrong place. Sorry! It should be on partner, not here....

etobella:
We shouldn't remove the current logic. It works and people expects to work as it is working now.

Imagine I created a template to be used on some specific documents if needed. it has no partner, but if I apply v14 logic, everything wil change. That happened to me once. We shouldn't repeat it.

pedrobaeza:
In this point I need @victoralmau's help, as I don't follow your explanations, but it may be because I don't know in deep current logic.

victoralmau (approved):
IMO this change is coherent with the behavior of v14.

pedrobaeza:
OK

/ocabot merge patch

PR #1036, opened 2025-06-20
[17.0][FIX] bi_sql_editor: `bi_sql_editor.menu_bi_sql_editor` not found

vvrossem:


bi_sql_editor/models/bi_sql_view.py
```diff
@@ -122,12 +122,13 @@ class BiSQLView(models.Model):
...
     def _default_parent_menu_id(self):
-        return self.env.ref("bi_sql_editor.menu_bi_sql_editor")
+        return self.env.ref(
+            "bi_sql_editor.menu_bi_sql_editor", raise_if_not_found=False
+        )
 
     parent_menu_id = fields.Many2one(
         string="Parent Odoo Menu",
         comodel_name="ir.ui.menu",
-        required=True,
```
legalsylvain: Hi. please, don't remove this one, as it is mandatory.

otherwise, user can create / update sql views without parent menu field, that doesn't makes sense.

yankinmax (approved):
Good catch indeed, thanks

bi_sql_editor/views/view_bi_sql_view.xml
```diff
@@ -218,6 +218,7 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
                                         <field
                                     name="parent_menu_id"
                                     readonly="state == 'ui_valid'"
+                                    required="True"
```
legalsylvain: same here, not necessary.

legalsylvain (changes requested):
a remark inline. Otherwise, lgtm. 

thanks !

legalsylvain (approved):
/ocabot merge patch

legalsylvain:
Hi @vvrossem. could you port this patch upstream ? 

thanks

PR #1046, opened 2025-07-24
[17] FIX CI

florian-dacosta:
Tries to fix CI.
From what I understand, `report_qweb_signer` module needs endesive.
`endesive` uses `cryptography.x509.verification` which seems to be part of `cryptography` since version 42.0.0.
`endesive` does not seem to pin any version of `cryptography`

But the problem is Odoo pin an old version of cryptography (3.4.8) for python < 3.12.

So in CI we got a recent version of `endesive`, which need cryptography >= 42.0.0, but Odoo cant't have this version.

The fix I propose is to pin endesive version to last compatible version of old cryptography for python < 3.12.
Not sure there is a best way, but feel free to propose of course.


florian-dacosta:
@HviorForgeFlow @pedrobaeza @augusto-weiss related to the PR https://github.com/OCA/reporting-engine/pull/1001 you tried to merge
@legalsylvain 


Review welcome, I don't know if the proposed fix is the right way to go. I don't use `report_qweb_signer` either...

HviorForgeFlow (approved):
Makes sense
Thank you for take the time to fix the CI

pedrobaeza (approved):
/ocabot merge patch

PR #1052, opened 2025-08-05
[18.0][FIX] bi_sql_editor: replaces deprecated group_operator by aggregator

chrisb-c01:
I noticed that group operators (e.g. `avg`) were not working correctly in v18, and the values were just summed.

Haven't analysed why the `group_operator` is not used as [this deprecation code](https://github.com/odoo/odoo/blob/18.0/odoo/fields.py#L483) should be able to handle it.

Also see:
* https://github.com/odoo/odoo/pull/127353/commits/e0280ed070c7927b8b8e8933ba381199a6a1ed4c

### DIFF
--- a/bi_sql_editor/__manifest__.py
+++ b/bi_sql_editor/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "BI SQL Editor",
     "summary": "BI Views builder, based on Materialized or Normal SQL Views",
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Thanks. 
Lgtm. No test.
Fyi, No need to bump version. Bot will do it.

/ocabot merge patch

PR #1053, opened 2025-08-06
test CI

AungKoKoLin1997:


nobuQuartile:
I confirmed that asn1crypto is listed under the optional dependencies of endesive.
https://github.com/m32/endesive/blob/master/pyproject.toml#L36

It doesn’t seem to be a problem with the endesive code.
I think the issue is that the test environment isn’t installing asn1crypto properly.
However, since I’m not familiar with the CI environment, I’ve decided not to try fixing it myself.

PR #1056, opened 2025-08-14
[18.0][FIX] - Field print report name should take into account

DucTruongKomit:
Steps to Reproduce:

Choose a Excel Report Action and edit Printed Report Name field:
For example: object.name + ' - ' + object.date_order.
Print the corresponding Excel Sale Order Report from Sale Order Form view.
Observer Behavior
The file name of excel sale order gets from name of action of ir.action.report.

Expected Behavior
The file name should get from Printed Report Name field from ir.action.report.


DucTruongKomit:
it seems ok now.

PR #1057, opened 2025-08-14
[18.0][FIX] report_xlsx: Preserve custom context by using context from report instead of user context

DucTruongKomit:
Steps to Reproduce:

Try to set with_context when print the XLSX report. for example with_context(uom_id=1).
when print the report the context uom_id never show up.
**Observer Behavior**

**Expected Behavior**


DucTruongKomit:
Hello @florian-dacosta ! Could you help to review?

DucTruongKomit:
@HviorForgeFlow Please help to review this.

HviorForgeFlow (approved):

PR #1059, opened 2025-08-19
[MIG] sql_export_mail: Migration to 18.0

Larsq1:


sql_export_mail/models/sql_export.py
```diff
@@ -39,10 +39,8 @@ def _prepare_cron_mail(self):
             "model_id": self.env.ref("sql_export.model_sql_export").id,
             "state": "code",
             "code": "model._run_all_sql_export_for_cron()",
-            "name": "SQL Export : %s" % self.name,
+            "name": f"SQL Export : {self.name}",
             "nextcall": datetime.now() + timedelta(hours=2),
-            "doall": False,
```
chrisb-c01: Are these removed intentionally?

chrisb-c01 (approved):
LGTM

sql_export_mail/models/sql_export.py
Larsq1: Yes, because these fields do no longer exist in v18

Larsq1:
@legalsylvain what do you think?

sanderlienaerts (approved):

PR #1060, opened 2025-08-20
[18.0][MIG] report_qweb_signer: Migration to 18.0

chienandalu:
Refactor to use pyHanko and the new Odoo core `certificate.certificate` model.

For now, a minimal refactor has been made as PoC to:

- Use PyHanko as the signer library (more complete, mantained and reliable)
- Remove all the forme java pdfsigner stuff.
- Be able to use Odoo core certificates.

Main pain points:

- `cryptography` version is fixed in Odoo `requirements.txt`. This makes very hard for other libraries to work with that version.
- The approach has been using the `pyhanko-cli` tool as backend so we can install it in it's own `venv` (or whatever isolation method is desired).

Next steps:

- [x] Given the Odoo approach with the `certificate.certificate` model, this functionality should pivot around it.
- [x] Local certs shouldn't be in this module, but on an extension to `certificate`.
- [ ] Signing method should be in an extension to that module as well
- [x] As said in the roadmap, maybe the place to set the report signing rules should be the reports themselves.

cc @moduon MT-11131

take a look @yajo


yajo:
I see conflicts, and a module `report_qr` in the diff that I'm not sure it belongs to this PR. Could you check that please?

PR #1061, opened 2025-08-24
[FIX] *: align inherited method signature with odoo update

niyasraphy:
* report_py3o, report_xlsx, report_xml

odoo core introduced an additional argument in [1] to the method
inherited by report_py3o, report_xlsx, report_xml.
When the same method is inherited in other addons,
it raises a "TypeError: ... takes X positional arguments
but Y were given" due to a mismatch in the argument count.

This commit updates the mentioned method signature to match the
latest odoo core definition, ensuring compatibility and preventing
runtime errors when multiple addons override the same function.

No functional changes are introduced. Only method signatures have
been synchronized with upstream Odoo

1. https://github.com/odoo/odoo/commit/fc9e0160cebd5646035a3df47cb0c20edd9f55a4

### DIFF
--- a/report_py3o/controllers/main.py
+++ b/report_py3o/controllers/main.py
@@ -57,7 +57,7 @@ class ReportController(ReportControllerBase):
         return request.make_response(res, headers=http_headers)
 
     @route()
-    def report_download(self, data, context=None, token=None):
+    def report_download(self, data, context=None, token=None, readonly=True):
         """This function is used by 'qwebactionmanager.js' in order to trigger
         the download of a py3o/controller report.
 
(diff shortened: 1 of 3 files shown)


legalsylvain:
Thanks for your contribution !

Could you fix precommit ?

Thanks !

niyasraphy:
✅ Done with pre-commit, but it looks like one check is still failing.

@legalsylvain  — is this acceptable, or should I check further into it?
Would be great if you could have a look and share any insights if needed

legalsylvain (approved):
Last check is about coverage, but it ils sometime broken. (Hère for exemple)

Thanks for tour contribution.

/ocabot merge patch

PR #1063, opened 2025-08-29
[16.0][revolution] report_py3o_fusion_server: connect to libreoffice directly + improve watermarking

alexis-via:
This is a completely new implementation where Odoo connects directly to libreoffice without any intermediary (no more py3o.fusion nor py3o render server). Odoo connects to libreoffice via the UNO python lib.

The installation procedure has been updated in the README.

It's kind of wired to update the module report_py3o_fusion_server to remove support for fusion server ! But we really need a path to move away from py3o fusion and py3o render server that still use python2. And the easiest path is to update the module that is currently installed. Maybe we could rename this module in v19...

Another possibility would be to implement this directly in report_py3o...

For the moment, the priority is to check that this code works well in production. We can discuss the module structure later.
At the moment, it only works with a local libreoffice daemon. So far, I haven't found a way to make it work with a remote libreoffice daemon (but I didn't investigate that in depth).


alexis-via:
=== OUTDATED... see my new ideas in the next comment ===
Here are my latest ideas:
- implement the feature of report_py3o_fusion_server (as implemented in this PR) directly in report_py3o: if uno is installed and a port is configured and odoo is able to connect to libreoffice, it will use the code currently in report_py3o_fusion_server and pdf options work ; otherwise, it will spawn a libreoffice as currently implemented in report_py3o (in this fallback scenario, pdf options don't work).
- drop model py3o.server : we will support only a single libreoffice daemon by defaut (we can make the code inheritable for those who want to have multiple libreoffice daemons). We don't need the "host" field because we only support "localhost" (unless if we find a way to send the file to a remote libreoffice daemon)
- the port would become an ir.config_parameter

alexis-via:
I continued to investigate the different options and think about the future architecture of the modules report_py3o*.
First, it seems to me that it's not possible to send the document to convert to libreoffice via the socket used to communicate with libreoffice. So the solution implemented in this PR works fine when libreoffice is installed on the same server as Odoo, but it doesn't work when libreoffice is on another server.

So I suggest the following for the future (Odoo v19+):

- **report_py3o** would work with a local libreoffice, either running as a daemon (what is implemented in this PR) or spawned for each document conversion (current implementation in report_py3o). The object py3o.pdf.options would be moved from report_py3o_fusion_server to report_py3o ; it would be used when libreoffice is running as a daemon, but not used when libreoffice is spawned for each conversion. For the connection to libreoffice, host would be hard-coded to "localhost" and port would be configured via an ir.config_parameter.
- **report_py3o_unoserver** would work when libreoffice is installed on another server.  It would require unoserver running on the server where libreoffice is installed. Unoserver is developped under MIT licence on https://github.com/unoconv/ and available on pypi. It would be a kind of replacement of py3o-fusion and py3o-renderserver. The object py3o.server would be moved from report_py3o_fusion_server to report_py3o_unoserver.
- report_py3o_fusion_server would not be migrated to newer versions any more.

I always run libreoffice locally on my Odoo servers, so I don't need report_py3o_unoserver myself, but I'm ready to help on the development of report_py3o_unoserver if some community members are interested by this solution. I would be ready to work on report_py3o and report_py3o_unoserver during the next code sprint in Liège on September 15-17th 2025.

alexis-via:
@JasminSForgeFlow @JordiMForgeFlow As the author of the v18 migration of report_py3o_fusion_server, your opinion on my proposal to restructure report_py3o would be very welcomed.

JordiMForgeFlow:
@alexis-via we have honestly worked very little with the py3o modules, but the cases we have seen have also been with libreoffice running locally. From what you comment I think the roadmap makes sense, but I am not sure about the effects for installations running it on another server.

alexis-via:
The new options for watermarking:

watermark

when repeat is enabled:

watermark_repeat

PR #1064, opened 2025-09-01
[18.0] [ADD] report_latex

len-foss:
Provides Latex reports for Odoo. Templating is provided by Jinja2.

Features:
- Syntax highlighting for Latex in the Ace code editor.
- Support for \input{...} syntax; split long documents into multiple files.
- Direct editing of the generated Latex source.

One of the most interesting aspect is that it gives the compiled Latex (from Jinja) that can be tweaked as desired and recompiled. I'm not sure of the best way to have this work in multi-record mode -- sources could be generated without a res_id, but the UX does not sound great, so I haven't allowed it in this version. 

The demo provides an example report showing how to configure a report; to have \input{} work, you need to declare the subtemplates used by the root template, and declare the attachments to embed. It does not support dynamic embed (yet, at least). The only thing is that it does not support subfolders, so the split have to be flat. As a long-time LaTeX user it does not seem to be a horrible restriction.

Perhaps the most annoying thing is that compilation just says that it fails. There's a way to get the report assets and modify them locally to ease development. Splitting files also help.

No options yet for compilation. It would be easy to change pdflatex to xelatex. Less easy to delegate compilation to a remote server, but it could be an option.

Last point, I'm not sure how it would go for testing (besides putting the compilation step in skip?). Latex packages are very heavy so I'm not sure it would be reasonable (but maybe docker images are more efficient than I think) -- and it could be restricted to texlive-base 

Screenshot_20250901_210652
Screenshot_20250901_210721
Screenshot_20250901_210747


rvalyi:
lol the 1st Akretion quotes 16 years ago were done using OpenERP and... yes Latex! cc @sebastienbeau 
Cool to see people are still willing to use Latex today.

len-foss:
@rvalyi 
That's interesting. To be honest I wasn't expecting it to have much interest in the community.
I've thought about using Latex since I've used Odoo -- and had to deal with wkhtmltopdf problems. 
Now this buggy project has been abandonned for years, and Odoo still depends on it... and their replacement is a blue-sky project. Oh well.

Let me know if you test it out :-)

len-foss:
Note: this PR won't be up-to-date with latest developments; I can push latest changes on demand. 

There's a bunch of additional features required to make the integration complete, in particular if a template might be used in the portal,  it should backup to an 'equivalent' qweb version. 

Also, on the latest Debian-based distros the version of Pandoc shipped is woefully outdated, meaning that Latex support is incomplete -- meaning not working in our context... Such systems should manually install Pandoc from the latest Github release, which is really simple.

PR #1072, opened 2025-10-05
[19.0][MIG] report_xml

etobella:
Migration to 19.0.

Added some code on the tests in order to load demo data.


pedrobaeza:
/ocabot migration report_xml

Please remind me one more time what this layer provides against putting a QWeb template with the XML header and content like Odoo does for Factur-X and others (example: https://github.com/odoo/odoo/blob/19.0/addons/account_edi_ubl_cii/data/cii_22_templates.xml)

etobella:
It allows to generate a report that delivers an XML file directly. Otherwise, Odoo will send you back an HTML (with xml format, but an HTML) or PDF.

pedrobaeza:
Let's include that in CONTEXT.md for later references.

pedrobaeza (commented):
Don't put `lxml` as `external_dependencies`. It's already a dependency of Odoo core.

pedrobaeza (approved):

luisDIXMIT (commented):
Should be "Demo xml report" listed? It doesn't appears.

image

etobella:
with the new default value of demo, demo is not imported by default.

luisDIXMIT (approved):
LGTM!

PR #1073, opened 2025-10-06
[18.0][revolution] report_py3o_fusion_server: connect to libreoffice directly + improve watermarking

alexis-via:
This is a completely new implementation where Odoo connects directly to libreoffice without any intermediary (no more py3o.fusion nor py3o render server). Odoo connects to libreoffice via the UNO python lib.

The installation procedure has been updated in the README.

This PR is the same as PR #1063  but for 18.0 instead of 16.0.

This PR is not designed to be merged ; I publish it here so that it can be used by those who would need it. I think the big change should happen in 19.0.


alexis-via:
The new options for watermarking:

watermark

an example of the "repeat" option:

watermark_repeat

PR #1078, opened 2025-10-13
[19.0][MIG] report_xlsx

tupaq:
Migration to 19.0.

In report_xlsx/tests/test_report.py the test_report method is removed because xlrd cannot read xlsx files [xlrd](https://github.com/python-excel/xlrd/commit/56c0bd76c8c8db5d7adb5f5210765c21a2ab1944#diff-b5d3d00cff0b461900e3b491cc6ccd3f626c36382b5d02c7b887c7c63147f118R156)


pedrobaeza:
/ocabot migration report_xlsx

Several things to attend:

- You have remove the test, but it's essential to test that this is working, so you have to find an alternative way to check it.
- What you call IMP should be part of the migration commit, as this is something that happens with the change of the version, and don't do search, but directly create a new report in the tests independently form the demo data.

tupaq:
> /ocabot migration report_xlsx
> 
> Several things to attend:
> 
>     * You have remove the test, but it's essential to test that this is working, so you have to find an alternative way to check it.
> 
>     * What you call IMP should be part of the migration commit, as this is something that happens with the change of the version, and don't do search, but directly create a new report in the tests independently form the demo data.

Restored the essential test for report_xlsx.partner_xlsx using openpyxl and added test-requirements.txt. This ensures xlsx report generation is properly tested.

I didn't include it in the commit because we expect the test server to create the database with demo data. This means that installing the add-on doesn't install the test report:

`"demo": ["demo/report.xml"],` in `__manifest__.py` file
[First commit. Errors](https://github.com/OCA/reporting-engine/actions/runs/18453094715/job/52569591686) 

This didn't happen before.

That's why, to test the module, we check with a search to see if the report is installed, and if not, we create it.

Sorry for the spelling. English is not my native language.

@pedrobaeza

report_xlsx/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2015 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Base report xlsx",
+    "summary": "Base module to create xlsx report",
+    "author": "ACSONE SA/NV,Creu Blanca,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/reporting-engine",
+    "category": "Reporting",
+    "version": "19.0.1.0.1",
```
pedrobaeza:
```suggestion
    "version": "19.0.1.0.0",
```

report_xlsx/tests/test_report.py
```diff
@@ -1,22 +1,39 @@
...
 except ImportError:
-    _logger.debug("Can not import xlrd`.")
+    _logger.debug("Can not import openpyxl`.")
 
 
 class TestReport(common.TransactionCase):
     def setUp(self):
         super().setUp()
         self.report_object = self.env["ir.actions.report"]
+        report_id = self.report_object.search(
```
pedrobaeza: I insist: please remove the search and directly work with one report created on the fly here (that don't overlap/crashes if demo data are also present).

tupaq: Remove the search and check for the existence of the external ID. I think it's cleaner and has no impact.
`if not self.env.ref("report_xlsx.partner_xlsx", False):`

pedrobaeza: Please don't use demo data. Go directly with the data created in the test.

pedrobaeza (approved):
Thank you very much.

rov-adhoc (approved):

PR #1079, opened 2025-10-17
[17.0][IMP]bi_sql_editor: many2one fields clickable on the list view if setting is enabled

AndreuOForgeFlow:
(cherry picked from commit 53d0a053f6e58142f6efefcc013f34796e328772)


bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -86,6 +86,12 @@ class BiSQLViewField(models.Model):
 
     index_name = fields.Char(compute="_compute_index_name")
 
+    is_many2one_clickable = fields.Boolean(
+        default=True,
+        help="If the field is of type Many2one and the option is enabled, the"
+        "field will be clickable on the list view.",
```
GuillemCForgeFlow:
```suggestion
        " field will be clickable on the list view.",
```
I can see a space is missing

bi_sql_editor/views/view_bi_sql_view.xml
```diff
@@ -127,6 +127,10 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
                                 required="ttype == 'many2one'"
                                 readonly="state in ('model_valid', 'ui_valid')"
                             />
+                            <field
+                                name="is_many2one_clickable"
+                                invisible="not field_description or ttype != 'many2one'"
+                            />
```
GuillemCForgeFlow, line 134:
```suggestion
                            <field
                                name="is_many2one_clickable"
                                invisible="not field_description or ttype != 'many2one'"
                                optional="show"
                            />
```
We could enable the possibility to hide in the UI 👀

bi_sql_editor/models/bi_sql_view_field.py
```diff
@@ -86,6 +86,12 @@ class BiSQLViewField(models.Model):
 
     index_name = fields.Char(compute="_compute_index_name")
 
+    is_many2one_clickable = fields.Boolean(
+        default=True,
```
GuillemCForgeFlow:
```suggestion
        default=False,
```
Perhaps we can set it as `False` by default, so there are no changes in current installations.

GuillemCForgeFlow:
@legalsylvain I know we discussed this back in the day in https://github.com/OCA/reporting-engine/pull/892. But I was thinking, with the changes now proposed, we should be able to include it in the module as a minor improvement.
WDYT?

bi_sql_editor/models/bi_sql_view_field.py
legalsylvain: In the meantime, I think that most of the users will want this feature.
In fact, I don't understand why put a setting here. (I don't understand when we want to put False here).
Do you have any use case ?

legalsylvain (commented):
Thanks for porting this feature !

bi_sql_editor/models/bi_sql_view_field.py
AndreuOForgeFlow: @legalsylvain, There are cases where we don’t want all `many2one` fields to be clickable, so users should be able to choose which ones are navigable on the SQL view.

GuillemCForgeFlow (approved):
lgtm 👍🏿

JordiMForgeFlow (approved):
code + functional 👍🏼

GuillemCForgeFlow:
@legalsylvain can we get this one merged by any chance? 🙂

PR #1081, opened 2025-10-21
[18.0][FIX] report_xlsx: report xlsx/xlrd 2.0 compatibility

vincent-hatakeyama:
- Removed the module requirements, they are already required libraries of odoo
- Adapt the tests to avoid failing when xlrd is at version 2.0 (the library does not handle xlsx): use openxyl by default and fall back on using xlrd if the version is compatible.

I’ve added a skip of the tests if both library are absent; not sure it should stay.

### DIFF
--- a/report_xlsx/__manifest__.py
+++ b/report_xlsx/__manifest__.py
@@ -9,7 +9,6 @@
     "version": "18.0.1.0.1",
     "development_status": "Mature",
     "license": "AGPL-3",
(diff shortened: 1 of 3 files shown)


vincent-hatakeyama:
In #1078, the use of xlrd has been completely removed in favour of openpyxl. The same could be done here.

HviorForgeFlow (approved):

lmignon (approved):

PR #1082, opened 2025-10-24
[18.0][MIG] report_pdf_zip_download: Migration to 18.0

AungKoKoLin1997:
@qrtl QT5817

PR #1084, opened 2025-10-29
[FIX] report_pdf_zip_download: py35 compatible code

HviorForgeFlow:
f-string not compatible with python3.5


yostashiro (approved):
Looks legit. Thanks for the fix!👍

AungKoKoLin1997 (approved):
Thanks for the fix.
LGTM

yostashiro:
@HviorForgeFlow Would be great if you could review https://github.com/OCA/reporting-engine/pull/1082. ;)

PR #1086, opened 2025-11-03
[17.0][IMP] bi_sql_editor: make field_description field translate as True

AndreuOForgeFlow:
That way SQL view columns are translatable just like any other view


legalsylvain:
makes sense.

/ocabot merge patch

PR #1088, opened 2025-11-11
[19.0][FIX] report_xlsx: Preserve custom context by using context from report instead of user context

DucTruongKomit:
Steps to Reproduce:

Try to set with_context when print the XLSX report. for example with_context(uom_id=1).
when print the report the context uom_id never show up.
Observer Behavior

Expected Behavior

### DIFF
--- a/report_xlsx/static/src/js/report/action_manager_report.esm.js
+++ b/report_xlsx/static/src/js/report/action_manager_report.esm.js
@@ -8,8 +8,14 @@ registry
         if (action.report_type === "xlsx") {
             const type = action.report_type;
             let url = `/report/${type}/${action.report_name}`;
+            for (const key_data in action.data) {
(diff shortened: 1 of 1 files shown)


DucTruongKomit:
@HviorForgeFlow Hello! Could you help to review this PR, please?

the fix is already existing in v17, but not in 18 an 19.
https://github.com/OCA/reporting-engine/pull/933

HviorForgeFlow (approved):

HviorForgeFlow:
Thanks @DucTruongKomit for moving this PR to this newer version, do you mind add also the fix for the branch 18.0? I will review it asap.

DucTruongKomit:
@HviorForgeFlow 
https://github.com/OCA/reporting-engine/pull/1057

I created already here.

PR #1090, opened 2025-11-12
[18.0][MIG] report_footer_html: Migration to 18.0

u0f:
MIgration to 18:
- Support new Odoo native documents layout
- Update translations 
- Update README

https://www.loom.com/share/7b1a82749aff4adc8c4776c67992f6e0

MT-12187 @moduon @Shide


report_footer_html/models/ir_actions_report.py
```diff
@@ -8,7 +8,7 @@ class IrActionsReport(models.Model):
     _inherit = "ir.actions.report"
 
     footer_html = fields.Html(
-        string="Footer HTML",
+        string="Footer",
```
chienandalu: I guess this change is due to the warning in the logs stating that the label is the same that the string. But you just need to remove the string parameter and so you don't have tu update your terms

chienandalu (approved):
Small comment (other than that seems ok)

report_footer_html/models/ir_actions_report.py
u0f: This change was made to match the string that Odoo uses natively, as @Shide proposed to me.

image

chienandalu: What about changing it to "Extra footer" as it describes better how it really behaves?

u0f: I spoke with @Shide and we thought "Extra Footer" is a good change, thanks! Just added.

chienandalu (approved):

fcvalgar (commented):
Thanks for your contribution and the work on this module @u0f . I’ve run an initial round of functional tests in Odoo to validate the behaviour, and most scenarios are working as expected. I’ve found one issue related to image rendering in the HTML footer that might require your review.

Below is the summary of the tests performed:

Test 1: Basic HTML in quotation footer – OK
Test 2: Complex HTML (text + link + simulated logo) in Standard layout – No OK  (the link is not clickable and the logo is not displayed)

image

Test 3: Coexistence with native footer in Striped/Wave layouts – OK
Test 4: Report without Extra Footer content – OK
Test 5: Bottom margin adjustment in Paper Format – OK
Test 6: Footer rendering in qweb-html preview – OK
Test 7: Different HTML footers per report type – OK
Test 8: Footer with image in Extra Footer – Not OK (the image is not rendered in PDF)

image
image

Could you please take a look at this case?

Thank you @u0f

u0f:
Thanks for your review, @fcvalgar.

I’ve been testing adding images to the Extra Footer and didn’t run into any issues — the image is correctly rendered in every Document Layout.

Tested on RunBoat

Regarding the links issue, I’ve been looking into it. This is not actually a bug but an Odoo limitation. In the footer, you can’t add hyperlinked text through the editor, but you can paste link as plain test and it will work.
I’ve added this information to the ROADMAP.md`.

image

fcvalgar (approved):
@u0f Thank you for the detailed explanation. The type of image I was using was not correct. 

LGTM

HviorForgeFlow (approved):
code review, appreciated implementation details :+1:

PR #1095, opened 2025-11-18
[18.0][FIX] report_xlsx 

alvaroGC-S73:
PR to Revert: #1057

Error to solve:

RPC_ERROR

Odoo Server Error

  File "/opt/odoo/src/oca/reporting-engine/report_xlsx/controllers/main.py", line 65, in report_download
    response = self.report_routes(

..........

  File "/opt/odoo/src/oca/account-financial-reporting/account_financial_report/report/abstract_report.py", line 167, in _get_report_values
    wizard = self.env[data["wizard_name"]].browse(data["wizard_id"])
KeyError: 'wizard_name'

        
        
The core issue is that the new JavaScript code introduced a bug: while attempting to fix context propagation for XLSX reports, it broke the standard method for passing wizard data (wizard_name, wizard_id). The old JS correctly serialized the essential action.data into the URL as an option query parameter, which the OCA backend modules rely on to find the wizard instance and generate the report.

### DIFF
--- a/report_xlsx/static/src/js/report/action_manager_report.esm.js
+++ b/report_xlsx/static/src/js/report/action_manager_report.esm.js
@@ -8,14 +8,8 @@ registry
         if (action.report_type === "xlsx") {
             const type = action.report_type;
             let url = `/report/${type}/${action.report_name}`;
-            for (const key_data in action.data) {
-                action.context[key_data] = action.data[key_data];
-            }
             const actionContext = action.context || {};
             if (action.data && JSON.stringify(action.data) !== "{}") {
-                if (!url.includes(`/${actionContext.active_id}`)) {
-                    url += `/${actionContext.active_id}`;
-                }
                 // Build a query string with `action.data` (it's the place where reports
                 // using a wizard to customize the output traditionally put their options)
                 const action_options = encodeURIComponent(JSON.stringify(action.data));
@@ -36,7 +30,7 @@ registry
                     url: "/report/download",
                     data: {
                         data: JSON.stringify([url, action.report_type]),
-                        context: JSON.stringify(actionContext),
+                        context: JSON.stringify(user.context),
                     },
                 });
             } finally {


micheledic (approved):
LGTM.
This bug blocks this PR https://github.com/OCA/account-financial-reporting/pull/1413

pedrobaeza:
But reverting this will undo the feature. This should instead fix the root cause, isn't it?

micheledic:
> But reverting this will undo the feature. This should instead fix the root cause, isn't it?

I tried this PR and the xls report works , imho can be merged to fix the issue

pedrobaeza:
Yeah, but the reason why it was put initially will come to not be resolved, so that's why I say that this should fix the problem without losing the feature.

micheledic:
@alvaroGC-S73 i just checked and it may be enough to remove those line and the xlsx report works
https://github.com/OCA/reporting-engine/blob/18.0/report_xlsx/static/src/js/report/action_manager_report.esm.js#L16-18
https://github.com/OCA/reporting-engine/blob/18.0/report_xlsx/static/src/js/report/action_manager_report.esm.js#L25-27

Or we have to push a PR to odoo core because the issue is caused on this line:
https://github.com/odoo/odoo/blob/70404624d7ea6f971182cbe4f4fe8fc064d2afd8/addons/web/controllers/report.py#L118

Of docids is passed, the date is not passed correctly

JordiBForgeFlow:
In my opinion this PR should be merged ASAP because it fixes a breaking error. The feature that was introduced in https://github.com/OCA/reporting-engine/commit/d931ea74a6ed37bab5f67362cd4b66f195baba81 was not fixing an error

JordiBForgeFlow:
> @alvaroGC-S73 i just checked and it may be enough to remove those line and the xlsx report works https://github.com/OCA/reporting-engine/blob/18.0/report_xlsx/static/src/js/report/action_manager_report.esm.js#L16-18 https://github.com/OCA/reporting-engine/blob/18.0/report_xlsx/static/src/js/report/action_manager_report.esm.js#L25-27
> 
> Or we have to push a PR to odoo core because the issue is caused on this line: https://github.com/odoo/odoo/blob/70404624d7ea6f971182cbe4f4fe8fc064d2afd8/addons/web/controllers/report.py#L118
> 
> Of docids is passed, the date is not passed correctly

I agree that the fix proposed here works, at least when we try to export the financial reports. But not sure if it breaks something else.

pedrobaeza:
OK, let's revert it and let @DucTruongKomit to propose an alternative solution to his problem.

/ocabot merge patch

micheledic:
I created fwport https://github.com/OCA/reporting-engine/pull/1099 @pedrobaeza

PR #1103, opened 2025-12-08
[14.0][ADD] report_qweb_text_control: Controls whitespace and special characters in text QWeb reports

dreispt:
Allow text reports to generate positional text files by providing 
precise control over whitespace, special characters, and field formatting.

Use Case
--------

Generate fixed-width text files for legacy systems, bank transfers, or any interface 
requiring exact positional formatting with specific characters and spacing.

Example
-------

Template::

    <span t-esc="A(partner.name, 20)" />
    <span t-esc="N(partner.id, 10)" />
    <span t-esc="M(partner.credit_limit, 12)" /> 

Output::

    John Doe           00000012340000015000+


dreispt:
NOTE: 
I had a bug in a fixed positional file export, because t-esc expands "&" into "&amp;", making the line 4 characters longer than expected.
The workaround was to use t-raw instead of t-esc, but t-raw is deprecated in more recent versions.
So an alternative solution will be needed.

PR #1107, opened 2025-12-18
[19.0][MIG] report_pdf_form

bosd:
Standard migration.

"Improvements" deliberatly made in seperate commits for cherry-picking


report_pdf_form/README.rst
```diff
@@ -0,0 +1,115 @@
...
+-  **docfly.com**: An online tool for adding form fields to PDFs
+-  **LibreOffice Draw**: Part of the LibreOffice suite, can add form
+   fields to PDFs
+-  **Scribus**: A desktop publishing application that can create PDFs
+   with form fields
+
+.. IMPORTANT::
+   This is an alpha version, the data model and design can change at any time without warning.
+   Only for development or testing purpose, do not use in production.
+   `More details on development status <https://odoo-community.org/page/development-status>`_
```
legalsylvain, line 56: Still Alpha ?

bosd: I think so.. This is the first migration. And have been adding some features.
Ci is :red_circle:  :crying_cat_face: 

Maybe later promote it to beta? Will check the requirements..

bosd: Promoted to **Beta** now. I also added test coverage for the features that were bundled in this migration (field-type formatting, dotted-path validation + constraint, the unique `report_id` guard, code variables and the PDF preview), and fixed a bug in the `write()` uniqueness check found while doing so. CI should be green now.

PR #1113, opened 2025-12-23
[18.0][MIG] report_partner_address: Migration to 18.0

AungKoKoLin1997:
This module adds a translatable Address Details field on partners for use in reports and portal.

Odoo prints documents in the partner's language but lacks a standard way
to display addresses in multiple languages. This module fills that gap.

@qrtl QT5088


report_partner_address/models/res_partner.py
```diff
@@ -0,0 +1,57 @@
...
+                """
+                UPDATE res_partner
+                SET address_details = p.address_details
+                FROM res_partner p
+                WHERE res_partner.id = %(partner_id)s
+                AND p.id = %(parent_id)s
+                """,
+                {"partner_id": self.id, "parent_id": parent.id},
+            )
+        return result
```
yostashiro: I realize that this sync is problematic as contacts' names won't display in reports and portal views when they have to, and therefore all this should be dropped (unfortunately...).

report_partner_address/tests/test_report_partner_address.py
```diff
@@ -55,27 +55,24 @@ def test_address_details_empty_uses_standard_format(self):
         self.assertIn("123 Main St", res)
         self.assertIn("Springfield", res)
 
-    def test_contact_inherits_address_details_from_parent(self):
+    def test_contact_does_not_inherit_address_details(self):
```
yostashiro: I think this test can be removed. There is no point in proving non-existence of functionality we didn't implement.

report_partner_address/readme/USAGE.md
```diff
@@ -0,0 +1,8 @@
+- Go to the partner record.
+- Fill in Address Details, including translations for the active languages.
+
+The address in reports and portal pages will use the value from Address Details.
+
+If Address Details is left blank, the standard address format is used.
+
+Note: Address Details does not sync to child partners like Odoo’s standard address fields.
```
yostashiro: To be slightly more precise.
```suggestion
Note: Unlike standard address fields, Address Details is not inherited by child contacts.
```

yostashiro (approved):
Code review. LGTM.

smorita7749 (approved):
Fill out Address Details for both JA and EN and set Japanese in Language.
Then, Address Details in Japanese are displayed in report and portal.

nobuQuartile (approved):
Functional review: LGTM

HviorForgeFlow (approved):

PR #1117, opened 2026-01-09
[18.0][FIX] report_csv: add None check for docids in _render_csv

deeniiz:
It is being assumed that the parameter docids is always iterable, and len(docids) is called without validation, which therefore raises an error.

### DIFF
--- a/report_csv/README.rst
+++ b/report_csv/README.rst
@@ -128,6 +128,10 @@ Contributors
 
   - Aung Ko Ko Lin
 
+- `NuoBiT <https://www.nuobit.com>`__:
+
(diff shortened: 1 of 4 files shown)


deeniiz:
@eantones , can you please review it?

eantones (approved):
LGTM!

sergijPf:
@deeniiz There is a similar fix for v16.0 (+related test) and it was already merged https://github.com/OCA/reporting-engine/pull/962,
you can replace your commit with that one, probably, it'll go faster then.

HviorForgeFlow (approved):

PR #1123, opened 2026-01-26
[18.0][MIG] report_layout_config: Migration to 18.0 

weinni2000:
standard migration

PR #1126, opened 2026-02-03
[18.0][FIX] report_csv: prevent crash when docids is None in _render_csv

### LINKED ISSUE #1125: [BUG][18.0] report_csv: _render_csv fails when docids is None
## Module
```report_csv```

## Describe the bug
Rendering a CSV report fails with a `TypeError` when `docids` is `None`.
The _render_csv method assumes `docids` is always iterable and calls `len(docids)` without checking its value.

## To Reproduce
**Affected versions**:

1. Trigger a CSV report rendering without passing `docids`
2. `_render_csv` is called with `docids=None`
3. The method executes `len(docids)`
4. A `TypeError` is raised

**Expected behavior**
CSV reports should render correctly even when `docids` is `None`, consistent with other Odoo report renderers.

**Actual behavior**
A `TypeError` is raised and the report rendering fails.

**Additional context**
Error:
```python
TypeError: object of type 'NoneType' has no len()
```
Code reference:
https://github.com/OCA/reporting-engine/blob/6d3b9f0bcaca525044fcf336f229350a1c0d8eef/report_csv/models/ir_report.py#L58

---
Environment:
OS: Linux
Python: 3.x
Odoo: 18.0

LoayMoh99:
Fixes #1125 

Add a defensive check in `_render_csv` to avoid calling `len()` on `None`
when `docids` is not provided.

This aligns CSV report rendering with other Odoo report renderers
and prevents a `TypeError`.

### DIFF
--- a/report_csv/models/ir_report.py
+++ b/report_csv/models/ir_report.py
@@ -55,7 +55,7 @@ class ReportAction(models.Model):
         report_sudo = self._get_report(report_ref)
         report_model_name = f"report.{report_sudo.report_name}"
         report_model = self.env[report_model_name]
-        res_id = len(docids) == 1 and docids[0]
+        res_id = docids and len(docids) == 1 and docids[0]
         if not res_id or not report_sudo.attachment or not report_sudo.attachment_use:
             return report_model.with_context(
                 **{


flotho (approved):
Code review,
LGTM
Thanks

HviorForgeFlow:
fixed in https://github.com/OCA/reporting-engine/pull/1123

HviorForgeFlow (approved):

PR #1127, opened 2026-02-03
[19.0][IMP] bi_sql_view: add view existence check before dropping and refre…

lef-adhoc:
…shing views


legalsylvain (changes requested):
I dont understand in which case the view is missing or not in the correct type. Could you elaborate the scénario ? Thanks !

PR #1128, opened 2026-02-08
[19.0][MIG] sql export: Migration to 19.0

feketemihai:


sql_export/wizard/wizard_file.py
```diff
@@ -0,0 +1,95 @@
...
+        """,
+            {
+                "last_date": fields.Datetime.to_string(fields.Datetime.now()),
+                "last_uid": self.env.user.id,
+                "id": sql_export.id,
+            },
+        )
+        url = f"""web/content/?model={self._name}&id={self.id}&
+            filename_field=file_name&"field=binary_file&download=true&
+            filename={self.file_name}"""
```
dannyadair: I think this URL got jumbled in migration.

Suggested fix:
```
url = (
    f"web/content/?model={self._name}&id={self.id}"
    f"&filename_field=file_name&field=binary_file&download=true"
    f"&filename={self.file_name}"
)
```

So "filename" is now "file_name" other than that I think the above is the correct f-string version of
https://github.com/OCA/reporting-engine/blob/18.0/sql_export/wizard/wizard_file.py#L88

P.S.: I think `filename_field=filename` in v18 (where the wizard field name was alreadt `file_name`) was wrong but without consequence because `filename` was given - see `_get_stream_from()` in `odoo/addons/base/models/ir_binary.py`.

dannyadair (changes requested):

dannyadair (approved):

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@tagged("post_install", "-at_install")
+class TestExportSqlQuery(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.sql_export_obj = cls.env["sql.export"]
+        cls.wizard_obj = cls.env["sql.file.wizard"]
+        cls.sql_export_partner = cls.sql_export_obj.create(
```
lmignon, line 18: @feketemihai The setup could be simplified by explicitely loading  the demo file....

```python
from odoo.tools import convert


@classmethod
def setUpClass():
    super().setUpClass()
    ....
    convert.convert_file(env=cls.env, module='sql_export', filename='demo/sql_export.xml', idref=None, mode='init')
    cls.sql_export_partner = self.env.ref("sql_export.sql_export_partner")
    ......

```

lmignon (changes requested):
Thank you for the migration. To minimize the diff in the migration process you should load the demo data into the setUpClass method of your tests

sql_export/tests/test_sql_query.py
feketemihai: Hi, i updated them.

feketemihai:
@lmignon I've updated some tiem ago the tests. Thanks.

sebalix:
@feketemihai can you squash all your fixes into the migration commit?

PR #1130, opened 2026-02-09
[19.0][MIG] bi_sql_editor: Prevent failure if default parent menu is not loaded.

baptiste-n42:
Issue occurs when upgrading from an old version of this module.

`ValueError: External ID not found in the system: bi_sql_editor.menu_bi_sql_editor`

### DIFF
--- a/bi_sql_editor/README.rst
+++ b/bi_sql_editor/README.rst
@@ -182,6 +182,8 @@ Contributors
 
 - Thien Vo thienvh@trobz.com
 
+- Baptiste P. baptiste@newlogic.com
+
 - This module is highly inspired by the work of
 
   - Onestein: (http://www.onestein.nl/) Module:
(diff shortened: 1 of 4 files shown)


legalsylvain (approved):
> Issue occurs when upgrading from an old version of this module.

I don't get it. upgrading should load new xml data. How it can occures?  could you elaborate.

(However, this patch doesn't hurt, so approving)

baptiste-n42:
> I don't get it. upgrading should load new xml data. How it can occures? could you elaborate.

It does, however it seems that Odoo is reading first the column definition (thus running the default lambda function) before checking for changes in the views. 
It's like: 
1. Odoo delete the views/menus from v15 (initial version)
2. makes the ORM upgrade (this is where it crashes)
3. load the new views.

legalsylvain:
> Odoo delete the views/menus from v15 (initial version)

How. What a weird upgrade process. I don't have such problem with OCA / OpenUpgrade tools.

So indeed, this patch is needed in your case.

/ocabot merge patch

PR #1131, opened 2026-02-11
[19.0][MIG] report_csv: Migration to 19.0

sergijPf:
standard migration,
few changes done:
 - minor code changes as per tasks To-Do https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#tasks-to-do-in-the-migration;
 - fixed tests (demo-report creation moved directly to tests), because of no demo data while tests run;
 - considered fix done for version 16.0 - https://github.com/OCA/reporting-engine/pull/962


report_csv/models/ir_report.py
```diff
@@ -0,0 +1,93 @@
...
+        return data, ext
+
+    @api.model
+    def _get_report_from_name(self, report_name):
+        res = super()._get_report_from_name(report_name)
+        if res:
+            return res
+        report_obj = self.env["ir.actions.report"]
+        qwebtypes = ["csv"]
+        conditions = [
```
ddejong-therp, line 88: You can use the new domain class here.
See https://www.odoo.com/documentation/19.0/developer/reference/backend/orm.html?highlight=orm#reference-orm-domains

ddejong-therp:
I have tested it and the code does work, but I left a small review point.

sergijPf:
> I have tested it and the code does work, but I left a small review point.

@ddejong-therp did you submit your review? there are no review comments.

ddejong-therp:
> > I have tested it and the code does work, but I left a small review point.
> 
> @ddejong-therp did you submit your review? there are no review comments.

It looks like I did not submit the review indeed!

report_csv/models/ir_report.py
sergijPf: @ddejong-therp thanks for noticing, yes, it might work with newly added 'Domain', but it still works as before + it looks more like core method which makes it easier to read as for me
image

ntsirintanis (approved):

PR #1138, opened 2026-02-26
[18.0][ADD] report_positioned_image

kanda999:
@qrtl QT6451


report_positioned_image/models/ir_actions_report.py
```diff
@@ -0,0 +1,161 @@
...
+<script>
+var page = window.location.search.match(/[?&]page=(\\d+)/);
+if (page && parseInt(page[1]) > 1) {{
+  var elements = document.getElementsByClassName('{css_class}');
+  for (var i = 0; i < elements.length; i++) {{
+    elements[i].style.display = 'none';
+  }}
+}}
+</script>
+"""
```
kanda999: Is it unavoidable to use JS?

AungKoKoLin1997:
Tests CI is fixed at https://github.com/OCA/reporting-engine/pull/1142.

report_positioned_image/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Report Positioned Image",
+    "summary": "Add positioned images to PDF reports generated by Odoo.",
```
yostashiro:
```suggestion
    "summary": "Add positioned images to PDF reports",
```

report_positioned_image/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2026 Quartile
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
```
yostashiro:
```suggestion
```

report_positioned_image/models/res_company.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    report_positioned_image_ids = fields.Many2many(
```
yostashiro, line 10: I think this should be one2many. It looks inconsistent with how company_id is defined as many2one in report.positioned.image.

report_positioned_image/models/ir_actions_report.py
```diff
@@ -0,0 +1,158 @@
...
+  }}
+}}
+</style>
+"""
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    include_company_report_image = fields.Boolean(
```
yostashiro:
```suggestion
    include_company_images = fields.Boolean(
```

yostashiro (changes requested):
Functional test. 'First Page Only' option doesn't seem to be working.

Image

Image

report_positioned_image/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_report_positioned_image_portal,report.positioned.image portal,model_report_positioned_image,base.group_portal,1,0,0,0
```
yostashiro: This one doesn't look necessary. Please confirm.
```suggestion
access_report_positioned_image_portal,report.positioned.image 
```

AungKoKoLin1997:
> 'First Page Only' option doesn't seem to be working.

@yostashiro For some reason, my latest code changes are not reflected in this PR.

In my update, I added `report_qweb_element_page_visibility` as a dependency to improve the usability of the first-page feature.

report_positioned_image/__manifest__.py
AungKoKoLin1997: Updated!

report_positioned_image/models/__init__.py
AungKoKoLin1997: Updated!

report_positioned_image/models/res_company.py
AungKoKoLin1997: I intentionally avoid using a One2many field because it can lead to unintended behavior.

For example, when we add a positioned image to an Order/Quotation report for a specific company, that image is also assigned to the company’s report images. As a result, every time we create a positioned image for a specific report, all of those images are added to the company as well.

Then, if we remove the report images from the company, they are also unlinked from the specific report, which is not the intended behavior.

report_positioned_image/models/ir_actions_report.py
AungKoKoLin1997: I wanted to include report to make the purpose clearer, but I think `include_company_images` is already enough. So I updated it.

report_positioned_image/security/ir.model.access.csv
AungKoKoLin1997: I initially thought that portal users would need access to the model.
After verifying the behavior, I confirmed that this is not necessary because `_show_report` in the portal runs with `sudo()`.

report_positioned_image/models/res_company.py
yostashiro: Okay, we may add a help text to company_id of the image model to make the intent clear.

report_positioned_image/models/report_positioned_image.py
```diff
@@ -0,0 +1,96 @@
...
+    height = fields.Float(string="Height (mm)")
+    respect_image_ratio = fields.Boolean(
+        default=True,
+        help="When enabled, changing width or height will automatically adjust "
+        "the other dimension to maintain the original image aspect ratio.",
+    )
+    first_page_only = fields.Boolean()
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        required=True,
```
yostashiro: Can we remove this `required=True`? I think it will allow more flexible use of the images. We should also add a record rule for this model.

AungKoKoLin1997: Yes. It makes sense to me. I updated the code.

report_positioned_image/views/report_positioned_image_views.xml
```diff
@@ -0,0 +1,39 @@
...
+        <field name="name">report.positioned.image.view.form</field>
+        <field name="model">report.positioned.image</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <field name="image" widget="image" class="oe_avatar" />
+                    <group>
+                        <group>
+                            <field name="name" />
+                            <field name="company_id" />
```
yostashiro:
```suggestion
                            <field name="company_id" groups="base.group_multi_company" />
```

report_positioned_image/views/report_positioned_image_views.xml
```diff
@@ -0,0 +1,39 @@
...
+            </form>
+        </field>
+    </record>
+    <record id="report_positioned_image_view_tree" model="ir.ui.view">
+        <field name="name">report.positioned.image.view.tree</field>
+        <field name="model">report.positioned.image</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="name" />
+                <field name="company_id" />
```
yostashiro:
```suggestion
                <field name="company_id" groups="base.group_multi_company" />
```

report_positioned_image/views/report_positioned_image_views.xml
yostashiro:
```suggestion
                            <field name="company_id" groups="base.group_multi_company" />
```

report_positioned_image/views/res_company_views.xml
```diff
@@ -0,0 +1,40 @@
...
+                            <field name="pos_left" string="Left (mm)" />
+                            <field name="width" string="Width (mm)" />
+                            <field name="height" string="Height (mm)" />
+                            <field name="respect_image_ratio" />
+                            <field name="first_page_only" />
+                            <field
+                                name="company_id"
+                                column_invisible="1"
+                                required="1"
+                            />
```
yostashiro: Shouldn't we take the same arrangement as the view for the report action? I wasn't sure why it should be different here.

yostashiro (changes requested):

report_positioned_image/views/res_company_views.xml
AungKoKoLin1997: The `company_id` field configuration differs by context. In report action views, it's visible and optional to allow global images (`company_id=False`) across companies. In company views, it's hidden and required to enforce that company-specific images belong exclusively to that company, preventing unintended cross-company visibility.

AungKoKoLin1997 (approved):

report_positioned_image/models/ir_actions_report.py
```diff
@@ -0,0 +1,116 @@
...
+    def _prepare_html(self, html, report_model=False):
+        image_configs = self._get_positioned_image_configs()
+        if not image_configs:
+            return super()._prepare_html(html, report_model=report_model)
+        result = super()._prepare_html(html, report_model=report_model)
+        if not isinstance(result, tuple):
+            return result
+        bodies, res_ids, header, footer, specific_paperformat_args = result
+        if image_configs:
+            header = self._inject_images_into_header(header, image_configs)
```
yostashiro: The steps wouldn't reach this point without image_configs.
```suggestion
        header = self._inject_images_into_header(header, image_configs)
```

report_positioned_image/models/report_positioned_image.py
```diff
@@ -0,0 +1,113 @@
...
+            return
+        ratio = self._get_aspect_ratio()
+        if ratio and self.height > 0:
+            # Set context flag to prevent circular onchange
+            self.with_context(from_height_onchange=True).width = round(
+                self.height * ratio, 2
+            )
+
+    @api.constrains("company_id")
+    def _check_company_id_context(self):
```
yostashiro: Shall we change this to onchange, as it's not a real constraint and the check depends on the view?

report_positioned_image/models/ir_actions_report.py
```diff
@@ -0,0 +1,116 @@
...
+            css_class = "first-page" if image.get("first_page_only") else ""
+            class_attr = f' class="{css_class}"' if css_class else ""
+            parts.append(
+                f'<div{class_attr} style="{style}">'
+                f'<img src="{data_uri}" style="width: 100%; height: 100%;"/>'
+                "</div>"
+            )
+        return Markup("".join(parts))
+
+    def _get_report_company(self, res_ids):
```
yostashiro: Can we re-order methods in this file in the manner callees comes before callers? This method should come at the top with this policy.

report_positioned_image/models/ir_actions_report.py
AungKoKoLin1997: Done!

report_positioned_image/models/report_positioned_image.py
AungKoKoLin1997: Yes. I change to onchange.

report_positioned_image/models/ir_actions_report.py
AungKoKoLin1997: Done!

yostashiro (approved):
Code review and functional test. LGTM. 👍

PR #1143, opened 2026-03-09
[18.0][FIX] report_csv: self is model cls here - must use report_sudo for encoding and encode_error_handling

wpichler:
Problem got visible as soon as i tried to use such a report within a mail.template. mail.template likes binary attachment, so i did set an encoding. But got the same error, so i found this bug...


hbrunn (approved):
please prefix your PRs with the version and the module name as I did in the edit.

CI failure seems unrelated, but must be fixed nonetheless before merging

HviorForgeFlow (approved):

PR #1146, opened 2026-03-13
[16.0][ADD] report_docx

hbrunn:
This module allows to provide docx files as report templates to generate
docx reports.

Use Cases / Context
===================

This module was developed because for some users it is easier to use
the docx format for their templates.

As it uses a slightly different templating language, it is *not* a
drop-in replacement for report_py3o.

It also does not provide any format conversions, though it should be
pretty simple to write a glue module that converts the resulting docx
files to formats supported by report_py3o.

Configuration
=============

To configure this module, you need to:

1. Go to Settings / Technical / Actions / Report
2. Create a new report with type DOCX
3. Fill in the name of a model, ie ``crm.lead``
4. Upload your DOCX file in the ``Template`` field
5. To help with crafting expressions in the template, switch to the DOCX
   tab, select a record and fill in some expression. When happy with the
   result, copy the code into your template document. Don't forget to
   read the extensive documentation on the right hand side

Known issues / Roadmap
======================
-  support images and embedded objects
-  support concatenating docx files for multiple records
-  support embedding html


hbrunn:
CI fixed in https://github.com/OCA/reporting-engine/pull/1147

Stephanvandenberg (approved):
Tested and works brilliantly for users that want to output a report that is pre-filled with data from odoo models but can still be modified after it has been generated.

lmignon (approved):
Thank you for this new addon @hbrunn A nice alternative to py3O that seems easier to use. Time will tell if that's the case. LGTM (Code review + Functional tests)

PR #1149, opened 2026-03-24
[18.0] [FIX] report_substitute: account.move.send.wizard report substitution

amkarthik:
Issue Description:
When sending emails from invoices, the substitution rules are not being applied as expected.

Resolution:
This PR addresses and resolves the issue by ensuring that the substitution rules are correctly applied during email generation from invoices.

### DIFF
--- a/report_substitute/__manifest__.py
+++ b/report_substitute/__manifest__.py
@@ -6,7 +6,7 @@
     "summary": """
         This module allows to create substitution rules for report actions.
         """,
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
     "license": "AGPL-3",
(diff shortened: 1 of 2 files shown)


atchuthan (approved):
👍

atangaraj (approved):
👍

lmignon (commented):
Are you sure that with this change the substitution are not ran 2 times in nomale cases? A first time in the _pre_render phase and a second time at _render?

HviorForgeFlow (changes requested):
@amkarthik can you attend comment https://github.com/OCA/reporting-engine/pull/1149#pullrequestreview-4119120083

PR #1151, opened 2026-04-14
[16.0][FIX] report_xml: Remove useless and faulty post init hook

lmignon:
The post init hook created an ir attachment using content read as a string as value for the 'datas' field which expects a b64 encoded binary value.... This code comes from previous odoo version and should have been removed when migrating this addon

image


nicolas-delbovier-acsone (approved):
Code review

HviorForgeFlow (approved):

PR #1153, opened 2026-04-14
[18.0][FIX] report_xml: Remove useless and faulty post init hook

lmignon:
The post init hook created an ir attachment using content read as a string as value for the 'datas' field which expects a b64 encoded binary value.... This code comes from previous odoo version and should have been removed when migrating this addon

FWP #1152


lmignon:
fast tracking since initial pr is approved

PR #1158, opened 2026-05-01
[18.0][FIX] base_comment_template: tests fail if other modules use odoo_test_helper

etobella:
The used method only works if there is no other tests that requires to modify models...


pedrobaeza (approved):
Clear patch, so fast-tracking it

/ocabot merge patch

PR #1160, opened 2026-05-03
[18.0][ADD] report_builder & report_builder_account

etobella:
This modules is a grid-based KPI reporting tool that works across any Odoo model — not just accounting. Rows are KPIs defined by readable formulas referencing pre-definied KPI computations.

This module is inspired by mis-builder and shares its goal of grid-based KPI reporting. However, it takes a different architectural approach:
rather than being tightly coupled to accounting, it is designed to be pluggable on any Odoo model through small glue modules.

@sbidoul @pedrobaeza I think this might be interesting for you. It is quite similar to mis-builder, but has been developed from scratch. Also, it is more focused on generic reporting. Still work to do, but I think it can provide an idea. Also, the data can be mixed between kinds of reporting.

From a UX perspective, you will see that it is quite similar, however, it differs a lot on the backend and JS part.

Still a lot of work to do:

- [ ] Review how it works with a crowded Database
- [ ] Check how to make other operations on KPIs
- [ ] How to handle budgets
- [ ] Comparison of columns
- [ ] Add an inverse to port easily from mis-builder
- [x] Add PDF printing
- [x] Add XLSX printing


etobella:
The idea is to finish all missing parts on the Spanish OCA Days and test it there with the people there

PR #1162, opened 2026-05-07
[18.0][IMP] report_qweb_pdf_watermark: Make watermark configurable in configure document layout wizard

tarteo:
Simple UX improvement that makes the watermark configurable in the configure document layout wizard

image

image


report_qweb_pdf_watermark/wizards/base_document_layout.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_base_document_layout" model="ir.ui.view">
+        <field name="model">base.document.layout</field>
+        <field name="inherit_id" ref="web.view_base_document_layout" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='layout_background']" position="after">
+                <field name="pdf_watermark" widget="pdf_viewer" />
+                <div role="alert" class="alert alert-info">
+                    Note that the watermark is not reflected in the PDF preview.
```
hbrunn:
```suggestion
                    Note that the watermark is not reflected in the preview to the right.
```

I had to think a while what you mean here. And possibly make this div only show when a watermark file is set?

hbrunn (changes requested):

report_qweb_pdf_watermark/wizards/base_document_layout.xml
tarteo: Ah okay, I changed it according to your suggestions. Thanks!

ByteMeAsap (approved):
LGTM!!

HviorForgeFlow (approved):

hbrunn (approved):
thanks

PR #1165, opened 2026-05-30
[18.0][FIX] report_py3o: use py3o report in invoice Send flow

### LINKED ISSUE #1144: [18.0][BUG] py3o_report with invoice Send button
When using py3o_report, I can use the ir.action.reports with an ODT file and have a good report. But when using the Send button, with just Download in the wizard, the default Qweb template is used

## Module
py3o_report

## To Reproduce
account and py3o_report modules

fsmw:
## Summary
- Override the report_py3o pre-render path for py3o reports when the invoice Send flow calls _pre_render_qweb_pdf.
- Generate per-record PDF streams compatible with the account move send pipeline.
- Raise an explicit ValidationError when py3o output is not PDF in that flow.
- Add regression tests for stream generation and non-PDF guard.

## Why
In invoice Send plus Download flow, account calls _pre_render_qweb_pdf directly.
For py3o reports this bypassed py3o rendering and defaulted to qweb behavior.

## Reproduction
- Configure an invoice report with report_type set to py3o.
- Open invoice Send wizard and click Download.
- Before fix: qweb template is used.
- After fix: py3o report is rendered.

## Tests
- report_py3o suite on db test_1144_report_py3o.
- Result: 0 failed, 0 errors.

Closes #1144

### DIFF
--- a/report_py3o/models/ir_actions_report.py
+++ b/report_py3o/models/ir_actions_report.py
@@ -2,6 +2,8 @@
 # Copyright 2018 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import logging
+from collections import OrderedDict
+from io import BytesIO
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


rrebollo (approved):
Code Review. LGTM!
Would you be so kind to review mine #1166 in return?

PR #1167, opened 2026-06-03
[19.0][FIX] base_comment_template: comment_template_ids domain

Qlasta:
The field filtered candidates by comparing the computed, non-stored relational field  against a model name string:

    domain=lambda self: [(model_ids, in, self._name)]

The 19.0 domain engine (odoo.fields.Domain) rewrites a relational-vs- scalar leaf into the operator with a sub-domain on the comodel:

    (model_ids, any, [(display_name, in, [account.asset])])
    
 This matches `ir.model.display_name` (a translatable label eg. "Asset/Revenue Recognition"), never the
technical name, so it resolves to no records and the templates are
filtered out on inheriting models.

Build the domain from _search_model_ids instead.

### DIFF
--- a/base_comment_template/models/comment_template.py
+++ b/base_comment_template/models/comment_template.py
@@ -26,7 +26,9 @@ class CommentTemplate(models.AbstractModel):
         compute_sudo=True,
         comodel_name="base.comment.template",
         string="Comment Template",
-        domain=lambda self: [("model_ids", "in", self._name)],
+        domain=lambda self: self.env["base.comment.template"]._search_model_ids(
+            "in", self._name
+        ),
         store=True,
         readonly=False,
     )


HviorForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

base_comment_template/models/comment_template.py
```diff
@@ -26,7 +26,9 @@ class CommentTemplate(models.AbstractModel):
         compute_sudo=True,
         comodel_name="base.comment.template",
         string="Comment Template",
-        domain=lambda self: [("model_ids", "in", self._name)],
+        domain=lambda self: self.env["base.comment.template"]._search_model_ids(
+            "in", self._name
+        ),
```
pedrobaeza, line 31: Shouldn't it be just?:

```suggestion
        domain=lambda self: [("model_ids", "=", self._name)],
```

HviorForgeFlow:
@Qlasta can you attend https://github.com/OCA/reporting-engine/pull/1167#discussion_r3472299253 thank you in advice

Qlasta:
> @Qlasta can you attend [#1167 (comment)](https://github.com/OCA/reporting-engine/pull/1167#discussion_r3472299253) thank you in advice

I will get back at this tomorrow and test @pedrobaeza solution. Thank you.

base_comment_template/models/comment_template.py
Qlasta: @pedrobaeza I did not fully explain the problem in the description before. Not depending on operator, this  `([("model_ids", "=", self._name)])` will be converted to
`("model_ids", "any", [("display_name", "=", "account.asset")])` —
`ir_model.display_name` will be "Asset/Revenue Recognition" which will not match anything.

pedrobaeza (approved):
/ocabot merge patch

PR #1182, opened 2026-07-27
[15.0][IMP] report_xlsx: remove python libs included in odoo from requirements

smorita7749:
This PR removes external libs from __manifest__.py since they have already been listed in https://github.com/odoo/odoo/blob/15.0/requirements.txt

@qrtl QT6819


yostashiro (commented):
@smorita7749 Can you please explain the rationale in the commit message and in the PR description?

PR #1184, opened 2026-07-27
[19.0][FIX] bi_sql_editor: forward-port missing 17.0 improvements

alan196:
Forward-ports to 19.0 the bi_sql_editor changes merged in 17.0 that never reached this branch, with original authorship preserved (cherry-pick -x):

- #908: form view for SQL views + sum/avg aggregates in list view (adapted: tree → list terminology, self.env._, migration script renamed to 19.0.2.0.0; also fixes an if that should be elif in _prepare_action which made form-first views open the graph view)
- #1079: clickable many2one in list view
- #1086: field_description with translate=True

Not ported: #1036, already present in 19.0 via 637db79fb.

Version bumped to 19.0.2.0.0. Tests green (5 post-install tests), pre-commit green.


legalsylvain:
Thanks.

/ocabot merge patch

PR #1188, opened 2026-08-14
[MIG] report_qweb_encrypt: Migration to 19.0

moylop260:
Import PdfReader/PdfWriter from odoo.tools.pdf instead of PyPDF2.
Odoo 19 pins PyPDF2==2.12.1 only up to python3.12;
from python 3.13 (Debian Trixie) it pins PyPDF==5.4.0,
where the legacy PdfFileReader/PdfFileWriter names no longer exist:
 - https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/odoo/tools/pdf/_pypdf.py#L1-L15

Going through odoo.tools.pdf keeps the module working on both backends,
since that wrapper already absorbs the difference for core.

appendPagesFromReader() is kept on purpose: it is the only spelling available on
every backend, the PyPDF2 1.x one does not provide append_pages_from_reader():
 - https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/odoo/tools/pdf/_pypdf2_1.py#L28-L33

Replace self._context with self.env.context.
In 19.0 the _context property is decorated with @api.deprecated, so every access emits a
DeprecationWarning; it still returns self.env.context, but the indirection is scheduled for removal:
 - https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/odoo/orm/models.py#L5921-L5924


moylop260:
@etobella @jarroyomorales @kittiu 
WDY?

fernandahf:
LGTM

fernandahf (approved):

nhomar (approved):

PR #1191, opened 2026-08-24
[FIX] report_qweb_encrypt: accept single res_id

moylop260:
Odoo's PDF report API accepts `res_ids` either as a list or as a single integer. This mirrors how Odoo normalizes `res_ids` internally before continuing the rendering flow:

- https://github.com/odoo/odoo/blob/80e1a4464f75df5beeae0d8205a282da755c1a7d/odoo/addons/base/models/ir_actions_report.py#L659

There are real callers that use the single-id form. For example, `stock.picking` renders the delivery report with `self.id`:

- https://github.com/odoo/odoo/blob/ba4315ec85341431dd9bcd4f4d64217c37a4865f/addons/stock/models/stock_picking.py#L2014

`report_qweb_encrypt` kept using the original `res_ids` value after calling `super()` and sliced it to evaluate the encryption password. When a caller passed a single id, slicing the integer raised `TypeError` before the password could be read.

This change normalizes the local value before calling `_get_pdf_password` and adds a regression test for the single-id form.

### DIFF
--- a/report_qweb_encrypt/models/ir_actions_report.py
+++ b/report_qweb_encrypt/models/ir_actions_report.py
@@ -33,6 +33,8 @@ class IrActionsReport(models.Model):
         )
         report_sudo = self._get_report(report_ref)
         if res_ids:
+            if isinstance(res_ids, int):
+                res_ids = [res_ids]
             encrypt_password = self.env.context.get("encrypt_password")
             report = self._get_report_from_name(report_sudo.report_name).with_context(
                 encrypt_password=encrypt_password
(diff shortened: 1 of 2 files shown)


nhomar (approved):

fernandahf (approved):

luisg123v (changes requested):
Regarding commit message:
- Lines are too long, please wrap to e.g. 72 characters
- It says "add a regression test", which I think is not as accurate, as no tests are added, an existing one is modified to cover that case.

moylop260:
@luisg123v addressed in the same commit: the message is now wrapped at 72 characters, and the last paragraph says that the existing test is changed to pass a single `res_id`, instead of claiming a new test is added. No code change, and the five checks are green.

moylop260:
ocabot merge patch

PR #1194, opened 2026-09-03
[FIX] report_xlsx: sanitize worksheet name to prevent InvalidWorksheetName

stephansainleger:
Excel worksheet names are limited to 31 characters, cannot contain []:*?/\ and cannot start or end with an apostrophe. Report names do not guarantee these constraints, so xlsxwriter can raise InvalidWorksheetName (e.g. when a report name ends with an apostrophe once truncated to 31 chars).

Sanitize the sheet name in PatchedXlsxWorkbook._check_sheetname before the library validation and the duplicate handling, so a normalized name never triggers the error again. Add a test covering forbidden characters, leading/trailing apostrophes and length truncation.

### DIFF
--- a/report_xlsx/report/report_abstract_xlsx.py
+++ b/report_xlsx/report/report_abstract_xlsx.py
@@ -24,6 +24,12 @@ try:
             hard to debug the original issue. Even so, different names can become the
             same one as their strings are trimmed to those 31 character limit.
 
+            In order to avoid the library raising InvalidWorksheetName because the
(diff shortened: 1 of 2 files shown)


stephansainleger:
I'd like to request feedback from the community on this PR. I'm proposing it following a bug I encountered with one of my clients, where the company name resulted in an apostrophe landing exactly at the end of the sheet name. The goal is to avoid breaking the worksheet creation process. I also took the opportunity to clean up names more broadly, so that undesirable characters get sanitized as well.

PR #1195, opened 2026-09-11
[IMP] report_qweb_pdf_watermark: extract _get_watermark method

AnmollGarg:
Extracts watermark resolution logic from `_run_wkhtmltopdf` into a dedicated `_get_watermark(report_ref, docids=False)` method. 

This provides a clean hook for downstream modules to inject custom watermarks without duplicating PDF processing logic.

### DIFF
--- a/report_qweb_pdf_watermark/models/report.py
+++ b/report_qweb_pdf_watermark/models/report.py
@@ -68,6 +68,30 @@ class Report(models.Model):
             )
         return True
 
+    def _get_watermark(self, report_ref, docids=False):
+        """Return the binary watermark for the given report and documents."""
+        report_sudo = self._get_report(report_ref)
+        watermark = None
+        if self.pdf_watermark or report_sudo.pdf_watermark:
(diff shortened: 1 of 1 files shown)


report_qweb_pdf_watermark/models/report.py
```diff
@@ -68,6 +68,32 @@ def pdf_has_usable_pages(numpages):
             )
         return True
 
+    def _get_watermark(self, report_ref, docids=False):
+        """Extract the raw binary watermark for the given report and documents.
+        Can be overridden by child modules to inject custom watermarks.
```
hbrunn:
```suggestion
        """Return the binary watermark for the given report and documents.
```

hbrunn (approved):
I've have expected that those cases use an expression and some custom function, but don't mind this refactoring

hbrunn (changes requested):
please squash your commits

AnmollGarg:
@hbrunn I squashed the commits.

hbrunn (dismissed):

PR #1198, opened 2026-09-15
[FIX] report_py3o: repair corrupted Turkish translations

volkantasci:
**Description of the issue**

The Turkish translation files of `report_py3o` are corrupted: several non-ASCII characters have been replaced with `??`, for example:

- `Olu??turan` instead of `Oluşturan`
- `Son g??ncelleyen` instead of `Son güncelleyen`
- `G??r??nen ad` instead of `Görünen ad`
- `Ad??` instead of `Adı`

As a consequence, several strings are displayed with broken characters in the Turkish user interface.

**What this PR does**

Restores the missing Turkish characters in `i18n/tr.po` and `i18n/tr_TR.po` (including obsolete entries). It only changes translation strings; no code or functional behavior is modified.

**Possible root cause**

This looks like an encoding issue in the tooling that generated the files (Weblate). It might be worth fixing the corresponding entries on Weblate as well, so the correction is not lost on the next automatic synchronization.

Thanks in advance for your review.


volkantasci:
Closing this in favour of fixing the strings on Weblate (OCA's translation workflow).

I checked where the corruption actually comes from: the Turkish file downloaded from Weblate itself for this component still contains 4 broken `??` lines. The corrupted text is stored in Weblate and pushed into this repository, so a direct `.po` change here would be overwritten by the next Weblate sync.

The 19.0 component to fix is https://translation.odoo-community.org/projects/reporting-engine-19-0/reporting-engine-19-0-report_py3o/ (both `tr` and `tr_TR`) — I will correct the Turkish strings there so the fix lands through Weblate.

